Issue #21, opened 2013-05-04
company-pseudo-tooltip-show-at-point bug


Bad-ptr:
in company-pseudo-tooltip-show-at-point there is a code:

> (1+ (cdr col-row))

with this code the tooltip overlay overlaps the point of input and breaks it when the tooltip should be shown above point.

It looks like there must be branch with 1- for 'above' case and 1+ for 'below' case.
Or just row position without  neither 1+ nor 1- .

PR #57, opened 2014-02-05
company-pseudo-tooltip-frontend: Use one overlay per tooltip line to imp...

sadboy:
Currently, `company-pseudo-tooltip-frontend` uses one big overlay to cover the entire range of lines spanned by the tooltip. This disrupts the display of the region if other modes are also using overlays to convey visual information, such as linum, semantic-decoration, etc, creating an annoying visual distraction, especially when typing very quickly.

This patch reimplements company-pseudo-tooltip-frontend using additional overlays (one-per-tooltip-line) to more accurately cover the rectangular region occupied by the tooltip.

Known issues:
- Should probably limit maximum tooltip width,  to prevent overly long candidate items from messing up display.


trashofmasters:
I can report the same problem affecting php-mode when used in conjunction with Semantic overlays (breadcrumbs header).

dgutov:
@trashofmasters Please post a screenshot.

trashofmasters:
Unfortunately I cannot post a full-window screenshot as the buffer contains code that isn't owned by me.

Emacs header section showing with `semantic-idle-breadcrumbs-mode` decoration:
screen shot 2015-05-08 at 13 26 00

Assume I add a new class at the bottom of the buffer and trigger company-complete (either with the delay or a key binding):
screen shot 2015-05-08 at 13 25 07

when the pseudo-tooltip shows the completion right above the common text and pushing down the rest preview (or the point) to the next line:
screen shot 2015-05-08 at 13 25 18
screen shot 2015-05-08 at 13 25 29

When a completion candidate is selected the point is restored at the end of the completed text.
screen shot 2015-05-08 at 13 33 28

I had another interesting issue with `company-idle-delay` which caused tooltips to immediately close and change the position of the point when the variable was set to anything different than 0. Although I can't consistently reproduce it yet.

dgutov:
@trashofmasters First of all, it has absolutely no relation to what's been discussed in this issue. Don't be shy to open a new one.

It could be the same as #21, but that example works for me now. Please specify your Emacs version, as well as whether it's running in graphical mode.

Are you sure that the problem is not that you have `semantic-idle-completion-mode` enabled? If so, disable it.

> Unfortunately I cannot post a full-window screenshot as the buffer contains code

Why don't you create a new file, and upload as a gist, so I can use it?

PR #85, opened 2014-03-31
Sort candidates by previous completion choices

ilohmar:
Here is an experiment I am working on: Store how often candidates are selected and use this to rank the offered candidates with an entry in company-transformers.  The most complicated thing so far was resizing the completion history ring (probably hardly ever used in practice :).

I plan on writing some tests, and on extending it to distinguish b/w files or projects, but I am not really sure what I want for that.  Anyway, there are probably lots of bugs and inconsistencies that I'd be happy to hear about, besides any other comments you might have!

### DIFF
--- a/company-statistics.el
+++ b/company-statistics.el
@@ -36,38 +36,12 @@
   "Completion candidates ranking by historical statistics."
   :group 'company)
 
-(defun company-statistics--history-resize (option new-size)
-  (when (and (fboundp 'company-statistics--initialized-p)
-             (company-statistics--initialized-p))
-    ;; hash scoresheet auto-resizes, but history does not
-    (let ((new-hist (make-vector new-size nil))
-          ;; use actual length, so we work for freshly restored history
-          (company-statistics-size (length company-statistics--history)))
-      ;; copy newest entries (possibly nil) to new-hist
-      (dolist (i (number-sequence 0 (1- (min new-size company-statistics-size))))
-        (let ((old-i (mod (+ (- company-statistics--history-replace new-size) i)
-                          company-statistics-size)))
-          (aset new-hist i (aref company-statistics--history old-i))))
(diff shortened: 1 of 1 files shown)


dgutov:
@ilohmar Hey, sorry, this question has totally slipped my mind.

Would you prefer to keep it a separate package or fold it into Company? If the former, you'll need to request the access to `emacs` group in Savannah. If the latter, I'll take care of synchronization myself.

ilohmar:
No need to be sorry, thanks for getting back to this.

If you approve, I would surely like this to be folded into Company, mainly because I think it would be good to have some ranking mechanism onboard.

I do realize, however, that it would be the first (I think) addition of something which isn't a backend.

dgutov:
Should be fine, it's small, and at least we'll have an example of code using `company-completion-finished-hook`.

To clarify: you are still using it yourself?

ilohmar:
Absolutely!  I find it very helpful for etags completion in SQL, for example.  It's nowhere near perfect, but very often the completion I want is among the top 2 or 3 (of something like 20 or more candidates for a short prefix).

I guess that's because I haven't bothered anymore to tune the ranking --- more user feedback would probably help, and I am very willing to look into this again.

Looking forward to this!  Please let me know if there's anything else I need to do or can do to help.

dgutov:
After thinking about it more, I've pushed it as a separate package. Provided nobody complains, this should be good for popularity: a separate entry in `list-packages` is better than a line in company-mode's NEWS.md.

About synchronization, we'll do it as you prefer: either you let me know each time the package is ready for release (`Version:` should be incremented), or ask for `emacs` membership at http://savannah.gnu.org/, then you can do it youself.

ilohmar:
Awesome and fine with me.  I think I'll ask for membership in any case.  Thanks for your help so far!

PR #228, opened 2014-11-10
Improve Clang completion performance

### LINKED ISSUE #225: Clang's Large Output Makes Completion Slow
**TL;DR:** `company-clang` could be sped up by pre-filtering clang's output with `grep prefix`.

When company-clang calls clang for completions, if there is not enough context clang will output a massive list of every single identifier in the global scope. For most projects that use any sort of library, this can be thousands and thousands of symbols (I counted 40k for an IOS project, even when it couldn't load the UI library which is the biggest one).

The thing is it only takes clang a fraction of a second to output all these lines, those 40k lines only took 0.2s to pipe to `wc -l` in my shell. However, emacs takes a LONG time to read them all into a buffer. For me company-clang never manages to activate in the global scope because it takes up to a minute to read the thousands of lines (sometimes even a megabyte of text) into it's `* clang-output *` buffer.

This could be dramatically improved if some sort of pre-filtering was used. Currently `company-clang` doesn't take advantage of the prefix. Those 40k lines contain thousands of internal things starting with `__` even if the current prefix is `prin`. If the output could for example be piped through `grep $prefix` before being put in the buffer, it would dramatically increase performance.

Maybe company could include a shell script that only is used on Posix systems that pipes clang's output through grep, or perhaps there is some way to make reading the output into emacs much faster.

trishume:
On systems with `grep` installed this significantly improves performance in scenarios with little context by filtering the results by the prefix. It does this because clang only takes `0.3`s or so to create the completion but it takes a long time for Emacs  to read that into a buffer, this just trims down the size before it enters the buffer and minimizes that hit. 

This fixes #225 see that issue for discussion of problem and some timings.

On systems that don't have `grep` (like Windows :disappointed: ) it falls back on running clang directly.
Expected number of users hitting this code path is zero because it's very hard to install Clang on Windows and nobody does it, and who else doesn't have grep? I did test it though.

This change improves performance for global scope completion in a small project from around 20s (the async timeout means the results are ignored when they do arrive) down to a fraction of a second for a 2 character prefix. From a qualitative perspective it feels SOOO much faster all around.

### DIFF
--- a/company-clang.el
+++ b/company-clang.el
@@ -180,14 +180,20 @@ or automatically through a custom `company-clang-prefix-guesser'."
         (setq buffer-read-only t)
         (goto-char (point-min))))))
 
+(defun company-clang--run-clang-command (buf prefix args)
+  (if (executable-find "grep")
+      (let* ((join-args (mapconcat 'identity args " "))
+             (command (format "%s %s | grep -i \": %s\"" company-clang-executable join-args prefix)))
+        (start-process-shell-command "company-clang" buf command))
+    (apply #'start-process "company-clang" buf company-clang-executable args)))
+
 (defun company-clang--start-process (prefix callback &rest args)
   (let ((objc (derived-mode-p 'objc-mode))
         (buf (get-buffer-create "*clang-output*")))
     (with-current-buffer buf (erase-buffer))
     (if (get-buffer-process buf)
         (funcall callback nil)
-      (let ((process (apply #'start-process "company-clang" buf
-                            company-clang-executable args)))
+      (let ((process (company-clang--run-clang-command buf prefix args)))
         (set-process-sentinel
          process
          (lambda (proc status)


dgutov:
@trishume Could you try the current Emacs master? With it (as opposed to `emacs-24`), on GNU/Linux, I see no effect whether I bind `process-adaptive-read-buffering` no nil or not. Both ways are equally fast.

trishume:
I've recently switched back to emacs-ycmd for Clang based company completion since the popup displacement issue I was having with it was fixed, so I don't particularly care about this issue being fixed anymore and I seem to be the only one that has it.

If you really really want to get the bottom of this I might consider taking the time to uninstall my emacs and compile the latest master, go through some testing and then reinstall my normal emacs distribution (Homebrew for OSX isn't the best package manager so this takes a while). Otherwise feel free to close this bug.

dgutov:
Yes, I'd like to get to the bottom of this. There's no hurry, so please do this at your convenience.

trishume:
Ok I just tested with Emacs master built from Homebrew. I tried with both the Cocoa front end and the console version and both still have the problem.

This makes me suspect it is a problem somewhere in the syl20bnr/spacemacs distribution I'm using.

dgutov:
Not necessarily: it could just as well be a problem in the OS integration code.

But you can run `emacs -Q` to avoid the loading of your configuration. You'll have to set up company-mode from scratch then, though, including adding it to `load-path`.

khajaphysist:
Thank You dgutov it improved the performance to a great extent

PR #245, opened 2014-11-29
Blank out whitespace in popups.

yurikhan:
I am using `whitespace-mode` to visualize whitespace:

```
$ emacs -Q
```

``` elisp
(whitespace-mode 1)
(add-to-list 'load-path "~/.emacs.d/company-mode")
(load-library "company")
(company-mode 1)
```

Now move to the end of the first line and enter `(def`.

At this point, the `company-mode` completion popup apppears. Since its background and the padding of shorter lines is drawn using spaces, they get affected by `whitespace-mode` and displayed as middle dots.

In normal editing they are fontified with a face which assigns them a very subtle color, however in `company-mode` popups they are displayed in normal popup or default face. This is distracting.

company-whitespace-before

In this patch, I force whitespace in and before `company-mode`’s popups to have foreground identical to the background, so whitespace is invisible.

company-whitespace-after

(This may require tweaking also the `:distant-foreground` face attribute on Emacs 24.5 — I haven’t yet had the chance to upgrade and so can’t test.)

### DIFF
--- a/company.el
+++ b/company.el
@@ -146,17 +146,17 @@ buffer-local wherever it is set."
 
 (defface company-scrollbar-fg
   '((((background light))
-     :background "darkred")
+     :background "darkred" :foreground "darkred")
     (((background dark))
-     :background "red"))
+     :background "red" :foreground "red"))
   "Face used for the tooltip scrollbar thumb.")
 
 (defface company-scrollbar-bg
   '((default :inherit company-tooltip)
     (((background light))
-     :background "wheat")
+     :background "wheat" :foreground "wheat")
     (((background dark))
-     :background "gold"))
+     :background "gold" :foreground "gold"))
   "Face used for the tooltip scrollbar background.")
 
 (defface company-preview
@@ -1891,7 +1891,7 @@ character, stripping the modifiers.  That character must be a digit."
       (aref company-space-strings len)
     (make-string len ?\ )))
 
-(defun company-safe-substring (str from &optional to)
+(defun company-safe-substring (str from &optional to propertize-padding)
   (if (> from (string-width str))
       ""
     (with-temp-buffer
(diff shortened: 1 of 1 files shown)


Alexander-Miller:
According to [this](https://www.reddit.com/r/emacs/comments/4241oy/xwidget_branch_has_been_merged_into_emacs_251/) GTK widgets can now be embedded within emacs. I've no notion of what that would realistically change, so I'm quit curious: could this be used as an alternative means to display company's popups?

dgutov:
@Alexander-Miller Not 100% sure, but probably not.

dgutov:
@Alexander-Miller But that's not to say that there's no way to make a better-rendered popup "soon".

`x-show-tip` has problems, but if someone was really motivated to work on a popup library based on it, and iron out the kinks (probably via bug reports to the core, and discussions, but patches to the C code would be welcome as well), it can be solved. Other approaches, like menu-based popups, should be viable as well.

rlazoti:
I'm having exactly the same issue related by @yurikhan 
Is there any change for this PR to be merged?

dgutov:
@rlazoti I don't think so, sorry.

dojeda:
Urgh, scratch that. I've just realized this doesn't work. Somehow, my whitespace-mode was disabled during my testing. The idea might work if one chooses the correct function, I guess.

dojeda:
All right, reporting back. I managed to pull it off with:
```lisp
(defvar my-prev-whitespace-mode nil)
(make-variable-buffer-local 'my-prev-whitespace-mode)
(defun pre-popup-draw ()
  "Turn off whitespace mode before showing company complete tooltip"
  (if whitespace-mode
      (progn
        (setq my-prev-whitespace-mode t)
        (whitespace-mode -1)
        (setq my-prev-whitespace-mode t))))
(defun post-popup-draw ()
  "Restore previous whitespace mode after showing company tooltip"
  (if my-prev-whitespace-mode
      (progn
        (whitespace-mode 1)
        (setq my-prev-whitespace-mode nil))))
(advice-add 'company-pseudo-tooltip-unhide :before #'pre-popup-draw)
(advice-add 'company-pseudo-tooltip-hide :after #'post-popup-draw)
```

dgutov:
@dojeda Doesn't that make the whole window blink, with dots being removed before completion and re-added after?

dojeda:
@dgutov : Yes, indeed. I guess that I don't mind too much because my whitespace faces are only slightly different than my regular faces. I don't think one can enable/disable minor modes on overlays...

dgutov:
I'm going to close this PR (sorry), but see the linked issue and the new workaround I have just posted there.

Issue #286, opened 2015-01-24
company-clang doc-buffer implementation


mehw:
Clang can parse comments/documentation with **-Xclang -code-completion-brief-comments**, but this feature is not implemented in company-clang backend.

I propose a patch which will make the comments/documentation available when **company-show-doc-buffer** (_F1_) is called; you may customize this behaviour by **M-x customize-group** _RET_ **company-clang** once the patch is applyed:
- **Company Clang Parse Documentation**: _When non-nil, clang will parse completion's comments/documentation._
- **Company Clang Documentation Fill Column**: _Column beyond which automatic line-wrapping should happen._
- **Company Clang Documentation Justify**: _Specifies which  kind of justification to do._

[company-quickhelp](https://gist.github.com/dgutov/6dd7669697c5c0cd7e8f/) proposed in the "Documentation popup" #138 issue will show a popup containing comments/documentation parsed by clang.
# diff file:

``` diff
Author: mehw

Clang doc-buffer implementation.


diff -U 4 a/company-clang.el b/company-clang.el
--- a/company-clang.el  2015-01-24 13:18:20.000000000 +0100
+++ b/company-clang.el  2015-01-24 22:41:50.601346782 +0100
@@ -79,8 +79,78 @@
   (with-temp-buffer
     (insert-file-contents-literally file nil beg end)
     (buffer-string)))

+(defcustom company-clang-parse-documentation t
+  "When non-nil, clang will parse completion's comments/documentation."
+  :type 'boolean)
+
+(defcustom company-clang-documentation-fill-column 70
+  "Column beyond which automatic line-wrapping should happen."
+  :type 'integer)
+
+(defcustom company-clang-documentation-justify 'full
+  "Specifies which  kind of justification to do."
+  :type '(choice (const :tag "Full" full)
+        (const :tag "Left" left)
+        (const :tag "Right" right)
+        (const :tag "Center" center)
+                 (const :tag "None" none)))
+
+(defvar company-clang-summary-list
+  "Association list of tags and them related documentation.")
+
+(defun company-clang-meta-for-tag (tag)
+  "Extract the meta data of a TAG from `company-clang-summary-list'."
+  (let ((element (assoc tag company-clang-summary-list))
+   (match)
+   (meta))
+    (when element
+      (setq match (car element))
+      (setq meta (get-text-property 0 'meta match))
+      meta)))
+
+(defun company-clang-documentation-for-tag (tag)
[...]

PR #288, opened 2015-01-27
Clang doc-buffer implementation

mehw:
Clang can now fetch candidates' comments/documentation as discussed in #286 , this feature is disabled by default in the [commit 831b2fa](https://github.com/mehw/company-mode/commit/831b2fa7695102c55edbf0fb8c0dc34a6c209630) due to performance concerns, can be turned back on setting **company-clang-parse-documentation** to **t**.

### DIFF
--- a/company-clang.el
+++ b/company-clang.el
@@ -80,6 +80,521 @@ or automatically through a custom `company-clang-prefix-guesser'."
     (insert-file-contents-literally file nil beg end)
     (buffer-string)))
 
+;; The options "-ast-dump -ast-dump-filter" are part of Clang since version 3.2,
+;; but can parse comments starting from version 3.3.
+;;
+;; NOTE: Clang can parse only comments wrote in Doxygen style.
+(defconst company-clang-dump-ast-min-version 3.3
+  "Minimum Clang's version to use for AST dumps.")
+
+(defcustom company-clang-ast-system-headers-comments nil
+  "Show system headers comments in the AST dump."
+  :type 'boolean)
+
+(defun company-clang--can-parse-comments nil
+  "Return non-nil if the Clang's version in use is appropriate to parse
+comments."
+  (>= company-clang--version
+      company-clang-dump-ast-min-version))
+
+(defun company-clang--doc-buffer (candidate)
+  "Create the documentation buffer for a CANDIDATE."
+  ;; If there's no meta, maybe the CANDIDATE is a tag. In this case we use the
+  ;; CANDIDATE's prefix as meta.
+  (let ((meta (or (company-clang--meta candidate) (regexp-quote candidate)))
+         (doc (company-clang--get-doc candidate)))
+    (when doc
+      (company-doc-buffer
+       (concat meta "\n\n" doc)))))
+
+(defun company-clang--get-doc (candidate)
+  "Return the documentation found in the CANDIDATE's source file(s)."
+  (let ((info-list (company-clang--describe-candidate candidate))
+        doc item)
+    (while info-list
+      (when (setq item (company-clang--process-info (pop info-list)))
+        (setq doc (concat doc (when doc "\n\n") item))))
+    doc))
+
+;; TODO:
+;; - use the column number to locate the comment?
+;; - parse comments on the same line of the declaration?
+(defun company-clang--process-info (info)
+  "Return the documentation found in the source file as expressed by the
+candidate's INFO."
+  (when (company-clang--can-parse-comments)
+    (let* ((source-file (company-clang--value 'source-file info))
+           (line-begin (company-clang--value 'line-begin info))
+           (source-buffer (company-clang--source-buffer source-file))
+           pos doc-begin doc-end doc)
+      (when (and source-buffer line-begin)
+        (with-current-buffer source-buffer
+          (save-excursion
+            ;; Line of the candidate.
+            (goto-char (point-min))
+            (forward-line (1- line-begin))
+            ;; Search a comment block in the lines before.
+            (or (setq pos (company-clang--doc-//)) (setq pos (company-clang--doc-/*))))
+          (when pos
+            (setq doc-begin (car pos))
+            (setq doc-end (cdr pos))
+            (company-clang--buffer-highlight source-buffer doc-begin doc-end)
+            (setq doc (buffer-substring-no-properties doc-begin doc-end))
+            (format "%s: %d\n%s"
+                    (company-clang--source-file source-file) line-begin
+                    (company-clang--trim-doc doc))))))))
+
+(defun company-clang--source-file (file)
+  "Evaluate the candidate's FILE name.
+
+Return the candidate's source file."
+  (when (stringp file)
+    (if (string= file "<stdin>") (buffer-file-name)
+      ;; Convert relative to absolute path.
+      (expand-file-name file (file-name-directory (buffer-file-name))))))
+
+(defun company-clang--source-buffer (file)
+  "Evaluate the candidate's SOURCE-FILE.
+
+Return the candidate's source buffer."
+  (let (buf (source-file (company-clang--source-file file)))
+    (when (and source-file (file-exists-p source-file))
+      (setq buf (get-file-buffer source-file))
+      ;; Use the opened candidate's buffer only if it's unmodified.
+      (when (or (not buf) (buffer-modified-p buf))
+        (setq buf (get-buffer-create "*clang-doc*"))
+        (with-current-buffer buf
+          (buffer-disable-undo)
+          (erase-buffer)
+          (insert-file-contents source-file)))
+      buf)))
+
+(defun company-clang--buffer-highlight (buffer begin end)
+  "Highlight the text in the BUFFER between BEGIN and END.
+
+Highlight the text only when the BUFFER is not a file buffer."
+  (unless (buffer-file-name buffer)
+    (put-text-property begin end 'face '(:foreground "red") buffer)))
+
+(defun company-clang--doc-// nil
+  "Search a block of // comment in the lines preceding the current line.
+
+Return a cons of position begin and end of the comment."
+  (let (pos-begin pos-end)
+    (save-excursion
+      ;; Previous line.
+      (goto-char (1- (line-beginning-position)))
+      ;; Skip previous empty lines.
+      (while (re-search-backward "^$" (line-beginning-position) t)
+        (goto-char (1- (line-beginning-position))))
+      (while (re-search-backward "^[ \t]*//[^\n]*$" (line-beginning-position) t)
+        (setq pos-begin (match-beginning 0))
+        (when (not pos-end) (setq pos-end (match-end 0)))
+        ;; Previous line.
+        (goto-char (1- (line-beginning-position)))))
+    (when (and pos-begin pos-end)
+      (cons pos-begin pos-end))))
+
+(defun company-clang--doc-/* nil
+  "Search a block of /**/ comment in the lines preceding the current line.
+
+Return a cons of position begin and end of the comment."
+  (let (pos-begin pos-end)
+    (save-excursion
+      ;; Previous line.
+      (goto-char (1- (line-beginning-position)))
+      ;; Skip previous empty lines.
+      (while (re-search-backward "^$" (line-beginning-position) t)
+        (goto-char (1- (line-beginning-position))))
+      (when (re-search-backward "[*]/" (line-beginning-position) t)
+        (setq pos-end (match-end 0))
+        (when (re-search-backward "^[ \t]*/[*]" nil t)
+          (setq pos-begin (match-beginning 0)))))
+    (when (and pos-begin pos-end)
+      (cons pos-begin pos-end))))
+
+(defun company-clang--trim-doc (doc)
+  "Remove surrounding white spaces from the DOC, but keep the indentation."
+  (let (left-spaces trimmed line)
+    (when (stringp doc)
+      (with-temp-buffer
+        (insert doc)
+        (goto-char (point-min))
+        ;; First line.
+        (when (re-search-forward "^\\([ \t]*\\)\\([^\n]*\\)[ \t]*$" nil t)
+          (setq left-spaces (match-string 1))
+          (setq trimmed (match-string 2))
+          (goto-char (1+ (match-end 0)))
+          ;; Lines after the first.
+          (while (re-search-forward "^\\([^\n]*\\)[ \t]*$" nil t)
+            (setq line (match-string 1))
+            (goto-char (1+ (match-end 0)))
+            ;; Remove LEFT-SPACES from the beginning of line.
+            (when (string-match (format "^%s" left-spaces) line)
+              (setq line (substring line (match-end 0))))
+            (setq trimmed (concat trimmed "\n" line)))
+          trimmed)))))
+
+(defun company-clang--describe-candidate (candidate)
+  "Describe the CANDIDATE within the current context.
+
+Return the CANDIDATE's info-list."
+  (let ((context (company-clang--get-context candidate)))
+    (company-clang--process-context context)))
+
+(defun company-clang--filter (action candidate)
+  "Execute ACTION on the CANDIDATE's filter property.
+
+If set, the filter property suggests to filter the CANDIDATE's AST dump.
+
+ACTION:
+ 'get: return the filter property.
+ 'put: enable the filter property.
+ 'rem: remove the filter property."
+  (cond
+   ((eq action 'get)
+    (get-text-property 0 'filter candidate))
+   ((eq action 'put)
+    (put-text-property 0 1 'filter t candidate))
+   ((eq action 'rem)
+    (remove-text-properties 0 1 (list 'filter) candidate))))
+
+(defun company-clang--root (action candidate)
+  "Execute ACTION on the CANDIDATE's root property.
+
+If set, the root property suggests that the CANDIDATE doesn't have a parent
+in the AST dump (it is not a member of a data structure).
+
+ACTION:
+ 'get: return the root property.
+ 'put: enable the root property.
+ 'rem: remove the root property."
+  (cond
+   ((eq action 'get)
+    (get-text-property 0 'root candidate))
+   ((eq action 'put)
+    (put-text-property 0 1 'root t candidate))
+   ((eq action 'rem)
+    (remove-text-properties 0 1 (list 'root) candidate))))
+
+(defun company-clang--get-context (candidate)
+  "Return a list of identifiers depicting the context of the CANDIDATE.
+
+If the CANDIDATE is a member of a data structure, then the list represents
+that branch, otherwise the list contains only the CANDIDATE.
+
+The processing is performed in the current buffer. `point' is at the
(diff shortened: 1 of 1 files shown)


company-clang.el
```diff
@@ -80,6 +80,76 @@ or automatically through a custom `company-clang-prefix-guesser'."
     (insert-file-contents-literally file nil beg end)
     (buffer-string)))
 
+(defcustom company-clang-parse-documentation nil
```
dgutov: I'd call it `-parse-comments`, for symmetry with the clang argument name.

company-clang.el
```diff
@@ -80,6 +80,76 @@ or automatically through a custom `company-clang-prefix-guesser'."
...
+  "When non-nil, clang will parse completion's comments/documentation."
```
dgutov: parse completions' documentation comments

company-clang.el
```diff
@@ -80,6 +80,76 @@ or automatically through a custom `company-clang-prefix-guesser'."
...
+  :type 'boolean)
+
+(defcustom company-clang-documentation-fill-column 70
+  "Column beyond which automatic line-wrapping should happen."
+  :type 'integer)
+
+(defcustom company-clang-documentation-justify 'full
+  "Specifies which  kind of justification to do."
```
dgutov: extra space?

company-clang.el
```diff
@@ -80,6 +80,76 @@ or automatically through a custom `company-clang-prefix-guesser'."
...
+  :type '(choice (const :tag "Full" full)
+		 (const :tag "Left" left)
+		 (const :tag "Right" right)
+		 (const :tag "Center" center)
+                 (const :tag "None" none)))
+
+(defvar company-clang-summary-list
```
dgutov: I'd say this should include the default value (nil), but this variable probably shouldn't be used at all, see below.

company-clang.el
```diff
@@ -80,6 +80,76 @@ or automatically through a custom `company-clang-prefix-guesser'."
...
+  "Association list of tags and them related documentation.")
+
+(defun company-clang-meta-for-tag (tag)
```
dgutov: Why do you need this function at all? `company-clang--meta` should work well enough.

company-clang.el
```diff
@@ -80,6 +80,76 @@ or automatically through a custom `company-clang-prefix-guesser'."
...
+	(match)
+	(meta))
+    (when element
+      (setq match (car element))
+      (setq meta (get-text-property 0 'meta match))
+      meta)))
+
+(defun company-clang-documentation-for-tag (tag)
+  "Extract the documentation of a TAG from `company-clang-summary-list'."
+  (let ((element (assoc tag company-clang-summary-list))
```
dgutov: Here's the problem: `assoc` will find you a match that's `equal` to TAG. What about function overloads? Different functions can be named the same, as long as they accept different sets of arguments.

These overloads will have the same string value but different `meta` text property, which we add to completions in `company-clang--parse-output`. Adding the documentation strings would probably be too confusing, though. 

(That has just occurred to me) you could the meta values to index `company-clang-summary-list` instead. However, you could also try this approach: for each completion, call `cl-gensym` and put the new symbol on it as the value of `cc-tag` property. Then add `meta` and `doc` to the symbol as its properties, using `put`. Retrieving them would likewise be one `get-text-property` plus one `get`. It's worth testing to see if this indirection's overhead will be noticeable.

company-clang.el
```diff
@@ -80,6 +80,76 @@ or automatically through a custom `company-clang-prefix-guesser'."
...
+    (when element
+      (setq doc (car (cdr element)))
+      doc)))
+
+(defun company-clang-doc-buffer (tag)
+  "Create the documentation buffer for a TAG."
+  (let ((meta (company-clang-meta-for-tag tag))
+	(doc (company-clang-documentation-for-tag tag))
+	(emptylines "\n\n"))
+    (unless (and doc meta)
```
dgutov: Do we know of any case where `meta` is empty?

company-clang.el
```diff
@@ -109,8 +179,9 @@ or automatically through a custom `company-clang-prefix-guesser'."
 
 ;; TODO: Handle Pattern (syntactic hints would be neat).
 ;; Do we ever see OVERLOAD (or OVERRIDE)?
+;; Parse completion's comments/documentation too
 (defconst company-clang--completion-pattern
-  "^COMPLETION: \\_<\\(%s[a-zA-Z0-9_:]*\\)\\(?: : \\(.*\\)$\\)?$")
+   "^COMPLETION: \\_<\\(%s[a-zA-Z0-9_:]*\\) : \\(.*?\\)\\(?: : \\(.*\\)\\)\\{,1\\}$")
```
dgutov: 1. One extra indentation column.

2. `\\{,1\\}"` is the same as `?`.

company-clang.el
```diff
@@ -227,6 +301,8 @@ or automatically through a custom `company-clang-prefix-guesser'."
 
 (defsubst company-clang--build-complete-args (pos)
   (append '("-fsyntax-only" "-Xclang" "-code-completion-macros")
+	  (when company-clang-parse-documentation
+	    (list "-Xclang" "-code-completion-brief-comments"))
```
dgutov: Please don't use tabs. Our `.dir-locals.el` even contains `(indent-tabs-mode . nil)`.

company-clang.el
dgutov: Also, the docstring here should mention the minimum version Clang that supports this.

mehw: I'm searching the minimum Clang version required which can fetch comments...

company-clang.el
mehw: By default comments are shown in a single line in the doc-buffer, **company-clang-string-to-paragraph** gets around that problem.

I changed a little **company-clang-string-to-paragraph**, now no longer *nil*, *'none*, *'left* give the same result: *nil* or *'none* disable the justification, while *'left* organize the text as mere paragraph left justified.

What do you suggest?

dgutov: I'd avoid `none`, and just use `nil` for that value.

But the above comment was about `company-clang-summary-list`, not about `company-clang-documentation-justify`.

company-clang.el
mehw: I've changed the regexp:
"^COMPLETION: \\_<\\(%s[a-zA-Z0-9_:]*\\)\\(?: : \\(.+?\\)?\\)?\\(?: : \\(.+?\\)?\\)?$"

Prefix: **foo**

STRING | GROUP CAPTURED
--------|--------
"COMPLETION: " | / (no prefix)
"COMPLETION: blabla" | / (wrong prefix)
"COMPLETION: foobar" | 1 (2, 3 are nil)
"COMPLETION: foobar : " | 1 (2, 3 are nil)
"COMPLETION: foobar : [#int#]foobar(<#int a#>)" | 1, 2 (3 is nil)
"COMPLETION: foobar : [#int#]foobar(<#int a#>) : " | 1, 2 (3 is nil)
"COMPLETION: foobar : [#int#]foobar(<#int a#>) : Comment 1" | 1, 2, 3
"COMPLETION: foobar : [#int#]foobar(<#int a#>) : Comment 1 : " | 1, 2, 3
"COMPLETION: foobar : [#int#]foobar(<#int a#>) : Comment 1 : xyz" | 1, 2, 3

dgutov: Could you give an example where we actually see any of the first 4 variants in the clang output?

mehw: I wrote "impossible" clang results to test the regexp, e.g. **"COMPLETION: "** and **"COMPLETION: foobar : "** are conditions that should never be met, instead **"COMPLETION: foobar : [#int#]foobar(<#int a#>) : "** is possible (notice the empty comment) which is the 6th variant in the table above, **"COMPLETION: blabla"** and **"COMPLETION: foobar"** are still legit (if they where internals..., still have to try to add an empty Javadoc comment to an internal to get something like "COMPLETION: foobar : ").

Consider a `silly.c` file where you want to complete **foo**:
* **"COMPLETION: extern"** is a common case where the prefix could not be found, it is filtered out by `company-clang--parse-output`, but still there in \*clang-output\* buffer.
* **"COMPLETION: foobar : [#int#]foobar(<#int a#>) : "** is part of the output when clang finds an empty comment (/** */).
```C
// silly.c

// the next comment is empty: "COMPLETION: foobar : [#int#]foobar(<#int a#>) : "
/** */
int foobar(int a){return 0;}

void my_test() {
  foo
}
```

dgutov: Okay, thanks for the explanation. Notes:

 * The last `.+?` can as well be just `.+`: `.` doesn't match newlines.

* It should actually be `*`, I think. Otherwise, in the "empty comment" case, the space-colon-space will get appended to the meta string.

mehw: I wrote [regexp-testing-ground.el] (https://github.com/mehw/company-mode/blob/clang-doc-testing/regexp-testing-ground.el), you may add/change *prefixes*, *regexps*, *strings*, it should give clear answers: call `regexp-testing-ground-run-tests` and look at the buffer `*regexp-testing-ground*` for the results (add the file's load path and `(require 'regexp-testing-ground)` to `~/.emacs`:
* `"^COMPLETION: \_<\(%s[a-zA-Z0-9_:]*\)\(?: : \(.+?\)?\)?\(?: : \(.*\)?\)?$"`: **wrong result**: "COMPLETION: foobar : [#int#]foobar(<#int a#>) : " group 3 is an empty string instead of nil.
* `"^COMPLETION: \_<\(%s[a-zA-Z0-9_:]*\)\(?: : \(.+?\)?\)?\(?: : \(.+?\)?\)?$"` and `"^COMPLETION: \_<\(%s[a-zA-Z0-9_:]*\)\(?: : \(.+?\)?\)?\(?: : \(.+\)?\)?$"` give **same results**.

The `+` mean *at least one match*, while `*` means *no match or any match*.
```lisp
prefix : 'foo'
regexp : '^COMPLETION: \_<\(%s[a-zA-Z0-9_:]*\)\(?: : \(.+?\)?\)?\(?: : \(.+?\)?\)?$'
 pattern: '^COMPLETION: \_<\(foo[a-zA-Z0-9_:]*\)\(?: : \(.+?\)?\)?\(?: : \(.+?\)?\)?$'
  string : 'COMPLETION: '
   group 1: nil
   group 2: nil
   group 3: nil
  string : 'COMPLETION: blabla'
   group 1: nil
   group 2: nil
   group 3: nil
  string : 'COMPLETION: foobar'
   group 1: 'foobar'
   group 2: nil
   group 3: nil
  string : 'COMPLETION: foobar : '
   group 1: 'foobar'
   group 2: nil
   group 3: nil
  string : 'COMPLETION: foobar : [#int#]foobar(<#int a#>)'
   group 1: 'foobar'
   group 2: '[#int#]foobar(<#int a#>)'
   group 3: nil
  string : 'COMPLETION: foobar : [#int#]foobar(<#int a#>) : '
   group 1: 'foobar'
   group 2: '[#int#]foobar(<#int a#>)'
   group 3: nil
  string : 'COMPLETION: foobar : [#int#]foobar(<#int a#>) : Comment 1'
   group 1: 'foobar'
   group 2: '[#int#]foobar(<#int a#>)'
   group 3: 'Comment 1'
  string : 'COMPLETION: foobar : [#int#]foobar(<#int a#>) : Comment 1 : '
   group 1: 'foobar'
   group 2: '[#int#]foobar(<#int a#>)'
   group 3: 'Comment 1 : '
```

company-clang.el
mehw: When filtering by prefix gives a candidate like `extern` (possible prefix: *ext*), `float` (possible prefix: *flo*), etc. there is no `meta` (no arguments) and also no comments:

**"COMPLETION: extern"**
**"COMPLETION: float"**
etc.

In these cases `meta` is empty.

dgutov: Yes, thanks.

company-clang.el
dgutov: Okay, thanks. `.*` is not good enough, and I've missed the next-to-last question mark.

dgutov:
@mehw Do I understand it right that the whole effect is reached just using socket buffering? Sounds like kludge (and an unportable one).

You should test this on Emacs master. Like mentioned in https://github.com/company-mode/company-mode/pull/228#issuecomment-72228268, the value of `process-adaptive-read-buffering` doesn't have any perceptible effect there either.

mehw:
@dgutov I did the tests.

> You should test this on Emacs master. Like mentioned in [#228 (comment)] (https://github.com/company-mode/company-mode/pull/228#issuecomment-72228268), the value of `process-adaptive-read-buffering` doesn't have any perceptible effect there either.

Emacs 25.0.50.1
git clone -b master git://git.sv.gnu.org/emacs.git

The results are consistent with those at [comment 71876926] (https://github.com/company-mode/company-mode/issues/286#issuecomment-71876926); parsing candidates + docs:
* without `process-adaptive-read-buffering`: total of 88.42s, clang's output in 86.13s, parsing in 2.19s.
* with `process-adaptive-read-buffering`: total of 6.16s, clang's output in 3.86s, parsing in 2.21s.
* **file socket** + `process-adaptive-read-buffering`: total of 3.00s, clang's output in 0.63s, parsing in 2.25s.
* **file socket** alone: total of 2.97s, clang's output in 0.64s, parsing in 2.22s.

> Do I understand it right that the whole effect is reached just using socket buffering? Sounds like kludge (and an unportable one).

Writing "Experimental" I meant \<\<*Could someone other than me run some benchmarks, please?*\>\>.

I didn't know how to call it... "socket buffering" sounds nice... it is a **redirection to a temporary file**, Emacs feeds on that instead of on Clang's output. It should be portable, OS X was based on BSD if I'm not wrong, so it does have redirection **>**, and Windows should too. Once confirmed I'll remove the "Experimental" label.

As I implemented the feature, it is not thread safe, two or more concurrent threads may write/read the same file, but since Emacs is not multithreading there is still time to improve :smile:

dgutov:
> Writing "Experimental" I meant <<Could someone other than me run some benchmarks, please?>>.

Gotcha. Here are the results from my machine (Ubuntu 14.10).

Without temporary file (let's not call it a socket, actually):

* with `process-adaptive-read-buffering` bound to nil: clang's output in 2.57, parsing in 1.98.
* with default `process-adaptive-read-buffering`: output 3.09, parsing in 2.20.

With temporary file:

* with `process-adaptive-read-buffering` bound to nil: clang's output in 1.93, parsing in 1.99.
* with default `process-adaptive-read-buffering`: output 2.21, parsing in 1.97.

But the times are highly varied in both cases. `process-adaptive-read-buffering` definitely has some effect, and its fastest times are smaller.

On average, using the temporary file is faster, but not radically so, so I'm not sure it's worth the added complexity. And conceptually, we still pipe this whole output through Emacs.

dgutov:
Ah, disregard the above. I was testing completion in the same huge file, a lot of time was spent on simply sending its contents to Clang.

mehw:
Thanks!

I'll change the name to something more descriptive like "temporary file". I'll do more tests to see if it is really worth keeping it... I'm working on making Clang to retrieve the doc for a specific candidate, I hope to commit in the next hours, expect more comments on #286 . After this is done I'll do benchmarks with real projects... cross your fingers!

dgutov:
With temporary file:

* Output in 0.59, parsing in 1.95. The value of `process-adaptive-read-buffering` doesn't seem to make a lot of difference (whether I comment out that binding or not).

Without temporary file:

* Output in 1.39, parsing in 1.94. With `process-adaptive-read-buffering` on (the default), it's a tiny bit better, but nothing to write home about.

So yeah, a temporary file gives an improvement here, but again, not having to pipe this big an output would be even better.

dgutov:
I've simplified the test code, and now both test predictably fail in both builds: https://travis-ci.org/company-mode/company-mode/builds/54445518, apparently when they can't find the doc for `test_enum_a`.

The Clang version on Travis is 3.4, and indeed when I tried to run the tests with that version locally, the same failures happened. Maybe it outputs a slightly different AST, I don't know.

Regarding ambiguous matches in AST, maybe we should simply return the doc for the first match, or for all matches. While this problem will occur not only with structs, but also with class methods, it's still better to show some docs than none.

dgutov:
@mehw Beep-boop.

mehw:
Sorry for taking so long, I had a terrible year, but it's just life I guess...

Code rewrite: fetch a candidate's documentation directly from its source file.

Rebased and squashed with the upstream master.

> The Clang version on Travis is 3.4, and indeed when I tried to run the tests with that version locally, the same failures happened. Maybe it outputs a slightly different AST, I don't know.

Due to the code rewrite, tests are not yet implemented. I have to investigate with different Clang versions.

> Regarding ambiguous matches in AST, maybe we should simply return the doc for the first match, or for all matches. While this problem will occur not only with structs, but also with class methods, it's still better to show some docs than none.

Ambiguous matches are now resolved, when this is not possible the docs of the remaining ambiguous matches are organized in the output. C++ code (classes and methods) requires more testing.

dgutov:
Thanks. And sorry about your year.

I figured you gave up because this problem might not have a satisfactory solution, with unavoidable ambiguities and Clang's AST format being unstable. And that it was my fault for being encouraging.

dgutov:
But if your approach works at least for some use cases without regressing completion performance (which should be the case), great. Some quick feedback:

- There should be no mention of `-code-completion-brief-comments` because it's not used.

- There's no need for the change in `company-clang--completion-pattern`, for the same reason.

mehw:
> I figured you gave up because this problem might not have a satisfactory solution, with unavoidable ambiguities and Clang's AST format being unstable. And that it was my fault for being encouraging.

Don't worry, I'm happy to collaborate with you. In the context of some tests I run: Clang 3.2 produces a not usable AST dump, from 3.3 to 3.4.2 and from 3.5.0 to 3.7.1 the dumps follow the same format; 3.3-3.4.2 and 3.5.0-3.7.1 are slightly different, which is why the build on Travis (Clang 3.4) failed, I'm working with Clang 3.5.0.

3.4.2:
Dumping `<`anonymous struct`>`::alfa:
FieldDecl 0x30b4900 <<stdin>:78:3, col:7> alfa 'int'

3.7.1
Dumping `(`anonymous struct`)`::alfa:
FieldDecl 0x41a6bd8 <<stdin>:78:3, col:7>` col:7` alfa 'int'

More tests are required, though. I have to look again at the code to make it conformant with versions 3.3 to 3.7.1 where it is possible.

> But if your approach works at least for some use cases without regressing completion performance (which should be the case), great.

There should be no impact on completion performance, since the documentation is shown upon user request. The disambiguation process requires different calls to Clang.

dgutov:
People say that Clang's internal structures will always be unstable. Would accessing the AST through libclang be more stable? Since Emacs 25.1 will support loading dynamic modules now, you could experiment with that, too.

mehw:
I'm trying to get it to work using `cindex.py`, Python bindings to libclang. There's no documentation though. I'm currently studying the python module in my spare time. I still don't know if this will be more stable than an AST dump via Clang cli, I have to try different libclang versions too.

mehw:
I'm still on it. Just to let you know. I see that Clang's releasing schedule is very busy; I hope they wont change libclang too much, though. Later.

dgutov:
I guess it's time to close this. Thanks for trying.

Issue #290, opened 2015-01-30
company-lighter is not customizable


milkypostman:
I will try to write up a PR to fix this if I get time.

dgutov:
Would you be fine with a defvar?

milkypostman:
Isn't it already a defvar? Why not defcustom?

dgutov:
Currently, its value is changed at runtime. The change I have in mind would make it use the mode-line format, like 3716acbb5529092fe7697c18f3a02e3919b02bfb did to `company-search-lighter`. I'll have to see if there are any performance concerns.

> Why not defcustom?

Would you make the user edit the mode-line format structure? With what `:type`?

milkypostman:
I didn't originally understand your response but now I do. But I already tried to make my suggestion in #291 

Customizable variables are really no different than `defvar` except that the customize function can see them and present the UI. I am trying using customize as much as possible instead of `setq` lately and it's been OK. I think that startup is a bit faster and it doesn't have as many problems with variables that are unknown but still trying to be set. Meaning if you don't have a package loaded you can still have the variable set.

PR #291, opened 2015-01-30
Update the lighter code to allow customization.

milkypostman:
Now use the variable `company-lighter` to define the character to use in the mode-line for company-mode. There is a buffer-local version of the variable that is used for each buffer to decorate the mode-line.

Renamed variables to make them more intuitive.

### DIFF
--- a/company.el
+++ b/company.el
@@ -662,9 +662,8 @@ asynchronous call into synchronous.")
       (unless (keywordp b)
         (company-init-backend b))))))
 
-(defvar company-default-lighter " company")
-
-(defvar-local company-lighter company-default-lighter)
+(defcustom company-lighter " company" "Default lighter string.")
+(defvar-local company-buffer-lighter company-lighter)
 
 ;;;###autoload
 (define-minor-mode company-mode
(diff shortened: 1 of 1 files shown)


dgutov:
Thanks for the effort, but it goes against my suggestion in #290. Do you feel strongly about customizing the ` company-` string, but not the rest of the behavior?

milkypostman:
To be clear, your suggestion is to have the lighter actually be an arbitrary piece of elisp that the mode-line will eval?

I definitely don't think that's a bad idea and I can't imagine the performance cost will be that high unless the function's you're calling will be blocking or some such. 

*but*, regardless of which direction you choose, there is likely to still be a prefix that you want to put on your lighter and it may be convenient to have that be a separate variable.

So your lighter will look something like,


```lisp
(defcustom company-lighter-pfx "c" "Prefix for company modeline functionality")
(defcustom company-lighter '(" " company-lighter-pfx (if (company-active-extension) ...)))
```

I ignored many of the details here but my point is that I normally just care about shortening "company" to "c" and I don't want to actually customize the whole `company-lighter` formatting. It's a bit different from search. In the search case i may want to totally change that. But in the mode-line, I normally just want to shorten one prefix but want to keep the rest of the formatting of the minor-mode-line the same.

dgutov:
I see, sounds good.

milkypostman:
I'll try to fix this when I get a chance.

dgutov:
Don't bother: unless you'd like to do the mode-line conversion, I'll do this myself.

dgutov:
And done.

PR #312, opened 2015-02-25
company-show-location: Bind other-window-scroll-buffer

cpitclaudel:
Should the changes in 448bcd8 be applied to the location function too?


dgutov:
Yes indeed, it should have the same changes applied.

PR #334, opened 2015-03-26
Fix docstring

vermiculus:


dgutov:
Oh. Thanks. :)

vermiculus:
Not at all :)

Issue #222, opened 2014-11-06
Backends that need setup to work correctly are enabled by default


r4nt:
I'm wonder what the rationale is to enable backends by default that will just spew errors to Messages when they aren't set up correctly (like the etags or clang backends).
Is there a list of backends that are known to not have any dependencies that would make them break?

PR #382, opened 2015-07-27
company-etags-find-table: fix fallback tags file expansion

ilohmar:
locate-dominating-file returns a directory or nil, so the `(file-regular-p file)` was always `nil`.  Also, the call to `expand-file-name` seemed weird --- the dir should be the optional 2nd argument, not the first.

With the fix, the function should actually return an absolute path to any "TAGS" file, as originally intended.

### DIFF
--- a/company-etags.el
+++ b/company-etags.el
@@ -51,11 +51,11 @@ buffer automatically."
 (defvar-local company-etags-buffer-table 'unknown)
 
 (defun company-etags-find-table ()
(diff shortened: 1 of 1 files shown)


dgutov:
Indeed, that's my bad.

Thank you, but please add a check that `TAGS` is not a directory. Otherwise, this re-introduces #222.

ilohmar:
Thanks for catching that, should be checked now.

dgutov:
:+1:

PR #384, opened 2015-07-30
company-etags: simpler and less intrusive table building

ilohmar:
I am trying to clean up all my etags addons and workarounds (it's not pretty..), and this was a blocker: It seems that `save-excursion` and `visit-tags-table-buffer` are not necessary here --- `tags-completion-table` does that dance properly itself.  But the current code *does* introduce problems when there are different completion tables per buffer: `v-t-t-b` selects a tags buffer itself, and when `t-c-t` tests if there is a buffer-local cached value for the completion table, it does *not* do so in the original buffer where we wanted to complete, but in the tags buffer.

Replacing that by a plain call to get the completions is simpler, and it avoids this and potential other problems.

### DIFF
--- a/company-etags.el
+++ b/company-etags.el
@@ -70,9 +70,7 @@ buffer automatically."
         (completion-ignore-case company-etags-ignore-case))
     (and (or tags-file-name tags-table-list)
          (fboundp 'tags-completion-table)
-         (save-excursion
-           (visit-tags-table-buffer)
-           (all-completions prefix (tags-completion-table))))))
+         (all-completions prefix (tags-completion-table)))))
 
 ;;;###autoload
 (defun company-etags (command &optional arg &rest ignored)


dgutov:
I'm not sure about this. See `tags-lazy-completion-table`, it does this dance as well.

AFAICT, `tags-completion-table` is supposed to be set locally in buffers visiting tag tables, it's not something to customize for each normal file buffer.

If you cache the table in normal file buffers, Emacs will have to re-read the tag tables each time you open a new file, which would be unfortunate.

ilohmar:
Ok, now I see that this is probably the way to understand the current code of etags and its intent.  I cannot seem to wrap my head around it...  How is that supposed to work if you have several tags files in `tags-table-list`, you only get completions from the first one that `v-t-t-b` deems fitting, right?

In any case, thanks for your comment and, for now, sorry for the noise.

dgutov:
> if you have several tags files in tags-table-list, you only get completions from the first one that v-t-t-b deems fitting, right?

No, `tags-completion-table` function iterates through all of them, and apparently saves the combined list to the buffer-local variable in the buffer visiting the first tag file.

Admittedly, it's a little weird and has room for improvement.

PR #391, opened 2015-08-18
Map C-n and C-p to company-select-next & -previous

egh:
I found myself typing `C-n` and `C-p` to choose company selections, but these keys were not mapped. This maps those keys during completion.


dgutov:
Thanks, but this is a deliberate choice: `previous-line` and `next-line` commands exit completion.

Users that prefer this customization should keep it in their init scripts.

egh:
Cool, thanks for considering this! I'll keep my local modifications.

PR #397, opened 2015-09-04
Makefile: allow package and elpa targets to run without git

bremner:
This is handy for people working from e.g. a zip or tar snapshot of the git
repo.

I'm currently using this patch in the Debian packaging for company-mode


dgutov:
Looks good, thanks.

PR #398, opened 2015-09-05
Separate how to write a backend from company-backends.

Wilfred:
I'd like to propose a change to the docstring for `company-backends`. You can probably think of a better way of doing this, this pull request is just a suggestion.

Currently, the docstring describes two different things: (1) how to configure company to use different backends, and (2) how to write a backend. I think it would be clearer if these were written in different places.

I completely missed the documentation on grouped backends the first time I read the docstring, because I just skimmed over the information on creating a backend and didn't spot it. I think this approach is clearer, what do you think?

### DIFF
--- a/company.el
+++ b/company.el
@@ -44,20 +44,114 @@
 ;; `company-begin-backend'.  For example:
 ;; M-x company-abbrev will prompt for and insert an abbrev.
 ;;
-;; To write your own back-end, look at the documentation for `company-backends'.
+;; Sometimes it is a good idea to mix several back-ends together, for example to
+;; enrich gtags with dabbrev-code results (to emulate local variables).
+;; To do this, add a list with both back-ends as an element in `company-backends'.
+
+;;; Writing your own backend:
+;;
 ;; Here is a simple example completing "foo":
 ;;
 ;; (defun company-my-backend (command &optional arg &rest ignored)
+;;   "Complete foo."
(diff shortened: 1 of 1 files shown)


dgutov:
Hmm, some improvement is clearly needed, and thank you for the attempt. However, it's usually that docstrings contain the most complete (and possibly dry) description of variable's values, function's calling conventions, etc, whereas Commentary and manuals contain the information for user consumption, often in more verbose form.

Here, however, you also moved the technical information into Commentary, and reading `company-backends` docstring, by itself, is insufficient to find out the details of backend definition.

Wilfred:
As a new company user, I want to just read how to use `company-backends` to use my favourite backends. It's only later, when I'm an enthusiastic company user, that I'm interested in writing new backends. I think it's clearer if they're separate, do you agree?

If you'd rather we didn't put the docs in header, perhaps elsewhere? Suppose we defined a `company-example-backend` and used its docstring to document defining a backend?

dgutov:
> I think it's clearer if they're separate, do you agree?

Sure, but I think `company-backends` docstring should have all information required to write a new backend, even if in short form.

The header commentary (which is displayed to people who press `RET` on an item in `list-packages` list) can, on the other hand, target new users exclusively. Even get rid of the "writing a backend" example, although I'm not sure where to move it to.

Issue #79, opened 2014-03-14
Merging disparate backends


dgutov:
`company-backends` allows list elements containing several backends. If it sees one, the results from the contained group are merged on an equal basis: if at least one of them returns a non-nil prefix, that prefix is returned, and all completion candidates returned from backends that respond with the same prefix are merged.

Problem: some backends are tied to specific major or minor modes, and others work in just about any context (like `dabbrev-code`, `keywords` and `yasnippet`). So creating a group containing one of the latter set and putting it at the front of `company-backends` virtually guarantees that no other elements of that list will be ever used.

Proposal: add a keyword that will separate "primary" backends from "secondary" backends in a group. `company--multi-backend-adapter` will only consider `prefix` results from backends before that keyword, and if all of them return nil, the next element in `company-backends` will be tried.

What name to use for the keyword?

Current candidates: `:secondary`, `:and`, `:skip-prefix`.

vspinu:
I am reading through this discussion and, quite frankly I don't understand how you came up with `:with` keywords. 

The original requirement of @dgutov was 

> add a keyword that will separate "primary" backends from "secondary" backends in a group. company--multi-backend-adapter will only consider prefix results from backends before that keyword

And then @proofit404  proposed 

> :with also a nice keyword for this. But for example I want yasnippet back end work together with any other back end.

Let's consider `(A B :with C)`. The @dgutov's quote above means "Use C only if A or B are used" the @proofit404's quote says "Use C with A and B".  So how did you end up with implementing `Use A and B with C'? 

Now `:with` is also used in `company-sort-by-backend-importance` and its meaning there is completely unrelated to all of the discussion above. 

If anything, `:secondary` is the most generic and has clear correspondence to the English language. `(A B :secondary C)` means just that, `C` is the secondary backend. That being said, it's not a good API to use :with for two marginally related things. 

I think grouped backends should have similar interface to normal backends. You would like to have keywords for such actions like `:sorted`, `:duplicated`, `:ignore-case`, `:transform` that would overwrite the behavior of sub-backends. 

How about building grouped backends hierarchically? Each grouped backend is a sublist with its own key-value pairs and can contain other sub list. No need to limit yourself to 2 layers only.  Then have plist style key-value pairs within each group to qualify what it should do with respect to the immediate parent backend. For example

```elisp
`(setq company-backends 
      '(A 
        (B (C :importance 2)) ; a number to allow multiple importance levels; default 1
        (D E :sorted t :duplicated nil)
[...]

PR #402, opened 2015-09-12
[Fix #317] Allow :sorted keyword in grouped back-ends and some doc improvements

### LINKED ISSUE #317: Concatenate (no sorting) in group backends?
I have a grouped backend of [two backends](https://github.com/emacs-ess/ESS/blob/master/lisp/ess-r-completion.el#L347-L384), one for arguments of a function (valid only within a function call) and one for all other objects.  I want that argument completions be shown before all other objects. I also don't want the arguments to be sorted, as they must appear in the declared order. 

The only solution that I see right now is to manually merge the two backends into one.

Any ideas? Thanks.

[related: #307]

vspinu:
I have designed it such that further addition of keywords will not break older versions of company. This is why all keywords are removed before the `pcase`.

### DIFF
--- a/company.el
+++ b/company.el
@@ -354,7 +354,8 @@ prefix, but match it in some backend-defined way).  Backends that use this
 feature must disable cache (return t to `no-cache') and might also want to
 respond to `match'.
 
-Optional commands:
+Optional commands
+=================
 
 `sorted': Return t here to indicate that the candidates are sorted and will
 not need to be sorted again.
@@ -414,25 +415,25 @@ The backend should return nil for all commands it does not support or
 does not know about.  It should also be callable interactively and use
 `company-begin-backend' to start itself in that case.
 
-Grouped backends:
+Grouped backends
+================
 
-An element of `company-backends' can also itself be a list of backends,
-then it's considered to be a \"grouped\" backend.
+An element of `company-backends' can also be a list of backends.  The
+completions from backends in such groups are merged, but only from those
+backends which return the same `prefix'.
 
-When possible, commands taking a candidate as an argument are dispatched to
-the backend it came from.  In other cases, the first non-nil value among
-all the backends is returned.
+Whenever makes sense, company commands taking a candidate as an argument
+are dispatched to the backend it came from.  In other cases, the first
+non-nil value among all the backends is returned.
 
-The latter is the case for the `prefix' command.  But if the group contains
-the keyword `:with', the backends after it are ignored for this command.
+The group can also contain keywords. Currently, `:with' and `:sorted'
+keywords are defined. If the group contains keyword `:with' , the backends
+listed after this keyword are ignored for the purpose of `prefix'
+command. If a grouped backend contains keyword `:sorted', the
+final (merged) list of candidates is not sorted.
 
-The completions from backends in a group are merged (but only from those
-that return the same `prefix').
-
-If a grouped backend contains keyword `:sorted', the final (merged) list of
-candidates is not sorted.
-
-Asynchronous backends:
+Asynchronous backends
+=====================
 
 The return value of each command can also be a cons (:async . FETCHER)
 where FETCHER is a function of one argument, CALLBACK.  When the data
--- a/company.el
+++ b/company.el
@@ -429,6 +429,9 @@ the keyword `:with', the backends after it are ignored for this command.
 The completions from backends in a group are merged (but only from those
 that return the same `prefix').
 
+If a grouped backend contains keyword `:sorted', the final (merged) list of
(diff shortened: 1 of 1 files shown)


vspinu:
BTW, I think `:with` keyword is not really suggestive of the task that it performs. How about `:prefix-ignore-rest`?

dgutov:
> How about :prefix-ignore-rest?

It makes more sense on the lower level, but misses the advantage of the more succinct `:with`: separating backends semantically in two groups, say, primary and secondary. It corresponds more closely to the effective meaning: "if none of the backends before :with are used, none of those after it are used either".

One has to read the docstring anyway to find out about the keyword's existence, so maybe it's okay if its meaning is not 100% apparent from its name, as long as they correlate well.

vspinu:
>  It corresponds more closely to the effective meaning: "if none of the backends before :with are used, none of those after it are used either".

Sorry, but :with doesn't ring a bell with me at all. The above sentence is imprecise; "not used" can mean anything. For a developer of a backend who  is reading the docstring it's crucial to understand what's going on a low level. You can ask for the opinion of others, but  I really had trouble understanding what :with meant till I got down to he code. 

The main problem is a semantic one.  It doesn't make sense for the "with" preposition to refer to the previous items. If anything :with in `(A B :with C D)` should refer to C and D. That is, "use prefix command *with* C and D" , which is exactly the opposite of what it does right now. 

You can have `:prefix-ignore` if length is a concern. `(A B :prefix-ignore C D)`.

dgutov:
> For a developer of a backend who is reading the docstring it's crucial to understand what's going on a low level.

But that a user-level keyword. A developer can read the implementation anyway.

> The above sentence is imprecise; "not used" can mean anything

Maybe; you can suggest improvements. What I mean it, skipped for the purpose of completion, isn't asked for its completions, etc. Which is a consequence of returning `nil` to `prefix`, but to make that connection, the user needs to read a different part of documentation.

> It doesn't make sense for the "with" preposition to refer to the previous items.

Why not? In English, "with" connects words that come before it with words that come after. "spoon with honey", for instance.

> You can ask for the opinion of others, but I really had trouble understanding what :with meant till I got down to he code.

I did ask for suggestions for the name of this keyword: #79.

vspinu:
>  "with" connects words that come before it with words that come after. "spoon with honey", 

I hope you don't mean that  `(A B :with C)`  is translated into "A, B with C".  `:with` doesn't connect A, B and C, but rather a computation of some sort (prefix filter) with some backends from the list. 

So your spoon is "prefix command" and your "honey" is backend C. So you get "prefix command works with C". And confusingly, prefix command currently does exactly the opposite - it works with A and B.

vspinu:
How about `:split`? It naturally refers to the list which is split in two. It can also be used in other contexts, not just with prefix filter (as it is currently happening with `company-sort-by-backend-importance`).

vspinu:
> How about :split? 

This might be great for multi-splits `(A B :split C :split D E)`. For instance `company-sort-by-backend-importance` can sort within each group of backens, but preserve the order between backends.

dgutov:
> I hope you don't mean that (A B :with C) is translated into "A, B with C". :with doesn't connect A, B and C

It does: it says "C is used only if A or B is used". And actually, C's `prefix` *is* consulted in this case: it's compared to A's or B's prefix (whichever was non-nil), and C is queries for completions only is they are equal. There's more context in the issue I've linked above; let's continue this discussion there, if you like.

> How about :split?

You'd have to define its semantics first. They seem to be different.

vspinu:
> It does: it says "C is used only if A or B is used". And actually, C's `prefix`
> *is* consulted in this case: it's compared to A's or B's prefix (whichever was
> non-nil), and C is queries for completions only is they are equal. There's more
> context in the issue I've linked above; let's continue this discussion there, if
> you like.

This is sooo complicated. Switching to that thread instead.

dgutov:
I've switched to the hyphen-less spelling across the project. Please rebase this branch on top of it, and I'll merge.

Also disregard my objections against `:sorted` without value. The implementation is easier this way, and the difference is not that important.

vspinu:
Sorry I haven't had a chance to look at it again :( Will fix this tomorrow.

> Also disregard my objections against :sorted without value

Are you sure? If other options will be added in the future, they will  likely need values. Or it's simply not worth spending too much time on this given the imminent obsoletion?

dgutov:
> Or it's simply not worth spending too much time on this given the imminent obsoletion?

This.

vspinu:
Should be good to go!

dgutov:
Thanks! :thumbsup:

Issue #47, opened 2014-01-21
[feature request] Flex/Fuzzy matching


PythonNut:
[`flx`](https://github.com/lewang/flx) is a very smart sublime-text like fuzzy/flex matching library. After using similar flex matching in Vim's [`YouCompleteMe`](https://github.com/Valloric/YouCompleteMe) I miss it in `emacs`. I would seriously recommend trying it out for a bit. It's amazing how much code you can generate with a few key strokes using these advanced algorithms.

    (kcdg<tab> eism<tab> "jj" ens<tab>)

Becomes

    (key-chord-define-global evil-insert-state-map "jj" evil-normal-state)

I already try to use `flx` for everything (minibuffer, hippie-expand etc.), and the only thing left is auto-completion. I can't get it to work myself because it's so complicated. I also have a [feature request](https://github.com/auto-complete/auto-complete/issues/273) for [`auto-complete-mode`](https://github.com/auto-complete/auto-complete), but they seem busy with their big 2.0 rewrite and it's no going anywhere.

By the way, thanks for `company-mode` it's very fast and gets out of my way. I think it's awesome!

Thanks,
PythonNut

PythonNut:
>  >   Why not work on the interface first?

> You (or anyone else) are welcome to implement the relevant completion-styles-alist elements. Should be relatively easy.

That part [is now done](http://emacs.stackexchange.com/a/13508/2642), and in fact it doesn't require `flx`, because `completion-style` does not control sorting. 

The systems that control sorting in vanilla minibuffer completion don't seem to be used by company. How can we now add a custom sorting function?

dgutov:
> How can we now add a custom sorting function?

Via `company-transformers`, for example. If you want it to apply to all backends.

dgutov:
> The systems that control sorting in vanilla minibuffer completion

What are those, by the way?

PythonNut:
@dgutov they're pretty ugly, and I admit, I still don't quite understand how they work. 

* Displayed completions are sorted in `minibuffer-completion-help`
* When cycling, sorting is defined in `completion-all-sorted-completions`

In both cases, the sorting comparison function is extracted from the completion metadata:
```emacs
(let* ((completion--metadata (buffer-substring-no-properties
                                            start (point))
                                           base-size md
                                           minibuffer-completion-table
                                           minibuffer-completion-predicate)
    (sort-fun (completion-metadata-get all-md 'cycle-sort-function)))
    ...)
```

[...]

PR #418, opened 2015-11-03
Allow advising company-dabbrev(-code)? regex generation

PythonNut:
This should allow me to write an advice that implements fuzzy matching for `company-dabbrev` and `company-dabbrev-code`. I doubt the extra function-call overhead will cost you very much relative to the real work done by the respective backends.

### DIFF
--- a/company-dabbrev-code.el
+++ b/company-dabbrev-code.el
@@ -69,7 +69,7 @@ also `company-dabbrev-code-time-limit'."
   "Non-nil to ignore case when collecting completion candidates."
   :type 'boolean)
 
-(defsubst company-dabbrev-code--make-regexp (prefix)
+(defun company-dabbrev-code--make-regexp (prefix)
   (concat "\\_<" (if (equal prefix "")
                      "\\([a-zA-Z]\\|\\s_\\)"
                    (regexp-quote prefix))
--- a/company-dabbrev.el
+++ b/company-dabbrev.el
@@ -91,7 +91,7 @@ This variable affects both `company-dabbrev' and `company-dabbrev-code'."
               (> (float-time (time-since ,start)) ,limit)
               (throw 'done 'company-time-out))))))
 
-(defsubst company-dabbrev--make-regexp (prefix)
+(defun company-dabbrev--make-regexp (prefix)
   (concat "\\<" (if (equal prefix "")
               company-dabbrev-char-regexp
             (regexp-quote prefix))


dgutov:
Of course.

dgutov:
@PythonNut I'm wondering, how do you deal with the caching problem (the built-in cache filterer is non-fuzzy)? Do you advise `company-dabbrev` to return `t` for `no-cache` as well?

PythonNut:
@dgutov I believe that's what I did. Currently, I don't have a version of `company-dabbrev` that performs well with fuzzy matching. I'm working on it, but I admit that it's currently on the back burner.

dgutov:
> I'm working on it

Any particular design in mind?

PythonNut:
@dgutov here's the code I wrote and abandoned a while back. As far as I know, it works. It just slows everything down just enough to be annoying at times.

```emacs
(defun company-flx-make-regex (prefix sep)
  (if (equal prefix "")
      company-dabbrev-char-regexp
    (if (> (length prefix) 4)
        (mapconcat
         (lambda (x)
           (concat (regexp-quote (string x))
                   sep))
         prefix
         "")
      (concat (regexp-quote prefix)
              sep))))

(defun company-flx/company-dabbrev--make-regexp (prefix)
  (let ((sep (concat "\\(?:"
                     company-dabbrev-char-regexp
                     "\\)*?") ))
    (concat "\\<"
            sep
            (company-flx-make-regex prefix sep)
            "\\>")))

(defun company-flx/company-dabbrev-code--make-regexp (prefix)
  (let ((sep "\\(?:[a-zA-Z0-9]\\|\\s_\\)*?"))
    (concat "\\_<"
            sep
            (company-flx-make-regex prefix sep)
            "\\_>")))

(defun company-flx/company-dabbrev (old-fun command &rest args)
  (if (eq command 'require-match)
      'never
    (apply old-fun command args)))

(advice-add 'company-dabbrev :around #'company-flx/company-dabbrev)
(advice-add 'company-dabbrev-code :around #'company-flx/company-dabbrev)

(advice-add 'company-dabbrev--make-regexp :override
            #'company-flx/company-dabbrev--make-regexp)

(advice-add 'company-dabbrev-code--make-regexp :override
            #'company-flx/company-dabbrev-code--make-regexp)
```

Whenever this does become mature, I'm not sure how to implement it in a package. It looks like I should need to use `with-eval-after-load`, but I've read that that's considered bad practice.

If you have any suggestions, I'm all ears.

dgutov:
> As far as I know, it works.

It doesn't work for `company-dabbrev-code`, because you use the same regexp generation function as for `company-dabbrev` (based on `company-dabbrev-char-regexp`). And for `company-dabbrev`, I had to make these changes:

```el
(defun company-flx-make-regex (prefix sep)
  (if (equal prefix "")
      company-dabbrev-char-regexp
    (mapconcat
     (lambda (x) (regexp-quote (string x)))
     prefix
     sep)))

(defun company-flx/company-dabbrev (old-fun command &rest args)
  (if (eq command 'no-cache)
      t
    (apply old-fun command args)))
```

The result is still wonky, but I don't really use `company-dabbrev` often, so I'm not going to tweak this code. 

Regarding the speed problem (I haven't noticed it, but my laptop is pretty fast, and I haven't tried opening a lot of buffers), `company-flx/company-dabbrev--make-regexp` is most likely at fault: even though you're using `*?`, it still results in way higher complexity than it could. We've previously discussed it in https://github.com/company-mode/company-mode/issues/47#issuecomment-33009601.

Because Emacs regexps don't have lookahead or lookbehind, you can try implement this differently: make `company-flx/company-dabbrev--make-regexp` return a regexp that doesn't depend on the prefix, and collect all words from all buffers within the allowed time limit. Then filter the resulting list with `flx`.

The "collect all words" phase could be sped up later using some smarter caching approaches.

dgutov:
Anyway, thanks for the reply. I thought you already had got it working.

PR #450, opened 2016-01-09
company-capf: Allow custom prefix function

fbergroth:
There is currently no way to have custom logic in the `prefix` command for `company-capf`, such as when you want to return a cons-cell instead of a string. This let's you specify a prefix function using `:company-prefix` in the capf-function.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,7 @@
 
 ## Next
 
+* `company-capf` accepts `:company-prefix-length` property function.
 * New face `company-tooltip-annotation-selection`, used for the annotation in
   the selected tooltip line.
 * `company-clang-objc-templatify` has been renamed to
--- a/company-capf.el
+++ b/company-capf.el
@@ -72,9 +72,15 @@
     (`prefix
      (let ((res (company--capf-data)))
        (when res
-         (if (> (nth 2 res) (point))
-             'stop
-           (buffer-substring-no-properties (nth 1 res) (point))))))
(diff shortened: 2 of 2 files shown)


dgutov:
In order not to return redundant information (which can conflict and lead to confusion), how about the property function (let's call it `:company-prefix-length`) will return only the imaginary prefix length (for lack of a better term). And `company-capf` will construct the cons in its `prefix` handler.

dgutov:
Or we can decide that the property function should only return boolean values, and call it appropriately for that.

fbergroth:
Sure, so if the property function exists and returns a non-nil value `v`, the `prefix` handler returns `("prefix" . v)`, otherwise the handler returns the plain string `"prefix"`?

dgutov:
Pretty much.

fbergroth:
Updated the commit.

dgutov:
Looks good. Please add a NEWS entry (at the top) as well.

Note that while we don't allow `end` to be more than `(point)` now, that should change in the future. Then we'll have to pass `(point)` as the second argument to the new function. Should we make this change now, to make sure no implementations get surprised later?

fbergroth:
That's probably a good idea. I updated the call to `(funcall f beg (point))`, and added a news entry. :-)

dgutov:
Thanks!

dgutov:
@fbergroth What do you think about `:company-prefix-length` being just a value, not a function?

Maybe it's an unnecessary complication.

fbergroth:
Sure, that works for me.

PR #453, opened 2016-01-12
Fix bbdb completion for multi-recipient messages

jtatarik:
Trigger completion for the last incomplete recipient address only,
instead of the whole recipient list.


dgutov:
Thanks!

Issue #454, opened 2016-01-13
Pressing `C-h` isn't always sufficient to show the doc buffer


cpitclaudel:
Hey Dmitry,

Given the following example, pressing <kbd>M-x c-h-twice RET C-h</kbd> shows the doc buffer. Pressing <kbd>C-w</kbd> after that shows the location buffer as expected. However, pressing <kbd>C-h</kbd> again after that doesn't do anything; one needs to press it twince for te doc buffer to appear. IOW, the key sequence to toggle between location buffer and docs buffer is <kbd>C-w C-h C-h C-w C-h C-h C-w C-h C-h</kbd> etc.

```elisp
(defun c-h-twice (command &optional arg &rest ignored)
  (interactive (list 'interactive))
  (pcase command
    (`interactive (company-begin-backend 'c-h-twice))
    (`prefix "")
    (`candidates (list "a" "b" "c"))
    (`doc-buffer (company-doc-buffer "Docs"))
    (`location (cons (company-doc-buffer "Location") 1))))
```

Is this expected? Did I do something wrong? This happens in Emacs 25.

dgutov:
I see it too. It happens, in part, because `C-h` is a poor choice for this command. If you change the binding of `company-show-doc-buffer` in `company-active-map` to, say, `C-d`, switching back and forth will work as expected.

With `f1` used, apparently the `read-key-sequence-vector` call in `company--electric-do` waits for you to press another key instead of returning `[f1]` to be interpreted in the current keymap.

There might be a solution to this, but I'd prefer if we simply picked a different binding for this command. And then somehow made `C-h k` work even when `company-active-map` is active.

Regarding the binding, though, I'm not sure what to pick: `C-d` is easy, but the user might want to use `forward-delete` even when the popup is displayed.

cpitclaudel:
Indeed, changing the binding works :) Is Emacs treating C-h specially?

I do like that binding though; if only because it's easy to memorize. In addition to looking into alternatives, maybe we could look into why Emacs is special-casing C-h?

It would be nice to be able to run C-h k  in company's active map, but that's not excessively useful, insofar as it wont give you all possible keys; meanwhile, the documentation of company-mode does :)

dgutov:
> In addition to looking into alternatives, maybe we could look into why Emacs is special-casing C-h?

I'll take a patch. :)

dgutov:
> Is Emacs treating C-h specially?

[...]

Issue #439, opened 2015-12-28
Inifinite loop when driving company from macros


cpitclaudel:
Hi there,

I'm trying to automatically generate screenshots of company-coq, and it's turning out to be harder than I thought :) I essentially want to automate the task of opening a Coq-mode buffer, setting the right variable, inserting some text, and taking a screenshot.

Everything is working mostly fine for regular examples, but when company is involved things get trickier. Here's a minimal example that runs in emacs -Q, with no need for anything related to `company-coq`:

```elisp
(package-initialize)
(toggle-debug-on-quit)
(global-set-key (kbd "<C-return>") #'company-manual-begin)

(with-current-buffer (get-buffer-create "*LOOP*")
  (set-window-buffer nil (current-buffer))
  (emacs-lisp-mode)
  (company-mode)
  (insert "(def")
  (execute-kbd-macro (kbd "<C-return> C-h")))
```

Running this snippet (e.g. with `emacs -Q --load example.el`) hangs emacs, as the `execute-kbd-macro` call never returns.

Investigating further shows that the `'doc-buffer` command of the backend gets called repeatedly by company: indeed, after waiting for a few seconds and pressing C-g, the *Messages* buffer contains:

```
Type C-x 1 to delete the help window. [279 times]
```

Is there a way to avoid this loop? Or am I doing something wrong?

Thanks!

dgutov:
Hi!

It seems the problem is in `company--electric-do`: it uses `read-key-sequence-vector`, which returns an empty vector during macro execution, instead of waiting for input (dunno why). And we want to wait for the next key, so that as soon as the user calls an unexpected command, the documentation buffer gets hidden.

This works for me:

```el
(with-current-buffer (get-buffer-create "*LOOP*")
  (set-window-buffer nil (current-buffer))
  (emacs-lisp-mode)
  (company-mode)
  (insert "(def")
  (push (aref (kbd "C-h") 0) unread-command-events)
  (push (aref (kbd "<C-return>") 0) unread-command-events))
```

cpitclaudel:
Hi Dmitry,

[...]

Issue #310, opened 2015-02-24
Keep the documentation buffer open if it is clicked


cpitclaudel:
Hi,

It would be neat if clicking the documentation buffer escaped out of `company--electric-do` entirely, and instead persisted the help buffer until it got manually closed by the user (or until something happened in the original buffer? That might be best). Alternatively, an equivalent of `ac-persist-help` would be nice (this probably just requires creating an equivalent of company-show-doc-buffer without the `company--electric-do` wrapper, I could submit a patch).

Thanks!

dgutov:
The first option sounds like a pain to implement. The second one should be fine.

dgutov:
A similar approach to the second one would be to handle the repeat of the same key sequence as what caused the doc buffer to appear (currently just `<f1>`), like you described (make it "persistent" and jump to it). That would require setting up a new keymap inside `company-show-doc-buffer`.

cpitclaudel:
Yes, that could be nice too.

PR #457, opened 2016-01-14
[WIP] Reimplement -electric-do using a pre-command-hook

cpitclaudel:
I wonder if changing the implementation of `-electric-do` would help with some of the things that are currently tricky to fix (some examples: #310, #439, #454).

IIUC, the main point of electric-do is to introduce a temporary window configuration, and restore it once an unrecognised command is issued. Could this be implemented using a `pre-command-hook`? And would that simplify things?

I've made a rough draft to illustrate that idea, which does fix the `C-h` thing.  I'd love to hear your opinion!

### DIFF
--- a/company.el
+++ b/company.el
@@ -1565,6 +1565,7 @@ from the rest of the backends in the group, if any, will be left at the end."
   (and (symbolp command) (get command 'company-keep)))
 
 (defun company-pre-command ()
+  (company--electric-restore-window-configuration)
   (unless (company-keep this-command)
     (condition-case-unless-debug err
         (when company-candidates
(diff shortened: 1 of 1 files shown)


dgutov:
I'm going to miss the cleverness of using `read-key-sequence-vector`, but this looks good, thanks.

Let's just use `company-pre-command` for this, though. No need to add another unconditional element to `pre-command-hook`.

#310 does indeed look closer: we could have a list of commands that don't restore `company--saved-window-configuration` and just erase it.

cpitclaudel:
Cool, I'll clean up the patch soon.
Is it guaranteed that the pre-command-hook will always run before we leave the buffer? I'm worried that the hook being buffer local might mean that in some weird cases the configuration isn't restored. But I guess these weird cases would already have broken the rest of company's code that uses pre-command hooks, right?

cpitclaudel:
(I used this to experiment with the pre-command hook)

```
(defvar my-counter 0)
(add-hook 'pre-command-hook
          (lambda () (message "Hook %S" (cl-incf my-counter)))
          nil t)
```

dgutov:
> Is it guaranteed that the pre-command-hook will always run before we leave the buffer?

I'm not aware of any exceptions, at least. This solution works for mouse clicks, which would be the main question to worry about.

cpitclaudel:
Great, thanks!

cpitclaudel:
I've updated the PR :) Should I remove the `cl-assert`?

dgutov:
> Should I remove the cl-assert?

No need, it shouldn't hurt.

Why not move `company--electric-commands` back where it was, though? And the related new variable and function with it, too. It's okay if `company-pre-command` calls something that's defined after it.

cpitclaudel:
> Why not move company--electric-commands back where it was, though? And the related new variable and function with it, too. It's okay if company-pre-command calls something that's defined after it.

Fixed, thanks!

dgutov:
Thank you :ship:

cpitclaudel:
:)

PR #459, opened 2016-01-20
Turn :company-prefix-length property into a value

fbergroth:
As suggested in #450.


NEWS.md
```diff
@@ -2,6 +2,7 @@
 
 ## Next
 
+* `company-capf` property `:company-prefix-length` is now a value.
```
dgutov: Please edit the original entry instead.

fbergroth: Updated.

dgutov:
:+1:

I wasn't waiting for you to do this, but thanks for taking the initiative!

fbergroth:
My pleasure :smile:

PR #461, opened 2016-01-27
Option to have tempo expand tags after completion.

rlph:
I've added a post-completion call to tempo-expand-if-complete so that tempo tags expand. company-tempo-expand seems like it's supposed to take care of that but it has never worked for me.

### DIFF
--- a/company-tempo.el
+++ b/company-tempo.el
@@ -60,12 +60,11 @@
   "`company-mode' completion backend for tempo."
   (interactive (list 'interactive))
   (cl-case command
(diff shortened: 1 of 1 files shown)


dgutov:
Thanks.

But it seems you should use `company-tempo-insert` instead of `tempo-expand-if-complete`, and remove that argument from the `company-begin-backend` call.

The current code seems to target interactive invocations only (`M-x company-tempo`).

rlph:
I updated it to use `company-tempo-insert`. It works now.

dgutov:
Thanks!

Issue #464, opened 2016-01-31
issues with `company-files`


vspinu:
There are some issues with directory completions on my system:

- the completions of directories are often not proposed with trailing "/". That means you need to type it yourself. 
- when you complete such a directory and type "/" you don't get the completions. The funny part is that if you delete trailing "/" and type it again you start getting the completions. 
- sometimes directories do have trailing "/" and if you complete that directory, company-files is not auto-restarted, so that you need to type initial letters, or delete / and re-type it again.

I am looking into these inconsistencies right now.

dgutov:
> the completions of directories are often not proposed with trailing "/". That means you need to type it yourself

Sometimes you want to type the directory name with a slash, sometimes not. Isn't that right?

The inconsistencies, of course, deserve investigation.

> sometimes directories do have trailing "/" and if you complete that directory, company-files is not auto-restarted

That's more or less by design: if you pick a completion, company-mode remembers the current position as "done", and doesn't try to initiate completion automatically on it. Not to mention that the `this-command` is not `self-insert-command` at that time. Generally, in order not to exit completion, you'd have to use `company-complete-common` there (and even that command could exit completion if there's just one option).

vspinu:
> The inconsistencies, of course, deserve investigation.

I am fixing this. It comes from two distinct paths that company-files--directory-files can take. One returns slashes another one not. 

We can achieve quite some optimization by dropping `file-directory-p` and checking for trailing slash. So I am now always return directories with slash. 

> Sometimes you want to type the directory name with a slash, sometimes not. Isn't that right?

I don't think so. It never matters. All readers are smart enough to recognize a directory irerspective of the slash.

>  Generally, in order not to exit completion, you'd have to use company-complete-common 

Hm. Do you say that it's not possible currently to re-initialize the completion programaticaly? The problem is that if you don't have children, you don't have a common part.

vspinu:
BTW, I don't think this [check](https://github.com/company-mode/company-mode/blob/ada2ede3ec75235ef0a6721e526b2fc94284093e/company-files.el#L59) is a right way to go.

It effectively doubles the completion time. For my remote's home dir with only 40 files this check takes 0.2 seconds. Of course we can try to optimize it out, but I doubt its usefulness in the first place. I am getting loads of junk from dabbrev if it gets through:

![company](https://cloud.githubusercontent.com/assets/1363467/12702217/319baf84-c822-11e5-86ba-c928b602ca99.png)

[...]

Issue #347, opened 2015-05-03
company triggers tramp?


expez:
I was trying to mention a redirect to `/login` in a docstring and then later in a commit message and in both cases a prompt from tramp appeared asking for a password for a host I haven't accessed in years.

Here's the seemingly relevant part of the stacktace:

```
  tramp-completion-file-name-handler(expand-file-name "/login.stud.ntnu.no:" "/home/expez/git/strom/.git/")
  file-directory-p("/login.stud.ntnu.no:")
  company-files--complete("/l")
  company-files(candidates "/l")
```

How can I get back to being able to write `/login` without fear?

dgutov:
Could try try edebug-ging through `company-files--complete`, to see where this completion comes from?

expez:
```elisp
(when (file-directory-p file)
          (push file directories))
```

This triggers it.  Specifically when `file` is `"login.stud.ntnu.no:"` (notice the trailing :).

dgutov:
No, where does the name come from?

expez:
from `company-files--completion-cache`

dgutov:
You can `(setq company-files--completion-cache nil)` to invalidate it, and then try again.

expez:
That didn't do it for me.  The unwanted entry was recreated by `company-files--directory-files` instead.

dgutov:
Right. Where does the host name come from?

Is there a file in `.git` with that name?

expez:
`(file-name-all-completions "/l" "/")` => `(login.stud.ntnu.no:")`

Called by `company-files--directory-files`

If I do `find-file` and navigate to `/` I get a bunch of completion suggestions for various other hosts too.

dgutov:
Thank you, we should handle that case now.

expez:
Thanks! :+1:

PR #465, opened 2016-01-31
[#464] Optimisation and fixes in `company-files`

vspinu:
- Don't rely on `directory-files`; it returns directories without
    trailing slash
  - Don't use `file-directory-p`; check for trailing / instead
  - Don't append children on remotes

### DIFF
--- a/company-files.el
+++ b/company-files.el
@@ -31,7 +31,8 @@
 (defun company-files--directory-files (dir prefix)
   (ignore-errors
     ;; Don't use directory-files. It produces directories without trailing /.
-    (let ((comp (file-name-all-completions prefix dir)))
+    (let ((comp (sort (file-name-all-completions prefix dir)
+                      (lambda (s1 s2) (string-lessp (downcase s1) (downcase s2))))))
       (if (equal prefix "")
           (delete "../" (delete "./" comp))
         comp))))
@@ -92,8 +93,7 @@
                                                  (company-files--directory-files d "")))
                                        directories))))
         (setq company-files--completion-cache
-              (cons key (sort (append children candidates)
-                              #'string-lessp)))))
+              (cons key (append candidates children)))))
     (all-completions prefix
                      (cdr company-files--completion-cache))))
 
--- a/company-files.el
+++ b/company-files.el
@@ -30,9 +30,11 @@
 
 (defun company-files--directory-files (dir prefix)
   (ignore-errors
-    (if (equal prefix "")
-        (directory-files dir nil "\\`[^.]\\|\\`.[^.]")
-      (file-name-all-completions prefix dir))))
+    ;; Don't use directory-files. It produces directories without trailing /.
+    (let ((comp (file-name-all-completions prefix dir)))
+      (if (equal prefix "")
+          (delete "../" (delete "./" comp))
+        comp))))
 
 (defvar company-files--regexps
   (let* ((root (if (eq system-type 'windows-nt)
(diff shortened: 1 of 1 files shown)


vspinu:
I don't have time to fix the cache part right now, but that's not critical anyways. 

If you are unwilling to drop the existing-files check in prefix, then we need to think a bit more of how to optimize it out.

company-files.el
```diff
@@ -61,30 +63,36 @@
...
+                                              (and (company-files--trailing-slash-p f)
+                                                   (company-files--connected-p f)))
+                                            candidates))
+             (children (unless (file-remote-p dir)
+                         (cl-mapcan (lambda (d)
+                                      (mapcar (lambda (c) (concat d c))
+                                              (company-files--directory-files d "")))
+                                    directories))))
+        (setq company-files--completion-cache
+              (cons key (sort (append candidates children)
```
dgutov: Why sort? Were you getting worse results without that, because of using `file-name-all-completions`?

dgutov: Oh. That's because you add children as a separate list? Is is to call `file-remote-p` only once? Does that provide a perceptible performance improvement in some case?

company-files.el
```diff
@@ -61,30 +63,36 @@
...
-                        (unless (eq (aref directory (1- (length directory))) ?/) "/")
-                        child) candidates)))
-      (setq company-files--completion-cache (cons key (nreverse candidates))))
+      (let* ((candidates (mapcar (lambda (f) (concat dir f))
+                                 (company-files--directory-files dir file)))
+             (directories (cl-remove-if-not (lambda (f)
+                                              (and (company-files--trailing-slash-p f)
+                                                   (company-files--connected-p f)))
+                                            candidates))
+             (children (unless (file-remote-p dir)
```
dgutov: Can we be sure that doing `file-remote-p` check only for the parent won't hit the same problem as https://github.com/company-mode/company-mode/issues/347?

vspinu: Good point. It surely won't hit the tramp problem, but it might hit other slow connectors (like archive). See `file-name-handler-alist`. `file-remote-p` is not quite what the name says:). I will run the same check for each parent directory in the list.

company-files.el
vspinu: For two reasons. First, `file-name-all-completions` doesn't return a sorted list. Second, for readability. Sideeffecting in dolist is difficult to reason about.

Sort is pretty cheap as compared to other stuff. It's about 1ms for 5000 files on my machine. So I would not be too concerned about it anyways.

company-files.el
dgutov: > It surely won't hit the tramp problem

Why not? Can't dir be `/`, and one of the files "inside" it, be `/login.stud.ntnu.no:`?

Not sure what you mean about archives, though. UUIC, either `dir` is already inside an archive, in which case it's "remote", or its direct children are "plain" files, and the name of an archive file won't end with `/`, so we wouldn't try to list its contents anyway. But I suppose we could see other kinds of "remote" directories.

company-files.el
dgutov: Ok, thanks.

company-files.el
vspinu: >Why not? Can't dir be /, and one of the files "inside" it, be /login.stud.ntnu.no:?

Ah right. That's the only corner case. 

> But I suppose we could see other kinds of "remote" directories.

Technically it could be a simlink to some slow network drive. So regexp check is not really a full proof solution anyways. The only complete solution that I can see is to remove the children part and re-start the backend after dir insertion. Cannot we do it in `post-completion` command? Looks easy enough.

dgutov: > Technically it could be a simlink to some slow network drive.

Or some drive can simply be mounted over nfs. Indeed, we can't anticipate all slow filesystem operations.

> remove the children part and re-start the backend after dir insertion

Why not remove the trailing slashes, and let the user type them, if they want to continue completion?

> Cannot we do it in post-completion command?

I'm not sure it's a good model of operation.

vspinu: > Why not remove the trailing slashes, and let the user type them 

You mean removing them in post-completion hook? Removing before would cause directories being displayed without trailing / which is confusing. 

I also think inserting directories without trailing / promotes bad style. It's always good to be explicit if a path is a directory or not. 

We can have a custom variable for this. Like `company-files-auto-restart-on-directories`. If `t` we insert / and re-start. If nil we don't insert /. What do you think?  I personally favor of the former. It's faster and more intuitive interface IMO.

company-files.el
dgutov: The previous behavior looks better to me: it listed directories first, and then the lists of their children. If sorting is necessary, you should sort the list of directories and the lists of children separately.

company-files.el
dgutov: > You mean removing them in post-completion hook?

Guess so.

> I also think inserting directories without trailing / promotes bad style. It's always good to be explicit if a path is a directory or not.

The current code does that anyway, and we haven't received many complaints about that.

> We can have a custom variable for this. Like company-files-auto-restart-on-directories.

It's not a question of user preference, but of framework capability. If it can work at all (I'm not 100% sure), that will only be by accident, as a result of 7a05d1cb.

Supporting this usage properly will require some consideration of Emacs's completion tables, which also provide a way to indicate that completion can still continue. Yet, when used through company-capf, completion will similarly end if there is only one candidate currently.

vspinu:
> The previous behavior looks better to me: it listed directories first, and then the lists of their children. 

This is what is happening right now. When sorted directories are arranged in natural order. Previous code was doing the same by merging the lists. If you sort separately you run the risk to have other directories between a dir and its children.

vspinu:
> Supporting this usage properly will require some consideration of Emacs's completion tables, 

I will try it for myself then and will let you know how it goes.

dgutov:
> If you sort separately you run the risk to have other directories between a dir and its children. 

That's the idea. And that's what the existing code does.

That way, you can easily pick and insert any directory with RET, not just the first one in the list.

vspinu:
I have fixed the sorting (in a separate commit this time). Also deleting the trailing / in post-completion. It feels weird at first, but it's one of those things that you quickly learn to anticipate.

dgutov:
Thanks a lot! :+1:

PR #466, opened 2016-02-05
Fix typo in dependency name

tarsius:


dgutov:
Oh, thanks! So it actually still works, 7 years later?

tarsius:
I have no idea, I am just preparing for the "re-announcement of the Emacsmirror" and in the process I investigate unsatisfied dependencies :-)

PR #473, opened 2016-02-14
add explicit option for which standard to use with company clang and make it buffer local

DavidSnider:
Presently company clang is somewhat clunky when it comes to different versions of the standard. in order to make c++11/14 completions work, some variant of the following must be used
```
(setq company-clang-arguments (list "-std=c++14"))
```
This is unfortunate for a few reasons.

1. Figuring this our requires some decent looking into the code of company-clang, which is not preferable.
2. Once the correct variable to edit is found, the fact that it must be a list is not obvious
3. If editing files of multiple types for which different standards are important (such as having a .c and a .cpp file open at the same time) this setup causes problems.

The solution I created is simply to 
a. Create an explicit option for the std to use
b. make it buffer local in order to not have multiple file types open wreck havoc.

With the new code, the way to set the standard would be to put the following in a hook
`(setq company-clang-std "c++14")`

### DIFF
--- a/company-clang.el
+++ b/company-clang.el
@@ -53,6 +53,16 @@ Prefix files (-include ...) can be selected with `company-clang-set-prefix'
 or automatically through a custom `company-clang-prefix-guesser'."
   :type '(repeat (string :tag "Argument")))
 
+(defcustom company-clang-std nil
+  "The language standard to use in Clang.
+
+The value of this variable is either a string denoting a language
+standard, or nil, to use the default standard.  When non-nil,
+pass the language standard via the `-std' option."
+  :type '(choice (const :tag "Default standard" nil)
+                 (string :tag "Language standard")))
+(make-variable-buffer-local 'company-clang-std)
+
 (defcustom company-clang-prefix-guesser 'company-clang-guess-prefix
   "A function to determine the prefix file for the current buffer."
   :type '(function :tag "Guesser function" nil))
@@ -245,6 +255,8 @@ or automatically through a custom `company-clang-prefix-guesser'."
           (unless (company-clang--auto-save-p)
             (list "-x" (company-clang--lang-option)))
           company-clang-arguments
+          (when company-clang-std
+            (list (concat "-std=" company-clang-std)))
           (when (stringp company-clang--prefix)
             (list "-include" (expand-file-name company-clang--prefix)))
           (list "-Xclang" (format "-code-completion-at=%s"


dgutov:
> b. make it buffer local in order to not have multiple file types open wreck havoc.

What does that mean?

Could you add some existing standards as the options for this variable? And add a NEWS entry?

Sarcasm:
> Figuring this our requires some decent looking into the code of company-clang, which is not preferable.

I don't get it. Except for the most trivial example, wouldn't you need more flags? For example to add include directories. Should an option be added for these? Then the next step is compile definition `-D` flags, then something more, ...

flycheck used to have an option for each flag I think, then added a catch-all options. There will always be more flags. Current list of flags: http://www.flycheck.org/manual/latest/Supported-languages.html#C_002fC_002b_002b


> Once the correct variable to edit is found, the fact that it must be a list is not obvious

Company could warn about this eventually.

> b. make it buffer local in order to not have multiple file types open wreck havoc.

I guess allowing `company-clang-arguments` to be local make sense.

dgutov:
> I guess allowing company-clang-arguments to be local make sense.

Do we prohibit it now somehow?

Sarcasm:
> Do we prohibit it now somehow?

No. :)

DavidSnider:
@dgutov 

> > b. make it buffer local in order to not have multiple file types open wreck havoc.

> What does that mean?

Basically, 
> If editing files of multiple types for which different standards are important (such as having a .c and a .cpp file open at the same time) this setup causes problems.

If you have a config for c-mode and a config for cpp-mode, which makes sense under the circumstances, either one will overwrite the over, causing company-clang to work nonsensically in the buffer with the overwritten standard setting, or they will both be added to the arguments which likely just uses the first config. Either way, one of your buffers is a loser. 

> Could you add some existing standards as the options for this variable? And add a NEWS entry?

it doesn't really make sense to add options for this as the user will either already know which option they need, or not need it at all. I don't think it's worth it to list to various c and c++ std options.

I'm new to the scene, can you quickly explain what a NEWS entry is?

@Sarcasm 

That sounds like a slippery slope argument to me. I found that I needed this specific config in order to get the wanted results from company-clang. Specifying a standard is an extremely common thing to do. Many people these days don't want to be using the default stds, which I believe are c89 and c++98/03. I don't often (extremely rarely) find myself needing ot specify include directories and preprocessor definition flags are another beast entirely. Those flags don't just vary by filetype or even necessarily by project. They will sometimes vary at file-level, which makes them nonsensical to have in a config. That being said, this is another argument for company-clang-arguments to be local.

Further regarding the issue of "where does it end", I think it wouldn't be a terrible idea to have specific options for the most commonly used items that make sense to have pre-specified in a config. As you said, they have the catch-all for the less commonly used ones, and company-clang uses a similar method.

dgutov:
> If you have a config for c-mode and a config for cpp-mode, which makes sense under the circumstances, either one will overwrite the over

Wherever you set its value, you'd just have to use `make-local-variable`, if you want it to be buffer-local. I don't think `make-variable-buffer-local` is justified here.

> it doesn't really make sense to add options for this as the user will either already know which option they need, or not need it at all.

Why don't they simply know what to add to `company-clang-arguments`, then?

> I'm new to the scene, can you quickly explain what a NEWS entry is?

There is a `NEWS.md` in this repository.

DavidSnider:
> Wherever you set its value, you'd just have to use make-local-variable, if you want it to be buffer-local. I don't think make-variable-buffer-local is justified here.

> Why don't they simply know what to add to company-clang-arguments, then?

The problem I'm trying to solve here isn't that of people who use this not understanding different standards levels for cc compilation. I think it's fair to assume that the goal of this package is not to help people with that. The problem I found was that the current set-up is very difficult to work with for anyone not decently familiar with elisp, which I would wager is most of the users of this package. 

Company is one of the better packages I use, but it took me way too much fiddling (on the order of 10 hours) to figure out how to make company-clang work correctly. This is 95% due to my level of elisp competency. Being able to setq configs to a constant or a simple add-to-list is what most people I know are competent with. They don't want to have to spend a significant amount of time drudging through the codebase to figure out that they need to set clang-arguments, which is nil, to a list containing a -std= statement. 

Assuming I figured that out, I didn't even know make-local-variable was a thing until exactly this moment. I believe the current setup is asking too much of the userbase. I made this config buffer-local so it didn't negatively interact across modes accidentally. I, as a user, assumed that since the value of the variable should be different across different modes then it was probably buffer-local to avoid cross contamination, and I was very confused and didn't know how to fix it when it became a problem for me.

This request is essentially me trying to fix the problems I ran into as an average user.

dgutov:
> The problem I'm trying to solve here isn't that of people who use this not understanding different standards levels for cc compilation.

You also have to translate the particular standard into a value for the `-std=` argument. I think it's counterproductive to assume that the user knows the appropriate values (and is confident in that), but doesn't know the name of the argument (i.e. can't just as easily set it via `company-clang-arguments`).

> I, as a user, assumed that since the value of the variable should be different across different modes then it was probably buffer-local to avoid cross contamination, and I was very confused and didn't know how to fix it when it became a problem for me.

If it's buffer-local to start with, you'll encounter different kinds of confusion when you actually start programming, and try to set a global value in one place, and then modify it as appropriate when each file is being visited. Better use `make-local-variable`, even if you only found out about it now.

> Company is one of the better packages I use, but it took me way too much fiddling (on the order of 10 hours) to figure out how to make company-clang work correctly.

That seems a bit excessive, but true, `company-clang` is optimized for ease of maintenance more than anything else. I don't however, understand the particular difficulty here: provided you know how to list the customizable variables (`M-x customize group`, mentioned on the home page), you should be able to find the variable in question easily. And its value is documented as well. If the user doesn't know what "Additional arguments to pass to clang when completing" should be, then they're not my kind of audience, and should look into third-party packages like `irony-mode` or `cpputils-cmake` that know how to set up the arguments automatically.

I do agree with @Sarcasm, by the way, that the standard to use is just one piece of the puzzle, and the include paths are just as (if not more) important. At least I've needed them when testing out completion in various examples. And users do set them too, AFAIK.

DavidSnider:
Very well.

PR #483, opened 2016-03-09
add Julia keywords and general types.

stardiviner:


dgutov:
I'm not so sure about the "general types" section; we don't list those for other languages, and they're not keywords.

Other than that, please sort the keywords alphabetically (that's what the build failure is about).

stardiviner:
I sorted keywords with `shell-command-on-region` with `sort`, and deleted julia general types. make a force push.

dgutov:
> make a force push.

I think you forgot to squash.

While we're on the subject of editing commits, please omit the period at the end of the commit title, and start it with a capital letter.

stardiviner:
Those general types is from Julia type system, You can check them out at here: https://en.wikibooks.org/wiki/Introducing_Julia/Types

I don't know whether should consider them as keywords. Just because they are built-in.

dgutov:
> I don't know whether should consider them as keywords. Just because they are built-in.

Probably not.

stardiviner:
I see. Thanks.

dgutov:
So, how about squashing commits together, and editing the commit message?

stardiviner:
@dgutov I updated, squashed two commits, and edited commit message.

dgutov:
Thanks!

PR #487, opened 2016-03-12
Bind non-essential to prevent tramp error

juergenhoetzel:
I use `company-capf` in `eshell`. An error is raised when completing _Tramp_ filenames like 
`
cd /scp:`
`
->
```
Debugger entered--Lisp error: (error "Company: backend company-capf error \"Host name must not match method \"scp\"\" with args (candidates /scp:)")
```
See discussion here:

http://lists.gnu.org/archive/html/tramp-devel/2016-03/msg00005.html

### DIFF
--- a/company-capf.el
+++ b/company-capf.el
@@ -79,7 +79,8 @@
             (length (cons prefix length))
             (t prefix))))))
     (`candidates
-     (let ((res (company--capf-data)))
+     (let* ((non-essential t)
+            (res (company--capf-data)))
        (when res
          (let* ((table (nth 3 res))
                 (pred (plist-get (nthcdr 4 res) :predicate))


dgutov:
Hi!

This seems to be a bug in Emacs: you can type `M-x pcomplete`, and get the same error.

You'll even get this error if the input is just `cd /`.

juergenhoetzel:
@albinus: Any options about fixing this in pcomplete?

albinus:
Jürgen Hötzel <notifications@github.com> writes:

> @albinus: Any options about fixing this in pcomplete?

I've tried to reproduce the bug in pcomplete, but it doesn't work for
me:

# ~/src/emacs-25/src/emacs -Q /

M-x load-library RET pcomplete
M-x pcomplete RET
=> Invalid function: pcomplete-here

Best regards, Michael.

juergenhoetzel:
> Jürgen Hötzel <notifications@github.com> writes:
>  @albinus: Any options about fixing this in pcomplete?
> I've tried to reproduce the bug in pcomplete, but it doesn't work for
> me:
> 
> emacs -Q /
> 
> M-x load-library RET pcomplete
> M-x pcomplete RET
> => Invalid function: pcomplete-here
> 
> Best regards, Michael.

@albinus Hi Michael,

i can reproduce this on emacs master branch, not in the dired pcomplete setup, but eshell:

```
emacs -Q /

M-x eshell RET
```

```
Welcome to the Emacs shell
```
Enter a tramp-filename prefix:
```
/ $ /scp:

```
then

`M-x pcomplete RET`

results in this error:

``` 
tramp-file-name-handler: Host name must not match method "scp"
 ```

albinus:
Jürgen Hötzel <notifications@github.com> writes:

> @albinus Hi Michael,

Hi Juergen,

> i can reproduce this on emacs master branch, not in the dired
> pcomplete setup, but eshell:
>
> emacs -Q /
>
> M-x eshell RET
> Welcome to the Emacs shell
>
> Enter a tramp-filename prefix:
>
> / $ /scp:
>
> then
>
> M-x pcomplete RET
>
> results in this error:
>
> tramp-file-name-handler: Host name must not match method "scp"

Thanks for the recipe, I could reproduce it here as well. Will fix it in
pcomplete, either the emacs-25 branch (frozen already), or in master.

Best regards, Michael.

PR #491, opened 2016-03-20
Fixed lighter for grouped backends

akirakyle:
Fixed issue where if the car of company-backend was a keyword (eg :sorted) the lighter would display that instead of the correct backend. (I believe this necessary due to the "small perf optimization" in company--multi-backend-adapter-candidates


dgutov:
Thanks! Why `(and ... x)`, though? Can `x` ever be nil?

dgutov:
Never mind that; I was being stupid. :smile:

Issue #493, opened 2016-03-29
how to company function name in lua mode?


yinchunlong:
My test step:
1.  (add-hook 'lua-mode-hook 'company-mode)
2. use command : gtags --gtagslabel=ctags to generate tag database.
3. calling company-gtags manually

But it's not working. 
Is there anyone can help me to config company-mode for lua-mode.

dgutov:
Doesn't work how? Shows no completions at all?

stardiviner:
@yinchunlong There is an repo called `company-lua` [https://github.com/ptrv/company-lua](company-lua). But it is not in MELPA yet.

yinchunlong:
@dgutov Only  complete function and keyword which in the Editing file, not through tag database.

yinchunlong:
@stardiviner ok, i will test it. thanks.

dgutov:
@yinchunlong Does ctags parse your project fine? Try using it by itself first (run `ctags -R -e`). Does it generate a `TAGS` file with all your definitions?

yinchunlong:
@dgutov ![image](https://cloud.githubusercontent.com/assets/11457359/14349977/1569880c-fcf9-11e5-94a7-4f45133518ec.png)
it generate a TAGS.

dgutov:
OK. The contents of that file seem problematic, though.

If you type `cla`, do you get `classafterlistener` in completions?

yinchunlong:
when i type cla, i get nothing

dgutov:
After running `gtags --gtagslabel=ctags`, right?

yinchunlong:
![image](https://cloud.githubusercontent.com/assets/11457359/14350191/68dffd1c-fcfa-11e5-9a6b-85061ef8c9c6.png)
still can't complete

dgutov:
All right, thanks. I'll take a look.

yinchunlong:
i use spacemacs in windows, here is my config file:
![qq 20160407200409](https://cloud.githubusercontent.com/assets/11457359/14350455/f57f4fe2-fcfb-11e5-8308-8eb6842cf007.png)
![qq 20160407200420](https://cloud.githubusercontent.com/assets/11457359/14350456/f581913a-fcfb-11e5-935c-7387872bc3d6.png)

then i use command : gtags --gtagslabel=ctags to generate tag database in the root of my project.
then i type any function name in a lua file, it can't complete.
 but i can jump to define, through the tag database.
if i miss something, plese tell me.
[...]

PR #506, opened 2016-04-28
fix company-etags doesn't support for lua-mode

zilongshanren:
fixed:  https://github.com/company-mode/company-mode/issues/493

Because `lua-mode` is 

```
(define-derived-mode lua-mode lua--prog-mode "Lua"
```

And

When using `(apply #'derived-mode-p company-etags-modes)` in lua-mode, it will fails.

### DIFF
--- a/company-etags.el
+++ b/company-etags.el
@@ -55,7 +55,7 @@ Set it to t or to a list of major modes."
   :package-version '(company . "0.9.0"))
 
 (defvar company-etags-modes '(prog-mode c-mode objc-mode c++-mode java-mode
-                              jde-mode pascal-mode perl-mode python-mode))
+                              jde-mode pascal-mode perl-mode python-mode lua-mode))
 
 (defvar-local company-etags-buffer-table 'unknown)
 


redguardtoo:
Looks good,  @dgutov, could you also add the web-mode  into the company-etags-modes?

dgutov:
Bug #493 was about `company-gtags`.

> (define-derived-mode lua-mode lua--prog-mode "Lua"

That seems like a bug in `lua-mode`, they shouldn't do that.

zilongshanren:
@dgutov  
I have tested Lua file with gtags, it seems gtags can't parse Lua file properly...

zilongshanren:
@dgutov  
`lua--prog-mode` is defined like this:

```
  ;; Backward compatibility for Emacsen < 24.1
  (defalias 'lua--prog-mode
    (if (fboundp 'prog-mode) 'prog-mode 'fundamental-mode))

```

dgutov:
> lua--prog-mode is defined like this:

Yes, and that breaks `derived-mode-p`.

dgutov:
`web-mode` is doing the same crap, too. :disappointed: 

https://github.com/fxbois/web-mode/blob/master/web-mode.el#L2254

dgutov:
Here's how to do it right:

https://github.com/emacs-mirror/emacs/blob/8c3245ef0107337d0a2284c58089e2e8e21e257c/lisp/progmodes/verilog-mode.el#L351
or
https://github.com/emacs-mirror/emacs/blob/8c3245ef0107337d0a2284c58089e2e8e21e257c/lisp/progmodes/cc-mode.el

zilongshanren:
@dgutov 
Thanks for pointing out the right direction, I will open issue on `lua-mode`

Issue #432, opened 2015-12-16
duplicates not removed when combining company-tern and company-dabbrev


aaronjensen:
![ret](https://cloud.githubusercontent.com/assets/8588/11832800/613cf5b6-a371-11e5-9f81-704ffeff388d.gif)

When using:

```lisp
(setq-default company-backends-js2-mode '((company-tern :with company-dabbrev)))
```

`ret` autocompletes to `return`, `return`. I would expect that the duplicates would be stripped, am I missing something else?

I'm using spacemacs, here is my config: https://github.com/aaronjensen/dotfiles/blob/master/spacemacs

#### System Info
- OS: darwin
- Emacs: 24.5.1
- Spacemacs: 0.105.0
- Spacemacs branch: develop (rev. 78bdfd0)
- Distribution: spacemacs
- Layers:
```elisp
(better-defaults emacs-lisp markdown syntax-checking auto-completion erlang elixir git dash html org colors spacemacs-layouts osx github javascript react deft floobits ruby semantic unimpaired
                 (shell :variables shell-default-height 30 shell-default-position 'bottom)
                 (spell-checking :variables spell-checking-enable-by-default nil)
                 ranger)
```

stardiviner:
I have this issue too.

dgutov:
In this case, `company-tern` seems to return an empty string for annotation (as opposed to returning nil), so we end up comparing them, find them non-equal and keep both elements. We could fix that by explicitly replacing empty string annotations with nils.

However, is this the only case you find problematic? What about if an item returned by `company-tern` *does* have a non-empty annotation, and `company-dabbrev-code` returns the same method's name?

Can anybody think of a case when we want to keep both an item with annotation, and an item with the same name without annotation, in the combined list?

aaronjensen:
I can't. It'd make sense to prefer the annotation, but ultimately 90% of the time I get an autocomplete dialog I just want to type the thing I was already thinking faster. The annotations don't do much for me--but this could be because most of my recent work has been in more dynamic languages.

stardiviner:
I checked the used company backends with command `company-diag` for `ret` (which returns two `return`):
```
Completions:
  #1=#("return" 0 1 (isProperty :json-false isKeyword t depth 0 name #1#)) ""
  #("return" 0 6 (company-backend company-keywords))
```

dgutov:
> the used company backends

The output is just as I expected.

stardiviner:
[...]

PR #509, opened 2016-05-05
Add option to company-yasnippet whether candidates are merged with other backends

mola-T:
Many people are using this:

    ;; https://github.com/syl20bnr/spacemacs/pull/179
    (defvar company-mode/enable-yas t
      "Enable yasnippet for all backends.")

    (defun company-mode/backend-with-yas (backend)
      (if (or (not company-mode/enable-yas)
              (and (listp backend) (member 'company-yasnippet backend)))
          backend
        (append (if (consp backend) backend (list backend))
                '(:with company-yasnippet))))

    (setq company-backends (mapcar #'company-mode/backend-with-yas company-backends))

I think there should be a choice whether the yas completion should replace the original completion.

### DIFF
--- a/company-yasnippet.el
+++ b/company-yasnippet.el
@@ -35,6 +35,13 @@
 (declare-function yas--template-expand-env "yasnippet")
 (declare-function yas--warning "yasnippet")
 
+(defgroup company-yasnippet nil
+  "Completion backend for yasnippet")
+(defcustom company-yasnippet-merge-candidates t
+  "Whether completion candidates should merge with other backends"
+  :tag "Whether completion candidates should merge with other backends"
+  :group 'company-yasnippet)
+
 (defun company-yasnippet--key-prefixes ()
   ;; Mostly copied from `yas--templates-for-key-at-point'.
   (defvar yas-key-syntaxes)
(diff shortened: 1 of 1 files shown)


dgutov:
Hi!

What will the space change?

mola-T:
@dgutov 
For example, I have a backend setting like this 

    (company-capf :with company-yasnippet)

If I type `def`
`company-capf` provides completion for `defun` while `company-yasnippet` also provides expansion for `defun`. Since both of them use the key `defun`, the duplication is resolved, and only the `company-yasnippet` expansion is provided to the frontend.

By adding the space, `defun` and `defun ` do not count as duplication. So, both of them are provided to the frontend.

dgutov:
Both of them should be displayed either way: Yasnippet completions have their own kind of annotations.

If two completions in the list have the same string value but different annotations, we consider them to be different, and display them together.

mola-T:
em..... But I found that results from `company-yasnippet`  shadow other backends most of the time.

dgutov:
I see, thanks. It seems to be the result of #432. We'll have to revisit that decision.

dgutov:
@mola-T Please clarify: would you personally use `company-yasnippet-merge-candidates` set to t?

mola-T:
I don't have this variable at all.

```elisp
(boundp 'company-yasnippet-merge-candidates)
;; ==> nil
```

mola-T:
Personally, I set `company-yasnippet-merge-candidates` to `nil` so the choices don't get merged.

dgutov:
The latest change should make this pull unnecessary. Please verify.

mola-T:
Yeah. It's working. Thanks!

Issue #246, opened 2014-11-30
Don't Auto-complete When I Don't Mean to


Haroogan:
I would like to keep the following settings everywhere:

```
(set 'company-auto-complete t)
(set 'company-idle-delay 0)
```

However, the problem is that when I type in comments and/or in plain text files I always get auto-completion when I don't need to. For example, typing word `be` and then hitting `SPC` may result in auto-completion of word `because`. This is quite annoying and that's indeed not the behavior I'm looking for.

As I see it, the issue stems from a very simple fact: when popup with completion candidates opens we immediately have the first candidate automatically selected for us and that's why it gets auto-completed according to `company-auto-complete-chars`. How about not selecting anything by default in the popup? Is there any option available for this or if not, then may I hope for the new feature?

By the way, it looks somewhat similar to #216.

dgutov:
@PythonNut Here's an easy modification of the `company-complete` definition that should come close:

```emacs
(defun company-complete-common-or-cycle ()
  (interactive)
  (when (company-manual-begin)
    (if (eq last-command 'company-complete-common-or-cycle)
        (let ((company-selection-wrap-around t))
          (call-interactively 'company-select-next))
      (call-interactively 'company-complete-common))))

(define-key company-active-map [tab] 'company-complete-common-or-cycle)
(define-key company-active-map (kbd "TAB") 'company-complete-common-or-cycle)
```

What may or may not be a drawback: the first time it's called, it always does `company-complete-common`, even if there's no longer common prefix. On the pro side, it's predictable, on the con side, it can waste a keypress. It's not hard to change the definition to not do that: call `company-select-next` anyway if the value returned by `buffer-chars-modified-tick` doesn't change after calling the former function.

> We don't care about the candidates out-of-view that break the common prefix.

This, I think would be stupid. We should care, or it's too easy to miss a definition.

PierreR:
Coming from AC I also have some difficulties to adapt my habit so I have tried the above function. Is it possible to add the last point from @PythonNut for reference ? 
`SPC` always terminates completion and inserts a space. Is there an option to do this ?

dgutov:
@PierreR What exactly does "terminate" mean in that context?

PierreR:
I would like `space` to do what the combination of `return + space` is doing: accept the highlight completion and insert a space.

dgutov:
@PierreR `(setq company-auto-complete t)` will do that.

dgutov:
[...]

Issue #216, opened 2014-10-31
youcomplete me vim behaviour: is possible??


cocodrino:
Hi..I'm interested in know if is possible get a similar behaviour to vim autocompletion..in vim when I type I get a popup displaying the suggestions (but they are not enable yet) then I can press tab and choose the first suggestion, or I can press tab again and select the second suggestion, press again and I going iterating across the popup..when a suggestion is enable automatically is added, so I don't need confirm

This is pretty nice because I dont need move my fingers to the arrow keys and I dont need a key combination for change, is pretty fast...

dgutov:
@cocodrino @alexmurray Please try the code snippet I posted in the neighboring issue: https://github.com/company-mode/company-mode/issues/246#issuecomment-68538735

alexmurray:
@dgutov Yep that seems to work nicely - thanks!

dgutov:
@alexmurray Good. Do you think this command should be added as-is? I.e. first tab expands to the common prefix, the following cycle. Or the first tab should also cycle if there's nothing to expand?

alexmurray:
I think it is fine as it currently is but if you have an implementation of the second one I'd be happy to try it out to see which seems better.

dgutov:
Here:

```emacs
(defun company-complete-common-or-cycle ()
  (interactive)
  (when (company-manual-begin)
    (let ((tick (buffer-chars-modified-tick)))
      (call-interactively 'company-complete-common)
      (when (eq tick (buffer-chars-modified-tick))
        (let ((company-selection-wrap-around t))
          (call-interactively 'company-select-next))))))

(define-key company-active-map [tab] 'company-complete-common-or-cycle)
(define-key company-active-map (kbd "TAB") 'company-complete-common-or-cycle)
```

alexmurray:
It doesn't work for me - tab doesn't appear to do anything in this case...

dgutov:
@alexmurray That's odd, works fine here (at least in emacs-lisp-mode). Are you using the latest MELPA version?

angelic-sedition:
@Sarcasm It seems to me like @sooheon was referring to being able to tab through the original popup suggestions and have them expand not to the fuzzy matching.

@dgutov 
[...]

Issue #123, opened 2014-05-29
Is it possible to show the first candidate inline?


si14:
Here is an example from `auto-complete`:

![autocomplete](https://cloud.githubusercontent.com/assets/316910/3124412/cbc934a6-e77c-11e3-8b03-d84c20cec094.png)

I like AC's "show first candidate immediately and then the list" logic a lot. It would be awesome to replicate this behavior with `company-mode`.

kencoken:
> * If you explicitly want to show the list on a timer but after a delay, someone will have to write a wrapper for `company-pseudo-tooltip-unless-just-one-frontend`

I've now done this in my own branch of company-mode:

https://github.com/kencoken/company-mode/commit/bbbf217643b84471125ddfa7bdc98632656daa4f

This allows an additional delay before showing a tooltip by doing the following:

* set the new `company-idle-delay-tooltip` variable to the desired additional delay
* replace `company-pseudo-tooltip-unless-just-one-frontend` in the frontends list with `company-pseudo-tooltip-unless-just-one-frontend-delay`

Popups then appear `company-idle-delay-tooltip` seconds after the new frontend is triggered (and the inline completion can be set to occur immediately by setting `company-idle-delay` to 0 as normal).

@dgutov – let me know if you think these changes would be useful upstream, as would be happy to issue a PR.

dgutov:
@kencoken Sure, I think this will be useful. Quite a few people asked, after all.

Depending on the size of the resulting patch, you might have to sign the copyright assignment papers first.

But I think we can simplify the solution a bit. Since you're not publishing it separately, maybe adding a new frontend is not necessary: the "tooltip idle delay" will simply control whether the pseudo-tooltip is only displayed only after a delay. This will slightly complicate the `run-with-timer` call, I imagine.

Some renaming suggestions:

`company-idle-delay-tooltip` -> `company-tooltip-idle-delay` or `company-tooltip-extra-delay`.
`company-timer-tooltip-delay` -> `company-tooltip-timer` or `company-tooltip-delay-timer`.

dgutov:
Also: I think you can cancel the new timer in the `pre-command` handler of the frontend, can't you?

PR #524, opened 2016-05-28
Add support for auto-complete like behavior

jojojames:
Added company-ac-frontend to mimic auto-complete's delayed popup.

Added company-ac-complete to mimic (approximation) auto-complete's tab behavior.

Added convenience method company-ac-setup to approximate similar
auto-complete behavior without user intervention.

Notes: (Mostly referencing below issue 123)
- Seemed like the feature was never merged in, I pulled in @kencoken's changes and added some more changes (approximating auto-complete's behavior) to bring it to a finish. The naming change recommendations were done as well as pushing the timer cancellation to the precommand.

- From the issue thread, looks like it was might be better to add the additional functionality to the preexisting frontend. I think splitting the frontend out is cleaner and easier to implement (for me anyways).

- Tried my best to put code where it should be but it could possibly use some improvements.

- Not sure if company-tooltip-visible is needed but I wasn't able to find an easy way to check if the tooltip was currently visible.

- I think the auto-complete behavior might still defer slightly from this but it's a very close approximation and in some ways better (in my opinion).

- I can sign any licensing forms.

In reference to: 
https://github.com/company-mode/company-mode/issues/123

Other similar topics:
https://github.com/company-mode/company-mode/issues/216
https://github.com/company-mode/company-mode/issues/246#issuecomment-68538735

Lots of credit to: 
https://github.com/kencoken/company-mode/commit/bbbf217643b84471125ddfa7bdc98632656daa4f

### DIFF
--- a/company.el
+++ b/company.el
@@ -2859,20 +2859,20 @@ Returns a negative number if the tooltip should be displayed above point."
 Delay is determined by `company-tooltip-idle-delay'."
   (cl-case command
     (pre-command
-     (company-pseudo-tooltip-frontend command)
+     (company-pseudo-tooltip-unless-just-one-frontend command)
      (when company-tooltip-timer
        (cancel-timer company-tooltip-timer)
        (setq company-tooltip-timer nil)))
     (post-command
      (if (or company-tooltip-timer
              (overlayp company-pseudo-tooltip-overlay))
-         (company-pseudo-tooltip-frontend command)
+         (company-pseudo-tooltip-unless-just-one-frontend command)
        (setq company-tooltip-timer
              (run-with-timer company-tooltip-idle-delay nil
                              'company-pseudo-tooltip-frontend-with-delay
                              'post-command))))
     (t
-     (company-pseudo-tooltip-frontend command))))
+     (company-pseudo-tooltip-unless-just-one-frontend command))))
 
 ;;; overlay ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
--- a/company.el
+++ b/company.el
@@ -2074,7 +2074,7 @@ With ARG, move by that many elements."
                  (eq old-tick (buffer-chars-modified-tick)))
         (company-complete-common))))))
 
-(defun company-complete-selection-or-select-next-if-tooltip-visible ()
+(defun company-select-next-if-tooltip-visible-or-complete-selection ()
   "This should be used with `company-pseudo-tooltip-frontend-with-delay'.
 Insert selection if only preview is showing or only one candidate, or select the next candidate."
   (interactive)
--- a/company.el
+++ b/company.el
@@ -2843,18 +2843,22 @@ Returns a negative number if the tooltip should be displayed above point."
 (defun company-pseudo-tooltip-frontend-with-delay (command)
   "`compandy-pseudo-tooltip-frontend', but shown after a delay.
 Delay is determined by `company-tooltip-idle-delay'."
-  (when (and (eq command 'pre-command) company-tooltip-timer)
-    (cancel-timer company-tooltip-timer)
-    (setq company-tooltip-timer nil))
-  (unless (and (eq command 'post-command)
-               (company--show-inline-p))
-    (if (or (not (eq command 'post-command))
-            (overlayp company-pseudo-tooltip-overlay))
-        (company-pseudo-tooltip-frontend command)
-      (setq company-tooltip-timer
-            (run-with-timer company-tooltip-idle-delay nil
-                            'company-pseudo-tooltip-frontend
-                            'post-command)))))
+  (cl-case command
+    (pre-command
+     (company-pseudo-tooltip-frontend command)
+     (when company-tooltip-timer
+       (cancel-timer company-tooltip-timer)
+       (setq company-tooltip-timer nil)))
+    (post-command
+     (if (or company-tooltip-timer
+             (overlayp company-pseudo-tooltip-overlay))
+         (company-pseudo-tooltip-frontend command)
+       (setq company-tooltip-timer
+             (run-with-timer company-tooltip-idle-delay nil
+                             'company-pseudo-tooltip-frontend-with-delay
+                             'post-command))))
+    (t
+     (company-pseudo-tooltip-frontend command))))
 
 ;;; overlay ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
--- a/company.el
+++ b/company.el
@@ -2075,8 +2075,10 @@ With ARG, move by that many elements."
         (company-complete-common))))))
 
 (defun company-select-next-if-tooltip-visible-or-complete-selection ()
-  "This should be used with `company-pseudo-tooltip-frontend-with-delay'.
-Insert selection if only preview is showing or only one candidate, or select the next candidate."
+  "Insert selection if only preview is showing or only one candidate, or select the next candidate.
+When used with with `company-pseudo-tooltip-frontend-with-delay',
+if tooltip is not visible or only one candidate, complete selection,
+else, select-active-regions the next candidate."
   (interactive)
   (if (and (company-tooltip-visible) (> company-candidates-length 1))
       (call-interactively 'company-select-next)
--- a/company.el
+++ b/company.el
@@ -2079,7 +2079,7 @@ With ARG, move by that many elements."
 Insert selection if only preview is showing or only one candidate,
 otherwise select the next candidate."
   (interactive)
-  (if (and (company-tooltip-visible) (> company-candidates-length 1))
+  (if (and (company-tooltip-visible-p) (> company-candidates-length 1))
       (call-interactively 'company-select-next)
     (call-interactively 'company-complete-selection)))
 
@@ -2947,7 +2947,7 @@ Delay is determined by `company-tooltip-idle-delay'."
        (or (eq (company-call-backend 'ignore-case) 'keep-prefix)
            (string-prefix-p company-prefix company-common))))
 
-(defun company-tooltip-visible ()
+(defun company-tooltip-visible-p ()
   "Returns whether the tooltip is visible."
   (when (overlayp company-pseudo-tooltip-overlay)
     (not (overlay-get company-pseudo-tooltip-overlay 'invisible))))
--- a/company.el
+++ b/company.el
@@ -2865,7 +2865,12 @@ Delay is determined by `company-tooltip-idle-delay'."
     (post-command
      (if (or company-tooltip-timer
              (overlayp company-pseudo-tooltip-overlay))
-         (company-pseudo-tooltip-unless-just-one-frontend command)
+         (if (not (memq 'company-preview-frontend company-frontends))
+             (company-pseudo-tooltip-unless-just-one-frontend command)
+           (progn
+             (company-preview-frontend 'pre-command)
+             (company-pseudo-tooltip-unless-just-one-frontend command)
+             (company-preview-frontend 'post-command)))
        (setq company-tooltip-timer
              (run-with-timer company-tooltip-idle-delay nil
                              'company-pseudo-tooltip-unless-just-one-frontend-with-delay
--- a/company.el
+++ b/company.el
@@ -2867,10 +2867,9 @@ Delay is determined by `company-tooltip-idle-delay'."
              (overlayp company-pseudo-tooltip-overlay))
          (if (not (memq 'company-preview-frontend company-frontends))
              (company-pseudo-tooltip-unless-just-one-frontend command)
-           (progn
-             (company-preview-frontend 'pre-command)
-             (company-pseudo-tooltip-unless-just-one-frontend command)
-             (company-preview-frontend 'post-command)))
+           (company-preview-frontend 'pre-command)
+           (company-pseudo-tooltip-unless-just-one-frontend command)
+           (company-preview-frontend 'post-command))
        (setq company-tooltip-timer
              (run-with-timer company-tooltip-idle-delay nil
                              'company-pseudo-tooltip-unless-just-one-frontend-with-delay
--- a/company.el
+++ b/company.el
@@ -2075,10 +2075,9 @@ With ARG, move by that many elements."
         (company-complete-common))))))
 
 (defun company-select-next-if-tooltip-visible-or-complete-selection ()
-  "Insert selection if only preview is showing or only one candidate, or select the next candidate.
-When used with with `company-pseudo-tooltip-unless-just-one-frontend-with-delay',
-if tooltip is not visible or only one candidate, complete selection,
-else, select-active-regions the next candidate."
+  "Insert selection if appropriate, or select the next candidate.
+Insert selection if only preview is showing or only one candidate,
+otherwise select the next candidate."
   (interactive)
   (if (and (company-tooltip-visible) (> company-candidates-length 1))
       (call-interactively 'company-select-next)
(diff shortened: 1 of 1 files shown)


dgutov:
Hi! Thanks for the patch.

Two issues:

* Copyright. Would @kencoken be willing to sign the necessary papers as well?

* Names like `company-ac-frontend` are not good enough, and "behave similarly to auto-complete mode" is not a great description. Basically, I'd like to see descriptive names and docstrings that do not mention `auto-complete-mode`. `company-ac-setup` also seems better suited for the wiki.

On a minor note, please don't end commit message summary with a period, and use the imperative voice. See here for the examples and longer explanation: http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

jojojames:
- Copyright. How should we move forward if @kencoken is not available? I guess implementing the delayed popup code another way is not out of question.
- Other Points. I'll make those changes when I can. What would you suggest for the frontend's name? 

I'll edit the wiki here https://github.com/company-mode/company-mode/wiki/Switching-from-AC and add an entry for company-ac-setup once we hash out all the issues.

dgutov:
> I guess implementing the delayed popup code another way is not out of question.

Right. This code is not hugely essential, so an implementation that looks reasonably different should suffice.

> What would you suggest for the frontend's name?

I don't have a particular name in mind, but I gave some related suggestions in the first issue you've linked to.

jojojames:
I modified company-pseudo-tooltip-frontend and added the timer code there. It hooks onto company-tooltip-idle-delay being non-nil before triggering the timer logic.

Otherwise, the logic seems straightforward. If the timer is non nil by the time the post command is hit, it means the timer has elapsed. If it's nil, that means the user pressed another key during that time period cancelling the timer.

Setting a default non-nil value on company-tooltip-idle-delay does break test company-pseudo-tooltip-does-not-get-displaced (makes sense, since the overlay count will be off in the test's instance now). I'm wondering how you'd update the test in this case. (Setting the default to be nil works fine with me too since that's going to be the standard case anyway.)

    (defun company-pseudo-tooltip-frontend (command)
      "`company-mode' frontend similar to a tooltip but based on overlays."
      (cl-case command
        (pre-command
         (company-pseudo-tooltip-hide-temporarily)
         (when (and company-tooltip-idle-delay company-tooltip-timer)
           (cancel-timer company-tooltip-timer)
           (setq company-tooltip-timer nil)))
        (post-command
         (if (and company-tooltip-idle-delay
                  (not company-tooltip-timer)
                  (not (overlayp company-pseudo-tooltip-overlay)))
             (setq company-tooltip-timer
                   (run-with-timer company-tooltip-idle-delay nil
                                   'company-pseudo-tooltip-frontend
                                   'post-command))
           (unless (when (overlayp company-pseudo-tooltip-overlay)
                     (let* ((ov company-pseudo-tooltip-overlay)
                            (old-height (overlay-get ov 'company-height))
                            (new-height (company--pseudo-tooltip-height)))
                       (and
                        (>= (* old-height new-height) 0)
                        (>= (abs old-height) (abs new-height))
                        (equal (company-pseudo-tooltip-guard)
                               (overlay-get ov 'company-guard)))))
             ;; Redraw needed.
             (company-pseudo-tooltip-show-at-point (point) (length company-prefix))
             (overlay-put company-pseudo-tooltip-overlay
                          'company-guard (company-pseudo-tooltip-guard)))
           (company-pseudo-tooltip-unhide)))
        (hide (company-pseudo-tooltip-hide)
              (setq company-tooltip-offset 0))
        (update (when (overlayp company-pseudo-tooltip-overlay)
                  (company-pseudo-tooltip-edit company-selection)))))

dgutov:
Actually, I like the "separate frontend" solution better.

jojojames:
I think I made most of the changes requested. Let me know if you had any other concerns.

Travis build seems to have failed with the last commit. I'm not sure why that is. (The change seems pretty straightforward for the failing commit)

dgutov:
Thanks, looks good. I think we can consider Ken's contribution to be below the "tiny change" limit now. Please see the copyright assignment instructions here: http://orgmode.org/cgit.cgi/org-mode.git/plain/request-assign-future.txt

> Travis build seems to have failed with the last commit.

It was just Travis being flaky. I've restarted it, and it's green now.

jojojames:
I've received the copyright assignment.

dgutov:
Looking over the code, `company-complete-selection-or-select-next-if-tooltip-visible` is iffy:

* Should it be called `company-select-next-if-tooltip-visible-or-complete-selection` instead?
* The docstring's first sentence should be imperative, short, and meaningful (like the second one here, but shorter).
* It seems like you want the behavior from #526 in the "other" case anyway. Considering it's imperfect, do you still use a command like that?

Also, shouldn't `company-pseudo-tooltip-frontend-with-delay` be wrapping `company-pseudo-tooltip-unless-just-one-frontend` instead? Showing the tooltip for just one completion seems counter-productive.

jojojames:
- I can make the name change to company-select-next-if-tooltip-visible-or-complete-selection as well as tightening up the docstring.

 - I'm not sure what the 'other' case is in this case though I do use YCM when I'm using vim. 

YCM's autocompletion behavior is nice but I think it has its own tradeoffs (namely due to lack of preview completion).

 - I'm using that completion (company-complete-selection-or-select-next-if-tooltip-visible) these days and it's working perfectly for me.

 - You're right about wrapping company-pseudo-tooltip-unless-just-one-frontend instead. I'll get that changed.

dgutov:
> I'm not sure what the 'other' case is in this case though

The "other" case is when the tooltip is visible, and TAB iterates between completions.

With YCM, you can press TAB a few times, and then continue typing. The completion is inserted automatically before that.

In Company, you'd have press RET first. Or are you using `company-auto-complete`, too?

jojojames:
Yeah, I hook company-auto-complete to whether or not the tooltip is visible which ends up with very sane completion behavior in my opinion.

 (defun company-ac-setup ()
    "Sets up company to behave similarly to auto-complete mode."
    (setq company-require-match nil)
    (setq company-auto-complete #'company-tooltip-visible)
    (setq company-frontends '(company-echo-metadata-frontend
                              company-pseudo-tooltip-frontend-with-delay
                              company-preview-frontend))
    (define-key company-active-map [tab]
      'company-complete-selection-or-select-next-if-tooltip-visible)
    (define-key company-active-map (kbd "TAB")
      'company-complete-selection-or-select-next-if-tooltip-visible))

jojojames:
Made the relevant changes.

Let me know if the docstrings or commit messages are not up to par yet. I could probably still use some improvement in this area. Thanks. !

dgutov:
> You're right about wrapping company-pseudo-tooltip-unless-just-one-frontend instead. I'll get that changed.

I think the frontend's name should also be updated. Yes, it's getting long. :)

> I hook company-auto-complete to whether or not the tooltip is visible

So, what happens if you type a word, stop to think for a second, and then press SPC? I don't imagine you'd want the first completion inserted in all such circumstances.

dgutov:
Regarding the docsrting, the first sentence  should be shorter. Using "...if appropriate..." should be fine, as long as you elaborate below. Mentioning `company-pseudo-tooltip-frontend-with-delay` in there also seems unnecessary.

Don't worry about commit messages here, I'm going to squash merge this one.

jojojames:
> I think the frontend's name should also be updated. Yes, it's getting long. :)

You're right. I'll get that changed.

> So, what happens if you type a word, stop to think for a second, and then press SPC? I don't imagine you'd want the first completion inserted in all such circumstances.

I usually just C-g or escape out of the completion and move on. I think that might be a usability problem with company-auto-complete (it's hard not to get unexpected completions when typing fast). I think it's hard to account for all use cases like this one but maybe we can add something smarter here.

> Regarding the docsrting, the first sentence should be shorter. Using "...if appropriate..." should be fine, as long as you elaborate below. Mentioning company-pseudo-tooltip-frontend-with-delay in there also seems unnecessary.

I'll go ahead and improve the docstring further.

dgutov:
> maybe we can add something smarter here

Maybe you can do:

```el
(defun my-company-visible-and-explicit-action-p ()
  (and (company-tooltip-visible)
       (company-explicit-action-p)))

(setq company-auto-complete #'my-company-visible-and-explicit-action-p)
```

By the way, the name of `company-tooltip-visible` should end with `-p`, too.

jojojames:
> (defun my-company-visible-and-explicit-action-p ()
  (and (company-tooltip-visible)
       (company-explicit-action-p)))

I'll give that a shot and see.

> By the way, the name of company-tooltip-visible should end with -p, too.

Ok.


Also noticed a small bug with the popup using this delay frontend. The popup's offset is pushed to the right by 30-40 pixels instead of lining up with the prefix. I'll take a closer look at how the popup is drawn later but a step in the right direction here would be great.

dgutov:
> The popup's offset is pushed to the right by 30-40 pixels instead of lining up with the prefix

Is it an offset the width of the prefix? Or of the overlay added by the preview?

jojojames:
Not too sure, attaching a screenshot.

screenshot 2016-06-13 16 23 02

dgutov:
"overlay added by the preview" looks like the correct guess. That's probably because the preview overlay is visible, and it makes the display engine think the current displayed position is at the end of it (similar to behavior described in http://debbugs.gnu.org/cgi/bugreport.cgi?bug=18417).

The way to go about it would be to make sure the preview overlay is not visible when we're calling `company-pseudo-tooltip-show-at-point`. Maybe hide all backends and then show them again, this time including the delayed tooltip? Not sure if it's easy to program cleanly.

jojojames:
I made the other updates.

> Maybe hide all backends and then show them again, this time including the delayed tooltip? Not sure if it's easy to program cleanly.

I was able to get some working functionality in this direction (definitely messily written). Not sure if this is directionally right or clean enough but the functionality is definitely working as far as I've tested.

    ;; post command inside company-pseudo-tooltip-unless-just-one-frontend-with-delay
    (post-command
     (if (or company-tooltip-timer
             (overlayp company-pseudo-tooltip-overlay))
         (progn
           (company-test-hide) ;; HIDE
           (company-pseudo-tooltip-unless-just-one-frontend command)
           (company-test-show)) ;; SHOW
       (setq company-tooltip-timer
             (run-with-timer company-tooltip-idle-delay nil
                             'company-pseudo-tooltip-unless-just-one-frontend-with-delay
                             'post-command))))

    (defun company-test-hide ()
      (cl-loop for frontend in company-frontends
               collect
               ;; todo account for the all the preview frontends
               (when (eq 'company-preview-frontend frontend)
                 (funcall frontend 'pre-command))))
    
    (defun company-test-show ()
      (cl-loop for frontend in company-frontends
               collect
               ;; ^todo
               (when (eq 'company-preview-frontend frontend)
                 (funcall frontend 'post-command))))

dgutov:
If you're just checking for one frontend, this seems much shorter:

```el
(if (not (memq 'company-preview-frontend company-frontends))
    (company-pseudo-tooltip-unless-just-one-frontend command)
  (company-preview-frontend 'pre-command)
  (company-pseudo-tooltip-unless-just-one-frontend command)
  (company-preview-frontend 'post-command))
```

jojojames:
Works like a charm for me.

dgutov:
No need for `progn` there: the "else" branch can have multiple statements.

jojojames:
Thanks! Let me know if there's anything else.

dgutov:
Thanks for sticking with it!

jojojames:
My pleasure. :)

dgutov:
> I'll give that a shot and see.

Have you tried that yet? Or is that drawback not much of a problem?

jojojames:
I tried it and it works pretty well.

My only minor quip about it is that I'd like to consider typing after the popup shows to be an explicit action.
That way the scenario where you type something like.

compa{cursor}

Available completions:
company-preview-frontend
company-select-next

When the completion tooltip is up, if I type all the way to company-s, I'd expect/want a space to be able to complete the rest. Maybe the explicit action is that I changed the prefix. I think the heuristic on determining when to complete is a little tricky so haven't noodled too much on it.

I think defining extra actions that can be considered explicit after the tooltip is up might solve my minor problem here.

dgutov:
I don't want to change the meaning of "explicit action", but if you find a way to program that predicate, you can use it either way.

The idea kind of makes sense, but this way you'll also see scenarios where you are forced to press `C-g` to cancel. So I'm torn.

Looking at auto-complete, it doesn't actually do that: you can wait until the popup shows, type more to narrow the completions down, but when you press SPC, none will be inserted. It does, however, insert the first option when you press TAB.

Have you tried using `company-complete-common-or-cycle` instead of `company-select-next-if-tooltip-visible-or-complete-selection`? Or combine their logic? I think the "complete up to the common part first" step, even when the tooltip is visible, would be more faithful.

jojojames:
I've used company-complete-common-or-cycle before, I think the main problem with it was seemingly having to tab more often in the general case. I haven't tried it with the new delay frontend but potentially something like company-complete-common-or-cycle-if-tooltip-visible-or-complete-selection might be a good solution. I think I'll play around with those options a little more and report back after.

jojojames:
Trying something like this for my tab completion.

    (defun company-complete-common-or-cycle-if-tooltip-visible-or-complete-selection ()
      "Complete common or cycle if appropriate, or complete selection.
    Insert selection if only preview is showing or only one candidate,
    otherwise complete common or cycle."
      (interactive)
      (if (and (company-tooltip-visible-p) (> company-candidates-length 1))
          (call-interactively 'company-complete-common-or-cycle)
        (call-interactively 'company-complete-selection)))

It works fine aside from the somewhat confusing behavior of company-complete-common-or-cycle which completes up to a common prefix (expected but still confusing). Sometimes that causes behavior like:

Start typing, completion popup comes up -> press tab (expecting tab to select next candidate) -> get tab to first selection. I think it's a little too inconsistent (from a user trying to complete selections fast without thinking).

I did end up switching to using 

    (defun my-company-visible-and-explicit-action-p ()
      (and (company-tooltip-visible-p)
           (company-explicit-action-p)))

as it does get rid of that c-g edge case on first candidate.

At this point it's been a while since I've used Auto-Complete so it might be a little off faithfully but I think the current behavior at this point is more than satisfying.

dgutov:
> the current behavior at this point is more than satisfying

Great! Thanks for reporting.

dgutov:
> I did end up switching to using...

Should we add that to " Switching from AC" as well?

jojojames:
Done.

sickHamm:
@jojojames  @dgutov 

Is `company-cancel` shoud remove `company-tooltip-timer` ?

And for https://github.com/company-mode/company-mode/pull/524/commits/69687f2bac976bef81885c0fd766970bfddca591 shouldn't presume is `company-preview-frontend` preview  candidate, should use a more general way.
```
(defun company-pseudo-tooltip-unless-just-one-frontend-with-delay (command)
  "`compandy-pseudo-tooltip-frontend', but shown after a delay.
Delay is determined by `company-tooltip-idle-delay'."
  (cl-case command
    (pre-command
     (company-pseudo-tooltip-unless-just-one-frontend command)
     (when company-tooltip-timer
       (cancel-timer company-tooltip-timer)
       (setq company-tooltip-timer nil)))
    (post-command
     (if (overlayp company-pseudo-tooltip-overlay)
	 (company-pseudo-tooltip-unless-just-one-frontend command)
       (if company-tooltip-timer
	   (if (not (overlayp company-preview-overlay))
	       (progn (company-pseudo-tooltip-unless-just-one-frontend command)
		      (setq company-tooltip-timer nil))
	     (company-preview-hide)
	     (company-pseudo-tooltip-unless-just-one-frontend command)
	     (setq company-tooltip-timer nil)
	     (company-call-frontends 'post-command))
	 (setq company-tooltip-timer
	       (run-with-timer company-tooltip-idle-delay nil
			       'company-pseudo-tooltip-unless-just-one-frontend-with-delay
			       'post-command)))))
    (t
     (company-pseudo-tooltip-unless-just-one-frontend command))))
```

jojojames:
Are you having a problem with another preview frontend?

PR #531, opened 2016-06-04
New transformer: company-sort-prefer-same-case-prefix

juergenhoetzel:
As discussed in [this F# mode issue](https://github.com/fsharp/emacs-fsharp-mode/pull/86),
this transformer might be useful for backends returning case insensitive candidates like
 [FsAutoComplete](https://github.com/fsharp/FsAutoComplete), because users might prefer the exact prefix match in this case:

```elisp
(let ((sorted-candidates '("AAAa" "aaba" "aaca"))
      (company-prefix "aa"))
  (company-sort-prefer-same-case-prefix sorted-candidates ))
("aaba" "aaca" "AAAa")
```

### DIFF
--- a/company.el
+++ b/company.el
@@ -465,6 +465,8 @@ without duplicates."
           (const :tag "Sort by occurrence" (company-sort-by-occurrence))
           (const :tag "Sort by backend importance"
                  (company-sort-by-backend-importance))
+          (const :tag "Prefer case sensitive prefix"
+                 (company-sort-prefer-same-case-prefix))
           (repeat :tag "User defined" (function))))
 
 (defcustom company-completion-started-hook nil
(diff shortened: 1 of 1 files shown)


company.el
```diff
@@ -1331,6 +1333,17 @@ from the rest of the backends in the group, if any, will be left at the end."
                  (let ((b1 (get-text-property 0 'company-backend c1)))
                    (or (not b1) (not (memq b1 low-priority)))))))))))
 
+(defun company-sort-prefer-same-case-prefix (candidates)
+    "Prefer CANDIDATES with the same case sensitive prefix.
+If a backend returns case insensitive matches, candidates with the an exact
+prefix match will be prioritized even if this changes the lexical order."
+  (cl-loop for candidate in candidates
+	   if (equal company-prefix
+                     (substring candidate 0 (length company-prefix)))
```
dgutov: Can we use `string-prefix-p` here?

company.el
```diff
@@ -1331,6 +1333,17 @@ from the rest of the backends in the group, if any, will be left at the end."
...
+	   collect candidate into same-case
+	   else collect candidate  into other-case
```
dgutov: Extra space.

juergenhoetzel:
Good catch!

I just amended the commit.

dgutov:
Thanks. Please also untabify.

How did you get tabs there? dir-locals.el has `(indent-tabs-mode . nil)`.

juergenhoetzel:
> How did you get tabs there? dir-locals.el has (indent-tabs-mode . nil).

Sorry: The code was copied over from a local `fsharp-mode`  branch. So the per-directory local variables for indentation had no effect in this case. 

I just amended the commit.

dgutov:
Thanks!

Issue #150, opened 2014-07-12
company-mode-clang: not completing when the whole function name is typed


xuhdev:
For the following code:

``` c
#include <stdio.h>
int main(void)
{
    print
           ^
}
```

If I type `<M-X>company-complete`, it automatically completes `printf`, including the parameters, which is perfect.

However, if the code is:

``` c
#include <stdio.h>
int main(void)
{
    printf
            ^
}
```

It won't complete if I type `<M-X>company-complete`.

dgutov:
Please try the `forced-completion-for-unique-match` branch. `M-x company-complete` should work there in this situation.

PR #537, opened 2016-06-07
Make company-nxml backend more schema aware

tplass:
With a RELAX NG schema loaded for the current buffer completion is more 
context sensitive and schema aware.

When in tag name context (after a '<'), requesting completion will
- auto-insert the next required element or default to presenting the candidate 
  list of optional elements along with the end tag for the current element
- auto-insert all required attributes
- prepare the next completion request to either
  - auto-insert the sole optional attribute or 
  - present the candidate list of optional attribute names. 
  
  This requires that the last completion action inserted a tag name but does
  not require a space to be manually inserted after the tag name.

Attributes are always inserted as attname="" with point positioned after the first quote.

Attribute completion is available at all legal positions in open as well as already
closed tags.

### DIFF
--- a/company-nxml.el
+++ b/company-nxml.el
@@ -139,9 +139,11 @@ This is based on the premise that a short name indicates significance.")
                                          (company-nxml-attribute-completions
                                           name (rng-match-possible-attribute-names))))))
                   (when candidates
-                    (setq company-nxml-last-command (if arg
-                                                        'company-nxml-attribute
-                                                        'company-nxml-attribute-from-tag))
+                    (if arg
+                        (setq company-nxml-last-command 'company-nxml-attribute)
+                      (setq company-nxml-last-command 'company-nxml-attribute-from-tag
+                            company-prefix "")) ; prevent tag name from acting as a prefix
+
                     candidates)))
     (sorted t)))
 
@@ -221,8 +223,8 @@ Called post completion when inserting a tag name."
     (candidates (cond
                  ((and (eq company-nxml-last-command 'company-nxml-tag)
                        (company-grab company-nxml-in-starttag-name-regexp 1 1))
-                  (setq company-prefix "") ; don't zap the just inserted tag name!
-                  (company-nxml-attribute 'candidates))
+                  (or (company-nxml-attribute 'candidates)
+                      (company-nxml-tag 'candidates arg)))
                  ((company-nxml-tag 'prefix)
                   (company-nxml-tag 'candidates arg))
                  ((company-nxml-attribute 'prefix)
--- a/company-nxml.el
+++ b/company-nxml.el
@@ -143,7 +143,6 @@ This is based on the premise that a short name indicates significance.")
                         (setq company-nxml-last-command 'company-nxml-attribute)
                       (setq company-nxml-last-command 'company-nxml-attribute-from-tag
                             company-prefix "")) ; prevent tag name from acting as a prefix
-
                     candidates)))
     (sorted t)))
 
@@ -169,7 +168,8 @@ Called post completion when inserting a tag name."
         (message "Auto-inserted %srequired attribute%s for element %s"
                  (or (and (cdr atts) (format "%d " (length atts))) "")
                  (or (and (cdr atts) "s") (concat " " (cdr (car atts))))
-                 tagname)))))
+                 tagname)))
+    atts))
 
 
 (defun company-nxml-attribute-completions (prefix alist)
@@ -217,8 +217,8 @@ Called post completion when inserting a tag name."
   (interactive (list 'interactive))
   (cl-case command
     (interactive (company-begin-backend 'company-nxml))
-    (prefix (or (company-nxml-tag 'prefix)
-                (company-nxml-attribute 'prefix)
+    (prefix (or (company-nxml-attribute 'prefix)
+                (company-nxml-tag 'prefix)
                 (company-nxml-attribute-value 'prefix)))
     (candidates (cond
                  ((and (eq company-nxml-last-command 'company-nxml-tag)
@@ -234,7 +234,10 @@ Called post completion when inserting a tag name."
                         'string<))))
     (post-completion (cond
                       ((eq company-nxml-last-command 'company-nxml-tag)
-                       (company-nxml-add-required-attributes arg (point)))
+                       (if (eq (aref arg 0) ?/)
+                           (insert ">")
+                         (or (company-nxml-add-required-attributes arg (point))
+                             (signal 'quit nil)))) ; this makes multiple attribute candidates work
                       ((eq company-nxml-last-command 'company-nxml-attribute-from-tag)
                        (backward-char (length arg))
                        (insert " ")


company-nxml.el
```diff
@@ -129,13 +219,29 @@
                 (company-nxml-attribute 'prefix)
                 (company-nxml-attribute-value 'prefix)))
     (candidates (cond
+                 ((and (eq company-nxml-last-command 'company-nxml-tag)
+                       (company-grab company-nxml-in-starttag-name-regexp 1 1))
+                  (setq company-prefix "") ; don't zap the just inserted tag name!
```
dgutov: This looks odd. Backends aren't supposed to modify `company-prefix`.

And do you absolutely require `company-nxml-last-command` variable? Is it impossible to distinguish via regexps, like the current code does, or by looking around?

We don't guarantee that the `prefix` command is always followed by the corresponding `candidates` command. Imagine e.g. `company-sort-by-occurrence` in combination with `company-search-toggle-filtering`.

dgutov:
Thanks.

Did you take a look what at would it take to bring these features in nXML in Emacs?

tplass:
The `company-nxml-last-command` variable is actually a design choice.

My intention for enhancing company-nxml was to combine as much user support as possible with the way XML works.  In XML, completion is not a series of isolated events.  Rather, it is something that depends not just on _where_ it is about to happen but also _when_ it is to happen.  That's why I'd like Emacs to look over my shoulder and anticipate what needs doing.  That means that the completion code needs to keep some sort of state in the form of the `company-nxml-last-command` variable.

Example: When the users adds a start tag, the next thing he wants to do is to add the attributes for the selected element type.  I find it unnecessary and clumsy to first insert a space character after the tag name has just been completion-inserted as a prerequisite for attribute name completion (for otherwise the tag name acts as a prefix for attributes which it is obviously not).  Even more so when the element has required attributes that absolutely must be added to the start tag.

Worse yet, if the element has no declared attributes, then that completion-requiring space was inserted in vain, clutters up my start tag and might need to be deleted/undone manually as well!


Consider adding the `` tag which has a mandatory `src` attribute  (`^` denotes point)

```xml
<p>See the picture <
                    ^get completion candidates for tag names

<p>See the picture <img
                       ^having selected "img" from list, immediately request completion again
                        for attributes (note: do not insert a space)

<p>See the picture <img src=""
                             ^ src="" was auto-inserted, no need to select from candidates
```

Compare element `<div>` with only optional attributes:

```xml
<body><
       ^get completion candidates for tag names

<body><div
          ^having selected "div" from list, immediately request completion again
           for attributes (note: do not insert a space)
    
<body><div class=""
                  ^ user selected "class" from candidate list 
```

As for setting `company-prefix`: I found that when I requested attribute name completion in the above scenarios, rather than being added to the buffer, the selected candidate would _replace_ the tag name I had just inserted!  It seemed that the tag name acted like a prefix.  I didn't really understand the workings behind this but found that a little `setq` brute force helped ;-)  By all means, if there's a better way to prevent the tag name from being replaced by the attribute name, then I'll use that.  Please advise.


So, adding my features to the new style completion in nxml-mode (and company-nxml as well?) would depend on whether the state variable approach is adopted or not.  Also, the new completion does auto-insert requested attributes, but only one, not all of them.  Is that a bug or a design choice?

tplass:
I need to amend my description a bit as the code is actually more advanced than my representation.

The `` example with the `src` attribute as described would apply only when the attribute is *optional* and is the *sole attribute*.  Only then is a repeated completion request already primed.

When `` has a required attribute `src` (as in real HTML world), then the attribute would be auto-inserted as a result of selecting "img" from the candidate list.  No completion request necessary.

dgutov:
> The company-nxml-last-command variable is actually a design choice.

I'm saying it could be a problem. But I think that information should be fetch-able by going back to the beginning of the current symbol and re-parsing.

> the next thing he wants to do is to add the attributes for the selected element type

This is something we normally use snippets for. Like Yasnippet or the built-in `company-template`. I'm not sure that kind of functionality can naturally fit `company-mode` in its current state.

> Worse yet, if the element has no declared attributes, then that completion-requiring space was inserted in vain

What completion-requiring space? Are you referring to the behavior in Emacs 26?

> I found that when I requested attribute name completion in the above scenarios, rather than being added to the buffer, the selected candidate would replace the tag name I had just inserted!

Is that a bug in the current code, or was it something you had to work around only in the new code?

dgutov:
> Also, the new completion does auto-insert requested attributes, but only one, not all of them. Is that a bug or a design choice?

From your description, it sounds reasonable, but honestly, I have no idea. I never really touched nXML code myself, and I don't use it with any regularity. You should ask that question on emacs-devel.

tplass:
> I'm saying it could be a problem.

What kind of problem?

> But I think that information should be fetch-able by going back
> to the beginning of the current symbol and re-parsing.

The problem with just looking around tells you where you are but
it doesn't tell you what to do next because you don't know what has
happend before.  Here's the case:

```xml
<P><
    ^ user requests cnmpletion: as per schema valid candidates are:
      B
      BUM
      BUMMER
```

User selects `BUM`.  `BUM` has an optional attribute `FROB`.

```xml
<P><BUM
       ^ user requests completion: now what?

```

With looking for regexp only, completion takes a wild guess: Hey, I see `BUM`
which is a nice prefix for `BUMMER`, so I'll present user with an
element choice.

Which is utterly wrong!  Why?  Because user just made a clear decision
to insert element `BUM`, *not* `BUMMER` which was also on the menu but
didn't get selected for a reason.  So why pester him with a different
element name?

With a state variable, completion knows that the user inserted the
element name he wanted and now is likely to want to add attributes.  This is
the right thing to do.


>     the next thing he wants to do is to add the attributes for the selected element type
>    
> This is something we normally use snippets for.

I'm sure there is a place for using templates when editing XML.  But
not as a workaround for limitations and quirks of completion.

> What completion-requiring space? Are you referring to the behavior in Emacs 26?

The space that disambiguates the situation with `BUM` and `BUMMER`
above.  If I insert space after `<BUM` (ie `<BUM `), then completion
won't erroneously recognise `BUM` as a prefix for `BUMMER` but will
say: Right, the only thing possible here is an attribute.

But as I said before: I don't want to have to insert that space.

> Is that a bug in the current code, or was it something you had to work around only in the new
> code?

I think that's the way company-prefix works irrespective of version.

dgutov:
> What kind of problem?

`prefix` is supposed to be a non-stateful command. It can be called at multiple buffer positions during the processing of one user command. Then your variable's value can become wrong.

> I'm sure there is a place for using templates when editing XML. But
not as a workaround for limitations and quirks of completion.

Have you tried any Company backends that use templates? Like `company-clang`, for example?

> I think that's the way company-prefix works irrespective of version.

company-prefix is just a variable. Do you have a problem scenario with the current `company-nxml`?

tplass:
> prefix is supposed to be a non-stateful command. It can be called at multiple buffer positions
> during the processing of one user command. Then your variable's value can become wrong.

It's `candidates` my variable is set, not `prefix`.  And it's not
clear to me how anyone else can corrupt it.  Have you actually run the
patch?

Never mind, I get the feeling there's no point in further arguing.

If you wish, close this pull request.

dgutov:
> It's candidates my variable is set, not prefix.

OK, that's better.

> Have you actually run the patch?

I did not. Even if I did, though, it's hard for me to exercise it in all possible way that can cause a problem. So I try to understand it first.

dgutov:
Speaking of trying it out, I like the automatic addition of `=""`.

I couldn't find an element with required attributes in HTML quickly, so I didn't test that.

Here's something that's broken in the new version:

I can type `<` inside `<head>...</head>` and initiate completion, and `meta` will be one of the candidates. But if I type `<m` and do the same, it says `No completions found`. If I type it out in full, then I get completion for its attributes. Then I can backspace over the last three characters, initiate completion, and *then* completion for `<m` works.

It's the "last command" trickery at fault here, isn't it?

tplass:
> I couldn't find an element with required attributes in HTML quickly, so I didn't test that.

You're on the right track.  If you load the XHTML schema (e.g. from
the menubar), then `img` should complete on two required attributes.
Also many elements share a lot of common attribtes.

> Here's something that's broken in the new version:

It is.  Thanks for testing!

I checked in changes that on the one hand fix the issue of
non-completion of manually inserted partial tag names.  But on the
other hand my new completion for attributes seems broken (and was so
all along).  I think I need your knowledge of the engine code here.

The problem is this: when a tag name has been completed, requesting
completion immediately again will insert an attribute even when there
is a list of attributes to choose from.  No tooltip menu is presented.

To see this in action, create a fresh nXML mode buffer and load the
XHTML schema:

```xml
<h
  ^ will complete to "html"
```

```xml
<html
     ^ requesting completion again will auto-insert dir="" 
       (instead of a popping up a menu)
```

Compare this with the situation where a space is inserted after `html`

```xml
<html 
      ^ requesting completion will popup a menu with three candidates
```

I've tried to track this difference in behaviour down in company.el
but have failed so far.  The value of `company-candiates` is the same
in both cases but somewhere along the line, the list isn't translated
into a menu but it's `car` seems to be taken as the single available
candidate.

My hunch is that cheekily `setq`ing `completion-prefix` to "" is
responsible for this.

Could you give me a hint where I should look for the function that
decides whether a tooltip menu needs to be created and displayed?

dgutov:
> The value of company-candiates is the same in both cases but somewhere along the line, the list isn't translated into a menu but it's car seems to be taken as the single available candidate.

That sounds rather unlikely.

> Could you give me a hint where I should look for the function that
decides whether a tooltip menu needs to be created and displayed?

`company-post-command`, `company--should-complete`, `company--begin-new`, `company-calculate-candidates`.

tplass:
This is one of those things that only work under the debugger.

> company-post-command, company--should-complete, company--begin-new, company-calculate-candidates.

I single-stepped through all of them and as luck would have it, the
tooltip is created correctly every time.  Re-evaluate the functions
and the faulty behaviour returns :-(

Are there any strategies to deal with the debugger-makes-it-work
situation?

dgutov:
Print-debugging? :)

You can add `message` calls to output various information.

tplass:
Took a while to get back but the situation is somewhat frustrating.

I added a `message` to one of the `just-one` frontend functions outputting the current values of the set of buffer globals discussed in the code of `company.el`:

```elisp
(defun company-pseudo-tooltip-unless-just-one-frontend (command)
  "`company-pseudo-tooltip-frontend', but not shown for single candidates."
  (message "pseudo-just-one: %s"
           (mapconcat
            (lambda (sym)
              (concat sym "=" (prin1-to-string (symbol-value (intern (concat "company-" sym))))))
            '("prefix" "candidates" "common" "selection" "point" "search-string")   
            ", "))
  (unless (and (eq command 'post-command)
               (company--show-inline-p))
    (company-pseudo-tooltip-frontend command)))
```



The editing situation in the nXML buffer is this:

```xml
<head
     ^  situation 1: no space after head
      ^ situation 2: space after head
```

situation 1: no space after `head`, completion request -> no menu shown, `dir=""` (== `(car company-candidates)` is inserted

```message
pseudo-just-one: prefix="", candidates=("dir" "lang" "profile"), common="", selection=0, point=nil, search-string=""
pseudo-just-one: prefix="", candidates=("dir" "lang" "profile"), common="", selection=0, point=61, search-string=""
pseudo-just-one: prefix=nil, candidates=nil, common=nil, selection=0, point=nil, search-string=""
```

situation 1: no space after `head`, "do something interactively, like switch buffer and back", then completion request -> menu shown

```message
pseudo-just-one: prefix="", candidates=("dir" "lang" "profile"), common="", selection=0, point=nil, search-string=""
pseudo-just-one: prefix="", candidates=("dir" "lang" "profile"), common="", selection=0, point=61, search-string="" [3 times]
pseudo-just-one: prefix=nil, candidates=nil, common=nil, selection=0, point=nil, search-string=""
```

situation 2: space after `head`, completion request -> menu shown

```message
pseudo-just-one: prefix="", candidates=("dir" "lang" "profile"), common="", selection=0, point=nil, search-string=""
pseudo-just-one: prefix="", candidates=("dir" "lang" "profile"), common="", selection=0, point=62, search-string="" [3 times]
pseudo-just-one: prefix=nil, candidates=nil, common=nil, selection=0, point=nil, search-string=""
```

The only difference I can see, apart from the obviously differing buffer positions, is that the `just-one` function is called 3 times (scroll right into view if necessary) when the menu is shown but only 2 times when `dir=""` is auto-inserted without the menu.  Note that in particular that `company-candidates` is identical in both situations.  So the menu/tooltip/overlay should always be created and shown.

Does the fact that doing "something random" interactively (like switching buffers) before requesting completion in situation 1 ring a bell?  Why would that have any bearing on how completion behaves?  At least it explains why things work under `edebug`.

dgutov:
Without digging into it closely, I could guess that it's the modification of `company-prefix` (which, like I said, is a bad idea), bites you in the ass here.

Like, in one scenario, the value set by the backend holds, and in the other scenario Company's internals overwrite it right away.

In any case, when point is right after `head` (without space), having the prefix be an empty string doesn't make sense to me.

>  Note that in particular that company-candidates is identical in both situations. So the menu/tooltip/overlay should always be created and shown.

The value of prefix has as much effect as the candidates list on whether completion will continue or not. See e.g. `company--continue`, and in particular this comparison:

```el
                       (= (- (point) (length new-prefix))
                          (- company-point (length company-prefix)))
```

> Does the fact that doing "something random" interactively (like switching buffers) before requesting completion in situation 1 ring a bell?

Never seen anything like this before, but then, I don't think I've seen anyone try to write a stateful backend.

tplass:
> Without digging into it closely, I could guess that it's the modification of company-prefix
> (which, like I said, is a bad idea), bites you in the ass here.

That's what I'm trying to prove to me.  But why is it so hard to find the
spot where things diverge?

> In any case, when point is right after head (without space), having the prefix be an empty string
> doesn't make sense to me.

Prefix must be "" then for otherwise `head` get's replaced by the attribute
completion.

That's where I pull the rug from under company.

> or not. See e.g. company--continue, and in particular this comparison:
>```elisp 
>                        (= (- (point) (length new-prefix))
>                           (- company-point (length company-prefix)))
>```

`company--continue` is never reached, in any of the scenarios.

I'm currently tasked with writing a piece of documentation and would
dearly like to get this fixed.  So, I'll keep looking into it for a
while.  Stay tuned.

tplass:
Cracked it.  Reason unknown as was the underlying problem.

I've found a, well, workaround for the creation of the overlay menu for multiple attribute candidates.  I wouldn't consider this a fix and appreciate any hints on why this works as well as suggestions on how to improve on this.

So, on tag name completion, a 'quit signal appears to clear whatever condition prevents the overlay (situation 1 above):

```elisp
 (post-completion (cond
                   ((eq company-nxml-last-command 'company-nxml-tag)
                    (if (eq (aref arg 0) ?/)
                        (insert ">")
                      (or (company-nxml-add-required-attributes arg (point))
                          (signal 'quit nil)))) ; this makes multiple attribute candidates work
                          ^^^^^^^^^^^^^^^^^^
```

Don't ask me why, I'm just poking at it.

I'll put this through further tests but it appears to work in all situations so far.

It even handles this ambiguous situation gracefully:

User wants a `blockquote` tag, inserts `<b`, 

```xml
 <b
   ^ 1. completion request shows attribute candidates as element "b" exists (is not a prefix)
     2. user Ctrl-g's out of it
     3. user requests completion again: now menu shows tag name completions for prefix "b"
```

tplass:
> Maybe it makes company-post-command call company-abort, 

`company-abort` doesn't get called, at least edebug never stops there.  Also, `company-cancel`'s argument is in all scenarios a string, never a symbol.

I have no idea what `signal` really does in Emacs' guts and what it means if a `'quit` makes Lisp code behave in a predictable fashion.

tplass:
I'd like to improve the handling of attribute completion further by adding a
user option to always display the tooltip even when there's only one
candidate (ie. `(not (cdr company-candidates))`)

Is there a company variable/hook/mechanism visible to a backend I could 
use for this?

dgutov:
"always display the tooltip even when there's only one candidate" is one of the available frontends, but that's the user's choice. Not of any backend.

tplass:
> "always display the tooltip even when there's only one candidate" is one of the available
> frontends

Ah ok, even better.

Let me ask as a user: which one is it?  I tried all those available
through `customize` but I get differing behaviour only for elements
with multiple attributes.

Elements with just a single declared attribute get their attribute
auto-inserted, no matter which frontend is selected.

If you'd like to try, save this as doc.rnc:
```rnc
doc = element doc { head, body }
attlist.head = attribute version { text }?, attribute id { text }?
attlist.body = attribute id { text }?
head = element head { attlist.head, text }
body = element body { attlist.body, text }
start = doc
```
find-file doc.xml in the same dir, compare attribute completion for
`head` and `body`.

dgutov:
> Let me ask as a user: which one is it?

`company-pseudo-tooltip-frontend`. As opposed to `company-pseudo-tooltip-unless-just-one-frontend` which is used by default.

> Elements with just a single declared attribute get their attribute auto-inserted, no matter which frontend is selected.

Frontends are there just for visualization. They don't effect the behavior.

tplass:
> company-pseudo-tooltip-frontend.

Tried that, same difference.

> Frontends are there just for visualization. They don't effect the behavior.

Sorry, by "behavior" I meant "show the tooltip" no matter how many
elements the list of candidates has.  My problem is that I can't get
the visualization to appear for single-element `company-candidates`.

dgutov:
> Tried that, same difference.

But you see the inline preview frontend?

tplass:
> But you see the inline preview frontend?

You mean `company-preview-frontend`?  That behaves like this:

* single-element candidate list: same as all the others, no visualization, goes straight to `post-completion`

* multi-element candidate list: first completion request inserts the naked attribute (apparently `(car company-candidates)`), second completion requests changes the naked `attribute` to `attribute=""` (as per `post-completion`)

dgutov:
> single-element candidate list: same as all the others, no visualization, goes straight to post-completion

Then it's not either of the frontends' fault. When there is no visualization, frontends are simply not asked to do anything, do they?

Anyway, if there is just one completion, why wouldn't you want to go straight to post-completion? It's not like there's anything to choose.

tplass:
> Then it's not either of the frontends' fault. 

I'm not saying it's the frontend's fault.  Yes, they simply get called.

> Anyway, if there is just one completion, why wouldn't you want to go straight to post-completion?
> It's not like there's anything to choose.

No and yes.

No: if you ask for completion and there's only a single choice, then insert it already.

Yes: even if there's only a single candidate, I'd like to *choose* whether to have it inserted right away or have it previewed in a tooltip and possibly decide not to have it inserted after all.  In that scenario I just want to see my options.

Look, I'm not saying company has a policy problem.  But I'd like to be able to force a visualization.  Is that possible from within a backend?  If not, take this as a feature request, please.

dgutov:
> I'm not saying it's the frontend's fault. Yes, they simply get called.

They probably aren't called at all, in this case.

> I'd like to *choose* whether to have it inserted right away or have it previewed in a tooltip and possibly decide not to have it inserted after all

Have you tried `M-x company-manual-begin`?

tplass:
> Have you tried M-x company-manual-begin?

Nice!  That's what I had in mind - almost.

What's required is a way to choose between `company-complete` and `company-manual-begin` depending on context while using one keybinding for both.

The scenarios would be this: for tag names (`<head`) completion is requested with `company-complete` while for attributes (`<head dir=""`) `company-manual-begin` is used.  Both share the same keybinding.  Would it be a good idea or even possible to fiddle the keymap  dynamically?

dgutov:
> The scenarios would be this: for tag names (<head) completion is requested with company-complete while for attributes (<head dir="") company-manual-begin is used.

Where would point be situated in the second example?

> Would it be a good idea or even possible to fiddle the keymap dynamically?

From the backend? Not a good idea. And anyway, you can do anything only after the command is dispatched. Changing the keymap would be too late then.

This request is reminiscent of the discussion in #150. But in that case, we expect the user to be able to choose whether to run post-completion or not, and the question is how to allow them to make that choice the most naturally. And whether many users will want the ability to make it, of course.

tplass:
After quite some testing and learning to appreciate what `(backtrace)`
does to debugging, I've been able to fix and extend the functionality
of `company-nxml`.

Before I commit any changes, I'd like to get some feedback on how to
proceed in the light of my findings and changes to the current
`nxml-mode`.

First, I finally tracked down the cause of the strange behaviour when
despite `company-candidates` being a list, no tooltip was shown but
instead the first candidate auto-inserted.


```elisp
(defun company-complete ()
  "Insert the common part of all candidates or the current selection.
The first time this is called, the common part is inserted, the second
time, or when the selection has been changed, the selected candidate is
inserted."
  (interactive)
  (when (company-manual-begin)
    (if (or company-selection-changed
            (eq last-command 'company-complete-common))
        (call-interactively 'company-complete-selection)
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
      (call-interactively 'company-complete-common)
      (setq this-command 'company-complete-common))))
```

So, two completion requests immediately following each other would be
redirected into different directions.  And along the way
`company-complete-selection` would do a `(setq company-selection 0)`
which selects the first candidate.

Now, switching between `company-complete-common` and
`company-complete-selection` is probably a design choice of `company`.
But for my backend, this is a showstopper as the user cannot use
`company-complete` on his keybinding but is forced to use
`company-manual-begin` or directly `company-complete-common`.

**First question**: Is there any way (except maybe some screwy
`defadvice`ing around `company-complete`) to enable a backend to set
`this-command` to whatever it sees fit?  If not, there's no point for
me to continue with this project.  Or could you add a variable that
would stop `company-complete` from switching between states?

The second big improvement for `company-nxml` is the support of XML
namespaces which is utterly missing from the production. This means
that currently only non-namespace XML schemas are supported.  E.g., it
is simply impossible to use `company` completion for editing XSLT
stylesheets as candidates are missing the "xsl:" namespace prefix.
And there's more that fails due to this shortcoming.

I therefore completely revamped the main XML functions by essentially
calling `nxml` functions and let *it* compute the candidates as well as
steal^H^H^H^H^H repurposing some of the code that does support work.

The `nxml` code for this has changed incompatibly from my Emacs
(25.0.50.1) to the version formerly known as 26 (which is now called
25.2?!).  However, only a few trivial changes were necessary to port
my code to the current `nxml` code base.

**So, my next question is**: does it make any sense at all to ship an
updated `company-nxml` in the light of the different incompatible
versions of `nxml`?

Oh btw, the completion implemented for the current master `nxml` is
just as basic and IMHO "broken" as it ever was.  That's why at least
*I* will continue to enjoy `company`.

dgutov:
Looking at the description of "situation 1", I don't see you mentioning that you called `company-complete` twice. Did you?

If so, yes, that's what this command is intended to do. With the assumption that the backend won't show you any new completions after you've just inserted the common part, just the ones that is showed you before. And if you're calling this command a second time, you mean to choose a particular completion.

Why is it a problem for you backend?

> enable a backend to set this-command to whatever it sees fit?

What semantic meaning would that have? Otherwise, it would be just another hack in a pile of hacks.

> Or could you add a variable that would stop company-complete from switching between states?

You could simply ask the users to stop using `company-complete`. Maybe provide an alternative command. This could go okay with you releasing this backend on your own: you pick a new name for it and publish on MELPA separately.

> does it make any sense at all to ship an updated company-nxml in the light of the different incompatible versions of nxml?

Does the current version of `company-nxml` become broken as well?

tplass:
> Looking at the description of "situation 1", I don't see you mentioning that you called
> company-complete twice. Did you?

My backend never calls `company-complete`.  In fact, it only makes
calls to `company-begin-backend` and `company-grab`.  But my problems
could be resolved if you'd be willing to open up the one-way channel
between engine and backend.

> And if you're calling this command a second time, you mean to choose a
> particular completion.
> 
> Why is it a problem for you backend?

It's the user who calls `company-complete` as may times as they
choose.  And it's the engine which fails to recognize that a full
circle of request/candidate-selection/insertion/post-completion has come
to an end.

I'm not saying that's a bug but I'd like my backend to be able to tell
the engine: "with the next `company-complete`, let's start afresh,
don't bother checking `last-command`".

Semantically, what this piece of advice does:

```elisp
(defadvice company-complete (after company-nxml-set-this-command activate)
  (setq this-command (if SOME_CONDITION
                         'company-complete-selection
                       'company-complete-common)))
```

where SOME_CONDITION is some sort of offical `company` API.  Why would
that be a hack?

> You could simply ask the users to stop using company-complete. Maybe provide an alternative
> command. This could go okay with you releasing this backend on your own: you pick a new name for
> it and publish on MELPA separately.

Should that be taken as a "go away"?

Look, I find myself between a rock and a hard place.  As much as I
like to use company and nxml, both are very rigid and have their own
ideas and assumptions about what is right for the user (never mind the
programmer).

The fact that there is no outcry about the missing namespace support
makes me wonder whether a company-nxml backend is actually
relevant.  I guess, all nxml users are happy with what nxml provides
in terms of completion.

So if there's seemingly no demand, then there's no need for me to
publish a company backend or stand-alone module.

> Does the current version of company-nxml become broken as well?

From looking at the code, I think it will continue to do what it does.
But the numerous bugs in the regexps as well as in computing the list
of candidates remain.  It's quality certainly does not compare to the
other backends.

dgutov:
> My backend never calls company-complete.

So the user does. But I don't see anything in the description about them doing it twice. If you did, maybe I could have saved you the investigation. Anyway...

> But my problems could be resolved if you'd be willing to open up the one-way channel between engine and backend.

There are lots of channels between the backends and the different parts of the engine. All of them are documented, so that any backend author knows when to use them.

> And it's the engine which fails to recognize that a full circle of request/candidate-selection/insertion/post-completion has come to an end.

This vaguely reminds me of the logic inside minibuffer.el that chooses whether to call `:exit-function` with `sole`, `finished` or `exact`. And that's predicated on the concept of "fields", which Company doesn't have, so far.

> where SOME_CONDITION is some sort of offical company API

What kind? How would it be called? Can you show a first cut of its description inside of `company-backends` docstring?

> Should that be taken as a "go away"?

It should be taken as stated. Since it looks like your backend is a special beast, I'm unlikely to accept it as-is, and publishing it separately can be a good choice.

> Look, I find myself between a rock and a hard place. As much as I like to use company and nxml, both are very rigid and have their own ideas and assumptions about what is right for the user (never mind the programmer).

They're as flexible as the people working on them managed to make them. Making Company more flexible requires more work on it, including maybe reimagining some concepts it's concerned with. Simply stating "this round peg doesn't fit in your square hole, why don't you file the sides a little" is not what I'm looking for.

> The fact that there is no outcry about the missing namespace support
makes me wonder whether a company-nxml backend is actually
relevant. I guess, all nxml users are happy with what nxml provides
in terms of completion.

Maybe so. Or maybe most of them just work on (X)HTML.

> So if there's seemingly no demand, then there's no need for me to publish a company backend or stand-alone module.

You won't really know until you publish anyway.

> From looking at the code, I think it will continue to do what it does. But the numerous bugs in the regexps as well as in computing the list of candidates remain. It's quality certainly does not compare to the other backends.

Thanks for checking. We'll probably switch to using the built-in nXML completion starting with Emacs 26. Does it have problems with namespaces, too?

dgutov:
`company-nxml` is now deprecated.

PR #550, opened 2016-06-29
Fixed company-semantic-annotation, so that it ..

differentiablef:
.. can appropriately deal with function pointers in C.

As an example of the original problem, consider the struct

`struct blah { int (*hates)(uint8_t, uint32_t);  void *blank; }`

   
the annotation returned by company-semantic-annotation for 'hates' would be
         `(*hates)(uint8_t, uint32_t)`
which when coupled with the default value of 'company-semantic-insert-arguments'
yields decidedly annoying results. 

This patch fixes the problem. Additionally. I've attached a sequence of screenshots
illustrating the problem.

prepatch
prepatch2
prepatch3
postpatch
postpatch2
postpatch3

The issue with the original code seems to be the implicit assumption that the 
type of blah::hates is not seen as a function by semantic. This is incorrect, as 
semantic properly designates it a function-pointer which being callable, makes it 
a function.

### DIFF
--- a/company-semantic.el
+++ b/company-semantic.el
@@ -40,6 +40,7 @@
 (declare-function semantic-tag-buffer "semantic/tag")
 (declare-function semantic-active-p "semantic")
 (declare-function semantic-format-tag-prototype "semantic/format")
+(declare-function semantic-format-tag-canonical-name "semantic/format")
 
 (defgroup company-semantic nil
   "Completion backend using Semantic."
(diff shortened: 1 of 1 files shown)


dgutov:
Thanks! This looks good, but a bit nontrivial. Could you add a couple of tests as well?

differentiablef:
Will do

dgutov:
@differentiablef Ping?

PR #560, opened 2016-07-20
Don't convert user-error signals to error

Wilfred:
If a backend calls (user-error ...), we want to preserve Emacs' behavior
of not opening a debugger, even when debug-on-error is enabled.

### DIFF
--- a/company.el
+++ b/company.el
@@ -898,6 +898,9 @@ matches IDLE-BEGIN-AFTER-RE, return it wrapped in a cons."
       (if (functionp company-backend)
           (apply company-backend args)
         (apply #'company--multi-backend-adapter company-backend args))
+    (user-error (user-error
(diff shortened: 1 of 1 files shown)


dgutov:
Why not? I wouldn't expect the backend to call `user-error`, and if some function in `completion-at-point-functions` does, the backtrace would be handy to find out which one.

Wilfred:
racer.el calls `user-error` when a user has misconfigured an important path. This isn't a bug in racer.el, so opening the debugger doesn't seem appropriate.

The other disadvantage is the message when debug-on-error is disabled:

> Company: backend company-capf error "/wrong-path is not a directory" with args (candidates collec)

This seems rather noisy: A user doesn't need to know about args.

dgutov:
Ok, let's try this and see if it mishandles any important errors.

Thanks.

PR #573, opened 2016-08-28
Add company-flow

aaronjensen:
This is my first attempt at creating a backend so I'm sure it could be better. It seems to work, but I'm open to any feedback.

Some of the process sending code was borrowed from flycheck.

Also, if this belongs in its own package that's fine to, just let me know. Thanks!

### DIFF
--- a/company-flow.el
+++ b/company-flow.el
@@ -1,37 +1,86 @@
 ;;; company-flow.el --- Flow backend for company-mode  -*- lexical-binding: t -*-
-(require 'cl-lib)
 
-(defun company-flow-handle-signal (process _event)
+;; Copyright (C) 2016 by Aaron Jensen
+
+;; Author: Aaron Jensen <aaronjensen@gmail.com>
+;; URL: https://github.com/aaronjensen/company-flow
+;; Version: 0.1.0
+;; Package-Requires: ((company "0.8.0") (dash "2.13.0"))
+
+;;; Commentary:
+
+;; This package adds support for flow to company. It requires
+;; flow to be in your path.
+
+;; To use it, add to your company-backends for your preferred javascript modes,
+;; for example:
+
+;; (setq company-backends-js2-mode '((company-flow :with company-dabbrev)
+;;                                     company-files
+;;                                     company-dabbrev))
+
+;;; License:
+
+;; This file is not part of GNU Emacs.
+;; However, it is distributed under the same license.
+
+;; GNU Emacs is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation; either version 3, or (at your option)
+;; any later version.
+
+;; GNU Emacs is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with GNU Emacs; see the file COPYING.  If not, write to the
+;; Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+;; Boston, MA 02110-1301, USA.
+
+;;; Code:
+(require 'company)
+(require 'dash)
+
+(defun company-flow--handle-signal (process _event)
   (when (memq (process-status process) '(signal exit))
     (let ((callback (process-get process 'company-flow-callback))
           (prefix (process-get process 'company-flow-prefix)))
-      ;; (message "%S" (company-flow-parse-output (company-flow-get-output process)))
+      ;; (message "%S" (company-flow--parse-output (company-flow--get-output process)))
       (funcall callback (->> process
-                             company-flow-get-output
-                             company-flow-parse-output
+                             company-flow--get-output
+                             company-flow--parse-output
+                             ;; Remove nils
+                             (--filter it)
                              (all-completions prefix))))))
 
-(defun company-flow-make-candidate (line)
-  (let* ((words (split-string line " "))
-         (text (car words))
-         (meta (mapconcat 'identity (cdr words) " ")))
-    (propertize text 'meta meta)))
+(defun company-flow--make-candidate (line)
+  "Creates a candidate with a meta property from LINE.
 
-(defun company-flow-parse-output (output)
-  (mapcar 'company-flow-make-candidate
+LINE is expected to look like:
+registrationSuccess () => {type: 'REGISTRATION_SUCCESS'}"
+  (let ((first-space (string-match " " line)))
+    (when first-space
+      (let ((text (substring line 0 first-space))
+            (meta (substring line (+ 1 first-space))))
+        (propertize text 'meta meta)))))
+
+(defun company-flow--parse-output (output)
+  (mapcar 'company-flow--make-candidate
           (split-string output "\n")))
 
-(defun company-flow-get-output (process)
+(defun company-flow--get-output (process)
   "Get the complete output of PROCESS."
   (with-demoted-errors "Error while retrieving process output: %S"
     (let ((pending-output (process-get process 'company-flow-pending-output)))
       (apply #'concat (nreverse pending-output)))))
 
-(defun company-flow-receive-checker-output (process output)
+(defun company-flow--receive-checker-output (process output)
   "Receive a syntax checking PROCESS OUTPUT."
   (push output (process-get process 'company-flow-pending-output)))
 
-(defun company-flow-process-send-buffer (process)
+(defun company-flow--process-send-buffer (process)
   "Send all contents of current buffer to PROCESS.
 
 Sends all contents of the current buffer to the standard input of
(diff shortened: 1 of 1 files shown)


dgutov:
Yes, I'd prefer to keep this a separate package. Unless you're planning on becoming a regular contibutor to the rest of Company as well. But here are some comments.

company-flow.el
```diff
@@ -0,0 +1,91 @@
+;;; company-flow.el --- Flow backend for company-mode  -*- lexical-binding: t -*-
+(require 'cl-lib)
+
+(defun company-flow-handle-signal (process _event)
+  (when (memq (process-status process) '(signal exit))
+    (let ((callback (process-get process 'company-flow-callback))
+          (prefix (process-get process 'company-flow-prefix)))
+      ;; (message "%S" (company-flow-parse-output (company-flow-get-output process)))
+      (funcall callback (->> process
```
dgutov: `->>` looks like it's coming from `dash`. Needs a `require`.

company-flow.el
```diff
@@ -0,0 +1,91 @@
...
+                             company-flow-get-output
+                             company-flow-parse-output
+                             (all-completions prefix))))))
+
+(defun company-flow-make-candidate (line)
+  (let* ((words (split-string line " "))
```
dgutov: Using `match-string`, `length` and `substring` should make this function more performant.

Of course, I'm not sure if you ever get enough elements here to worry about memory pressure.

company-flow.el
```diff
@@ -0,0 +1,91 @@
...
+         (text (car words))
+         (meta (mapconcat 'identity (cdr words) " ")))
+    (propertize text 'meta meta)))
+
+(defun company-flow-parse-output (output)
+  (mapcar 'company-flow-make-candidate
```
dgutov: Similarly, this one could be replaced with `re-search-forward` in the output buffer. `company-clang--parse-output` could be used as an example.

company-flow.el
```diff
@@ -0,0 +1,91 @@
...
+
+(defun company-flow-get-output (process)
+  "Get the complete output of PROCESS."
+  (with-demoted-errors "Error while retrieving process output: %S"
+    (let ((pending-output (process-get process 'company-flow-pending-output)))
+      (apply #'concat (nreverse pending-output)))))
+
+(defun company-flow-receive-checker-output (process output)
+  "Receive a syntax checking PROCESS OUTPUT."
+  (push output (process-get process 'company-flow-pending-output)))
```
dgutov: Is there a reason to use process properties for these values? Buffer-local (or even global) variables should work just as well, and look more straightforward.

company-flow.el
```diff
@@ -0,0 +1,91 @@
...
+
+Sends all contents of the current buffer to the standard input of
+PROCESS, and terminates standard input with EOF."
+  (save-restriction
+    (widen)
+    (process-send-region process (point-min) (point-max)))
+  (process-send-eof process))
+
+(defun company-flow-candidates-query (prefix callback)
+  (let* ((default-directory "~/Source/the-link")
```
dgutov: Leftover from testing?

company-flow.el
```diff
@@ -0,0 +1,91 @@
...
+  "Grab prefix for flow."
+  (and (not (company-in-string-or-comment))
+       (or (company-grab-symbol-cons "\\." 1)
+           'stop)))
+
+(defun company-flow-annotation (candidate)
+  (format " %s" (get-text-property 0 'meta candidate)))
+
+(defun company-flow-meta (candidate)
+  nil)
```
dgutov: Are these functions going to get non-trivial definitions in the future?

company-flow.el
```diff
@@ -0,0 +1,91 @@
...
+
+(defun company-flow-doc (candidate)
+  nil)
+
+;;;###autoload
+(defun company-flow (command &optional arg &rest _args)
+  (interactive (list 'interactive))
+  (cl-case command
```
dgutov: Using `pcase` is a bit more preferable than `cl-case`, these days.

company-flow.el
```diff
@@ -0,0 +1,91 @@
...
+    (interactive (company-begin-backend 'company-flow))
+    (prefix (company-flow-prefix))
```
dgutov: And we also prefer to "privatize" utility functions (include `--` in the name), to emphasize that nothing outside of the current package is supposed to use them.

company-flow.el
aaronjensen: I had a hard time understanding what meta was for. It seems to return things similar to annotation in other backends.

dgutov: `meta` is normally longer. Annotation is displayed in the completions list, whereas meta is shown in the echo area, for the currently selected completion.

company-flow.el
dgutov: Oh, I see: you're using one process call per completion. In that case, that makes sense.

An alternative approach would be to make the sentinel an anonymous closure that captures the values you want it to use down the line. Like `company-clang--start-process` does.

aaronjensen: I borrowed this from flycheck. I don't know what guarantees company has in terms of how many in flight async requests to a given backend it could have at once. If it's only one then yeah, buffer local or global variables would probably be fine. 

I also like that it keeps the state only in the place that needs it, but I assume that I just need to get more used to buffer local/global variables in emacs

company-flow.el
aaronjensen: There isn't an output buffer at all in this, just a string that gets passed around. The completion options are likely to be few enough that memory probably won't be a problem. Are you thinking that putting it into a buffer then `re-search-forward`ing through it would be faster or more efficient? I can see it being more memory efficient possibly because the split list doesn't exist, but I don't understand enough about emacs perf or memory to know if it'd be all that much better.

I thought about using `-unfold` and `string-match`to build the list of candidates directly, but I'd only want to do that if it was measurably better.

aaronjensen:
@dgutov thank you very much for the feedback. I think I've hit most of it.

I submitted to melpa, but I had a question about usage instructions.

This is what I have now: https://github.com/aaronjensen/company-flow/

My question is, is that appropriate instructions? I see most backends say to just add it to `company-backends` but because there is no `flow-mode` and there are multiple javascript modes (they're not even all `js-mode` based, like `web-mode`) I wouldn't feel comfortable putting a mode predicate in `prefix`. Is there a better way?

Thanks again for the feedback and company-mode, it's great!

company-flow.el
dgutov: > There isn't an output buffer at all in this, just a string that gets passed around.

There is. The process is associated with a buffer. You're just using `process-filter` (which gets passed the output string(s)), so you don't see that.

The alternative is parsing the output buffer contents from the sentinel.

> Are you thinking that putting it into a buffer then re-search-forwarding through it would be faster or more efficient?

Same thing?

> I can see it being more memory efficient possibly because the split list doesn't exist, but I don't understand enough about emacs perf or memory to know if it'd be all that much better.

Yes, naturally you'll have to measure first, before tweaking. Evaluating `(benchmark 1 '(let ((company-backend 'company-flow)) (company-call-backend 'candidates "")))` should do that, but you might need to find a big enough project, so that the time is noticeable (or not).

dgutov:
I prefer the instructions in https://github.com/Sarcasm/company-irony (with `eval-after-load`)

`company-backends-js2-mode` doesn't mean anything to a regular Emacs user.

> I wouldn't feel comfortable putting a mode predicate in prefix. Is there a better way?

Like many existing backends do, you can create a user option that holds a list of major modes where this backend will be applicable. Not sure how well that will work with `web-mode` anyway.

Maybe `prefix` should also perform some Flow-specific check, to make sure the current project is not plain JavaScript. Look for a certain file in a parent directory, for example.

aaronjensen:
> Maybe prefix should also perform some Flow-specific check, to make sure the current project is not plain JavaScript. Look for a certain file in a parent directory, for example.

Something like 

```elisp
       (locate-dominating-file buffer-file-name ".flowconfig")
```

would work, but that has a seemingly significant performance hit associated with it. Is there a better way to check it? Or is memoization a good option?

> Like many existing backends do, you can create a user option that holds a list of major modes where this backend will be applicable. Not sure how well that will work with web-mode anyway.

Ah, good idea, I'll try that.

dgutov:
> that has a seemingly significant performance hit associated with it

Not here. But I have an SSD, and it could a problem using Tramp anyway.

> Or is memoization a good option?

I believe so. `company-eclim` does that, for example.

company-flow.el
```diff
@@ -43,17 +41,34 @@
 (require 'company)
 (require 'dash)
 
+(defgroup company-flow ()
+  "Flow company backend."
+  :group 'editing
```
dgutov: The group should be `company`, I think.

aaronjensen:
@dgutov it seems that the `process-send-region` is super slow, it takes about 340ms to do once on my MBP w/ SSD and i7. It seems specific to flow, however as if I change the process command to `cat -` it is instant again.

Is emacs waiting for something when it sends a region?

aaronjensen:
Ah! `(setq process-connection-type nil)`

aaronjensen:
I'm going to go ahead and close this out. I really appreciate the feedback @dgutov. The package is up on melpa: https://melpa.org/#/company-flow

dgutov:
> Is emacs waiting for something when it sends a region?

No idea, sounds like an OS-dependent performance regression.

> I really appreciate the feedback @dgutov

Thanks for acting on it. :)

Issue #510, opened 2016-05-09
Popups flickers with asynchronous backend and caching disabled


Yevgnen:
Hi,

Thanks for the awesome package !

I find that what when grouping a asynchronous backend with other backends, the company candidate menu flashes quite often each time I type a single char. This behavior is not occurred with manual completion.Details can be found in [https://github.com/syohex/emacs-company-jedi/issues/16](https://github.com/syohex/emacs-company-jedi/issues/16)

In my opinion, grouping backends is still quite useful even the 'main backend' works fine in most situations. So, if I want to use auto completion and grouping backends, is it possible or will it be difficult to fix this ?

Any help will be appreciated :)

dgutov:
Thanks for the report, it's a somewhat known problem, see also #337 and the linked discussion.

When there is a grouped backend in a group, the whole group works asynchronously. If at least one of the backends in it disables caching, the caching is disabled, again, for the whole group.

The async+no-cache combination currently leads to flickering.

dgutov:
Should work better now. Also see #654 for some related comments (and the end).

Issue #337, opened 2015-03-30
company-require-match t forces synchronous mode of operation


dgutov:
https://github.com/abingham/emacs-ycmd/issues/148#issuecomment-87501458

dgutov:
Shouldn't be a problem now with https://github.com/company-mode/company-mode/pull/1471.

PR #581, opened 2016-09-07
Allow async backends to cache future results

### LINKED ISSUE #580: Allow async backends to use candidate results for previous, but prefix matching callbacks calls
Let's say you set `company-idle-delay` very low, like `0` and use an `async` backend. You then type a few characters like `ite`. The `it` may set off an async candidate call and then the `ite` will set off an async candidate call. Likely the `it` call will return first, but since both the `tick` and `pos` have been modified, `company-idle-begin` will discard this result and wait for the callback of the last prefix.

This means that for a cacheable backend, the "cacheability" is of no benefit if one types faster than the backend can return. Idealy, the result for `it` would come back, and even if I had made it all the way to `ite` or `item` it would display the completion results, filtered down with `all-completions`, of course.

I'm not sure of the best place to do this or if there are any dragons so I wanted to get your input on it before submitting anything.

Thanks!

**edit** A natural extension of this is that we do not even ask the backend for `ite` if we know the backend is cacheable and `it`'s result is still pending. This would prevent quite a bit of work I would guess.

aaronjensen:
If an async backend has not yet called its callback and the user
continues to type along the symbol that was being completed and the
backend supports caching, then instead of firing the backend again we
wait for the result of the first and use it.

Fixes #580 

Here is one possible way to do this. What do you think?

### DIFF
--- a/company.el
+++ b/company.el
@@ -1187,14 +1187,19 @@ can retrieve meta-data for them."
             (cl-dotimes (i (1+ len))
               (when (setq prev (cdr (assoc (substring prefix 0 (- len i))
                                            company-candidates-cache)))
-                (setq candidates (all-completions prefix prev))
-                (cl-return t)))))
+                (cl-return t)))
+            (cond ((eq prev 'async)
+                   ;; Pretend we found results so we don't trigger another search.
+                   t)
+                  (prev
+                   (setq candidates (all-completions prefix prev))))))
         (progn
           ;; No cache match, call the backend.
           (setq candidates (company--preprocess-candidates
                             (company--fetch-candidates prefix)))
           ;; Save in cache.
-          (push (cons prefix candidates) company-candidates-cache)))
+          (if candidates
+              (push (cons prefix candidates) company-candidates-cache))))
     ;; Only now apply the predicate and transformers.
     (setq candidates (company--postprocess-candidates candidates))
     (when candidates
@@ -1218,6 +1223,9 @@ can retrieve meta-data for them."
             (tick (buffer-chars-modified-tick))
             (pt (point))
             (backend company-backend))
+        ;; Indicate that we are currently fetching this result so we do not make
+        ;; another async call again unless we need to.
+        (setq company-candidates-cache (list (cons prefix 'async)))
         (funcall
          (cdr c)
          (lambda (candidates)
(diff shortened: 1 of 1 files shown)


company.el
```diff
@@ -1218,6 +1223,9 @@ can retrieve meta-data for them."
             (tick (buffer-chars-modified-tick))
             (pt (point))
             (backend company-backend))
+        ;; Indicate that we are currently fetching this result so we do not make
+        ;; another async call again unless we need to.
+        (setq company-candidates-cache (list (cons prefix 'async)))
```
aaronjensen: Does this need to be guarded with a `no-cache` check?

company.el
```diff
@@ -1187,14 +1187,19 @@ can retrieve meta-data for them."
...
+                  (prev
+                   (setq candidates (all-completions prefix prev))))))
         (progn
           ;; No cache match, call the backend.
           (setq candidates (company--preprocess-candidates
                             (company--fetch-candidates prefix)))
           ;; Save in cache.
-          (push (cons prefix candidates) company-candidates-cache)))
+          (if candidates
+              (push (cons prefix candidates) company-candidates-cache))))
```
aaronjensen: Was there any reason to cache this before? It conflicted w/ the async entry

company.el
```diff
@@ -1230,7 +1238,12 @@ can retrieve meta-data for them."
                    (list (cons prefix
                                (company--preprocess-candidates candidates))))
              (unwind-protect
-                 (company-idle-begin buf win tick pt)
+                 (if (company-call-backend 'no-cache prefix)
+                     (company-idle-begin buf win tick pt)
+                   (let ((current-prefix (company-call-backend 'prefix))
+                         (ignore-case (company-call-backend 'ignore-case)))
+                     (when (string-prefix-p prefix current-prefix ignore-case)
+                       (company-idle-begin buf win (buffer-chars-modified-tick) (point)))))
```
aaronjensen: There's a danger here in some cases where you end up on another matching prefix on a different line that is in a different context and wouldn't otherwise match. That could be mitigated by either remembering our line or if `prev` is `async` above, resetting `tick` and `point` in some way that this closure would have access to it. It could also be fine because maybe caching would work that way any way.

aaronjensen:
There are failing tests here that I haven't looked into yet, I wanted to get your thoughts on this first before digging in more. If you have any pointers on the tests it'd be appreciated as well. Thanks!

company.el
dgutov: No particular reason. It's just simpler without the condition, and the async handler overwrites the cache later anyway.

company.el
dgutov: I don't think remembering the line number is, strictly speaking, enough. If the user just pasted code code (without newlines), this may result in some faulty completions. The odds are low, of course, but we err on the side of correctness, so far.

We could, however, introduce a serial "current completion" number, and see if it hasn't changed. Together with the `(string-prefix-p prefix current-prefix ignore-case)` check, it might be enough.

company.el
dgutov: No reason to keep the previous entries: if any of them were usable, the async call wouldn't have happened. And filling in cache for prefixes longer than the current is fast enough.

dgutov:
@aaronjensen I've answered the questions, but TBH I'd much prefer help on #337 and #510.

dgutov:
Since the recent push, at least, this became un-mergeable. There's nothing stopping each individual backend to contain logic such as this, though.

aaronjensen:
👍 new changes look good, it feels almost unsettling for the tooltip (or in my case, child frame) to not flicker... I got so used to it 😄

PR #609, opened 2016-11-12
tests: use /bin instead of /lost+found

dunn:
`/lost+found` isn't guaranteed to exist on a filesystem, so this
switches to something that is.


dgutov:
Thanks!

PR #618, opened 2016-12-07
Preview the common part of completion candidates

kkatsuyuki:
The new frontends to preview the common part of candidates was added.
 `company-preview-common-frontend`

I changed a little the function `company-preview-show-at-point` to be used 
in this frontend. The default frontend `company-preview-if-just-one-frontend` 
originally uses the function, so both frontends reference it.

Configuration of customized variable `company-frontends` and
the fuction `company-frontends-set` to set the frontends are also
changed since the new frontend is added.

In the default configuration the new frontend is not activated.

### DIFF
--- a/company.el
+++ b/company.el
@@ -2916,14 +2916,13 @@ Delay is determined by `company-tooltip-idle-delay'."
 
 (defvar-local company-preview-overlay nil)
 
-(defun company-preview-show-at-point (pos &optional company-common)
+(defun company-preview-show-at-point (pos completion)
   (company-preview-hide)
 
-  (let ((completion  (or company-common (nth company-selection company-candidates))))
-    (setq completion (copy-sequence (company--pre-render completion)))
-    (font-lock-append-text-property 0 (length completion)
-                                    'face 'company-preview
-                                    completion)
+  (setq completion (copy-sequence (company--pre-render completion)))
+  (font-lock-append-text-property 0 (length completion)
+                                  'face 'company-preview
+                                  completion)
     (font-lock-prepend-text-property 0 (length company-common)
                                      'face 'company-preview-common
                                      completion)
@@ -2960,18 +2959,19 @@ Delay is determined by `company-tooltip-idle-delay'."
       (let ((ov company-preview-overlay))
         (overlay-put ov (if ptf-workaround 'display 'after-string)
                      completion)
-        (overlay-put ov 'window (selected-window))))))
+        (overlay-put ov 'window (selected-window)))))
 
 (defun company-preview-hide ()
   (when company-preview-overlay
     (delete-overlay company-preview-overlay)
     (setq company-preview-overlay nil)))
 
-(defun company-preview-frontend (command &optional company-common)
+(defun company-preview-frontend (command)
   "`company-mode' frontend showing the selection as if it had been inserted."
   (pcase command
     (`pre-command (company-preview-hide))
-    (`post-command (company-preview-show-at-point (point) company-common))
+    (`post-command (company-preview-show-at-point (point)
+                                                  (nth company-selection company-candidates)))
     (`hide (company-preview-hide))))
 
 (defun company-preview-if-just-one-frontend (command)
@@ -2992,15 +2992,19 @@ Delay is determined by `company-tooltip-idle-delay'."
     (not (overlay-get company-pseudo-tooltip-overlay 'invisible))))
 
 (defun company-preview-common--show-p ()
+  "Returns whether the preview of common can be showed or not"
   (and company-common
        (or (eq (company-call-backend 'ignore-case) 'keep-prefix)
            (string-prefix-p company-prefix company-common))))
 
 (defun company-preview-common-frontend (command)
-  "`company-preview-frontend', but only shown for single candidates."
+  "`company-mode' frontend preview the common part of candidates."
   (when (or (not (eq command 'post-command))
             (company-preview-common--show-p))
-    (company-preview-frontend command company-common)))
+    (pcase command
+      (`pre-command (company-preview-hide))
+      (`post-command (company-preview-show-at-point (point) company-common))
+      (`hide (company-preview-hide)))))
 
 ;;; echo ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
--- a/company.el
+++ b/company.el
@@ -242,7 +242,7 @@ The visualized data is stored in `company-prefix', `company-candidates',
                          (const :tag "preview" company-preview-frontend)
                          (const :tag "preview, unique only"
                                 company-preview-if-just-one-frontend)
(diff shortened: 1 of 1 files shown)


company.el
```diff
@@ -237,6 +242,8 @@ The visualized data is stored in `company-prefix', `company-candidates',
                          (const :tag "preview" company-preview-frontend)
                          (const :tag "preview, unique only"
                                 company-preview-if-just-one-frontend)
+                         (const :tag "preview, unique and common part"
```
dgutov: When unique, the common part is the same, right?

company.el
```diff
@@ -2912,7 +2919,7 @@ Delay is determined by `company-tooltip-idle-delay'."
 (defun company-preview-show-at-point (pos)
   (company-preview-hide)
 
-  (let ((completion (nth company-selection company-candidates)))
+  (let ((completion (if (cdr company-candidates) company-common (nth company-selection company-candidates))))
```
dgutov: Any particular reason for this change?

company.el
dgutov: Or: doesn't this affect the other `-preview-` frontends as well?

kkatsuyuki:
To see the efficiency of this frontend I make the animated gif.
company_common

In the video case I will input `add-log-current-defun-function`.
When I input 'add-lo', all the candidates had 'add-log-' and the common
substring 'g-' was previewed. I continued to type 'add-log-cu', then 
'rrent-defun' is previewed. After the completion I selected 
the corresponding candidate `add-log-current-defun-function` and completed.

Now to use this frontend need to reset the variable `company-frontends`, but 
I think this frontend should be used in the default settings.

company.el
dgutov: You can just say "common", I think. When there's just one, it'll be the whole candidate.

kkatsuyuki:
To see the efficiency of this frontend I made the animated gif.
company_common


In the video case I will input `add-log-current-defun-function`.
When I input 'add-lo', all the candidates had 'add-log-' and the common
substring 'g-' was previewed. I continued to type 'add-log-cu', then 
'rrent-defun' is previewed. After the completion I selected 
the corresponding candidate `add-log-current-defun-function` and completed.

company.el
dgutov: As an alternative, it can grow a new argument.

kkatsuyuki: Thank you for the good advice. I will change the function to take a new argument.

kkatsuyuki:
I modified the preview function `company-preview-show-at-point` to 
take the new argument of 'common' flag.
By this change, the following frontends work well.
 - `company-preview-frontend`
 - `company-preview-if-just-one-frontend` (default)
 - `company-preview-common-frontend` (newly added)

kkatsuyuki:
I modified the preview function `company-preview-show-at-point` to 
take the new argument of 'common' flag.
By this change, the following frontends work well.
 - `company-preview-frontend`
 - `company-preview-if-just-one-frontend` (default)
 - `company-preview-common-frontend` (newly added)

company.el
```diff
@@ -2909,10 +2916,10 @@ Delay is determined by `company-tooltip-idle-delay'."
 
 (defvar-local company-preview-overlay nil)
 
-(defun company-preview-show-at-point (pos)
+(defun company-preview-show-at-point (pos &optional common)
```
dgutov: Why not just pass in the exact candidate this function should use?

company.el
```diff
@@ -2984,6 +2991,19 @@ Delay is determined by `company-tooltip-idle-delay'."
...
+  (and company-common
+       (or (eq (company-call-backend 'ignore-case) 'keep-prefix)
+           (string-prefix-p company-prefix company-common))))
+
+(defun company-preview-common-frontend (command)
+  "`company-preview-frontend', but only shown for single candidates."
+  (when (or (not (eq command 'post-command))
+            (company--show-common-inline-p))
+    (company-preview-frontend command t)))
+
```
dgutov: Extra line here.

company.el
```diff
@@ -2984,6 +2991,19 @@ Delay is determined by `company-tooltip-idle-delay'."
   (when (overlayp company-pseudo-tooltip-overlay)
     (not (overlay-get company-pseudo-tooltip-overlay 'invisible))))
 
+
```
dgutov: And here.

company.el
```diff
@@ -2984,6 +2991,19 @@ Delay is determined by `company-tooltip-idle-delay'."
...
+(defun company--show-common-inline-p ()
```
dgutov: I think this should be called `company-preview-common--show-p`.

company.el
kkatsuyuki: I see..., it seems that passing the exact common part can make the code simplified. OK, I will remodify.

kkatsuyuki: I guessed wrong. The code seems not to be simplified. I will let the function to take the common and the completed strings will be selected by the number of argumets.

dgutov: Number of arguments? I'm not sure what you mean.

kkatsuyuki:
`company-preview-show-at-point` is modified to take the exact
common part of candidates.

Additionaly the function name `company--show-common-inline-p`
is modified to `company-preview-common--show-p`.

Exact line is deleted.

company.el
```diff
@@ -2909,10 +2916,10 @@ Delay is determined by `company-tooltip-idle-delay'."
 
 (defvar-local company-preview-overlay nil)
 
-(defun company-preview-show-at-point (pos)
+(defun company-preview-show-at-point (pos &optional company-common)
   (company-preview-hide)
 
-  (let ((completion (nth company-selection company-candidates)))
+  (let ((completion  (or company-common (nth company-selection company-candidates))))
```
dgutov: Sorry, I don't like this. This is pretty much the same definition as you showed initially. This function should take a new, non-optional argument.

And you can't thread it through `company-preview-frontend`: what if we want to add a new argument to all frontends later, that would do something else?

Instead, `company-preview-common-frontend` should not delegate to `company-preview-frontend` when COMMAND is `post-command`. It should call `company-preview-show-at-point` directly there.

kkatsuyuki:
By this change `company-preview-show-at-point`
takes the exact completion candidate
as an nonoptional arg. So the candidate was added as
the arg where this function is called.

At the previous change `company-preview-frontend` was modified to
take the new argument, but it was returned at this time.
Made `company-preview-common-frontend` not to call
`company-preview-frontend`, and instead it called
`company-preview-show-at-point` directly.

dgutov:
Looking good, thanks!

Please note that this pretty much exhausts the allowed limit of contributions to `company-mode` (or other packages in GNU ELPA, or Emacs itself) without copyright assignment.

kkatsuyuki:
Thank you for your good warning. 
Would you tell me what I should do  for much contribution like this?

dgutov:
Yes, please follow the instructions here: http://orgmode.org/cgit.cgi/org-mode.git/plain/request-assign-future.txt.

kkatsuyuki:
Thanks!

PR #621, opened 2016-12-09
Use pipes instead of PTYs to receive output from Clang

### LINKED ISSUE #620: company-clang always No completion found
hi, i am new to company.
when i setup the company, and want to use it.
there are always:
```
No completion found
user-error: Cannot complete at point
```

here is my init setup:

```
(require 'cc-mode)
(require 'company)
(add-hook 'after-init-hook 'global-company-mode)
(setq company-backends (delete 'company-semantic company-backends))
(define-key c-mode-map [(tab)] 'company-complete)
(define-key c++-mode-map [(tab)] 'company-complete)
(setq company-clang-executable "/usr/bin/clang-3.8")
```

and i also found, when i want to exit the emacs, i found compay-clang still running(have a active buffer, shown in process list):
`company-clang   run     *clang-output*  /dev/pty1    /usr/bin/clang-3.8 -fsyntax-only -Xclang -code-completion-macros -x c++ -I/cygdrive/e/lua/include -I/usr/include -Xclang -code-completion-at=-:28:5 -`

btw, i use this setup(almost the same, such as include path) for auto-complete-clang, and it works fine.

juergenhoetzel:
We don't need any job control features. Also Pipes are more efficient and more
robust than PTYs, which Emacs uses by default.

Fixes #620.

### DIFF
--- a/company-clang.el
+++ b/company-clang.el
@@ -208,8 +208,9 @@ or automatically through a custom `company-clang-prefix-guesser'."
       (with-current-buffer buf
         (erase-buffer)
         (setq buffer-undo-list t))
-      (let ((process (apply #'start-process "company-clang" buf
-                            company-clang-executable args)))
+      (let* ((process-connection-type nil)
+             (process (apply #'start-process "company-clang" buf
+                             company-clang-executable args)))
         (set-process-sentinel
          process
          (lambda (proc status)


dgutov:
Thanks!

Issue #462, opened 2016-01-29
Support for remote files via TRAMP?


root42:
I am editing quite big C++ projects via ssh / TRAMP using Emacs. I have set up company and clang in principle. When editing a project locally, everything works fine, e.g. #includes are found.

However when editing my remote project, clang failes, because it cannot find the includes. In the clang-error buffer I can see that my local clang is called, instead of the remote one, which probably already would alleviate the problem:

`clang failed with error 1:
/usr/bin/clang -fsyntax-only -Xclang -code-completion-macros -x c++ -Xclang -code-completion-at=-:229:7 -

<stdin>:17:10: fatal error: 'foo/bar.h' file not found
# include <foo/bar.h>

```
     ^`
```

Is there a way to run the clang command remotely? When using (shell-command ...) in a remote buffer, this will usually launch a remote shell, which I use e.g. to find tags with GNU global. So I think this behavior should be possible.

root42:
It seems to me that something could be achieved if the usage of start-process is substitued with start-file-process (in [line 221 of company-clang.el](https://github.com/company-mode/company-mode/blob/master/company-clang.el#L211])).

According to the [Emacs Lisp documentation](https://www.gnu.org/software/tramp/#Remote-processes) this function will be treated by TRAMP if the default directory is remote, which should be the case when editing a remote buffer.

dgutov:
> the usage of start-process is substitued with start-file-process

Sounds about right. Could you try it out?

root42:
> > the usage of start-process is substitued with start-file-process

> Sounds about right. Could you try it out?

I have tried it, but so far with no success. Company gives me this:

```
Company: An error occurred in auto-begin
Writing to process: bad file descriptor, company-clang
company-begin-backend: Cannot complete at point
```

And clang outputs this:

```
clang failed with error 1:
clang -fsyntax-only -Xclang -code-completion-macros -x c++ - -Xclang -code-completion-at=-:2288:13 -

clang-3.8: error: no input files
```

It seems that the stdin gets lost on the way. I am a bit at a loss here. Perhaps this is easy to fix, but I would need some help here -- maybe there are some people with experience in elisp and tramp?

What definitely works seems to be the default-directory. It is correctly set depending on the buffer. Local buffers look like `/Users/$USER/path/to/my/code` and remote buffers will show soemthing like `/scp:myremotehost:/path/to/my/code`. I checked this by debugging the `company-clang--start-process` function.

[...]

PR #627, opened 2017-01-04
Disable idle completion on remote buffers

juergenhoetzel:
This disables idle-completion in TRAMP buffers (due to ```(eq non-essential t)```. and fixes [Support for remote files via TRAMP](https://github.com/company-mode/company-mode/issues/462)

Also see related: [Bind non-essential to prevent tramp error ](https://github.com/company-mode/company-mode/pull/487)


juergenhoetzel:
Upstream Bug in `start-file-process`

PR #656, opened 2017-03-28
Bind message-truncate-lines around call to message in company-echo-show

### LINKED ISSUE #652: Truncating 'meta strings
Hi,

Emacs has a variable `message-truncate-lines` that prevents calls to `message` from growing the minibuffer.  Should there be a way to bind it around company's call to `message` when displaying `meta` annotations?

The rationale is that it isn't always easy to truncate a meta string to make it fit exactly: for example, if it contains symbols like `∀` or `⟶`, truncating by length isn't enough.  Setting message-truncate-lines is much simpler.

There are a few ways to do this:

* Add a backend command
* Let meta return a cons instead of a single string
* Add a customizable variable:

```elisp
(defcustom company-meta-truncate nil
  "When non-nil, ensure that 'meta annotations never grow the minibuffer."
  :type 'boolean
  :package-version '(company . "0.9.3"))

(defun company-echo-show (&optional getter)
  (when getter
    (setq company-echo-last-msg (funcall getter)))
  (let ((message-log-max nil))
    (if company-echo-last-msg
        (let ((message-truncate-lines company-meta-truncate))
          (message "%s" company-echo-last-msg))
      (message ""))))
```

I have a small preference for the second solution.  What do you think?

cpitclaudel:
Closes #652.

### DIFF
--- a/company.el
+++ b/company.el
@@ -3015,10 +3015,16 @@ Delay is determined by `company-tooltip-idle-delay'."
 
 (defvar company-echo-delay .01)
 
+(defcustom company-echo-truncate-lines t
+  "Whether frontend messages written to the echo area should be truncated."
+  :type 'boolean
+  :package-version '(company . "0.9.3"))
+
 (defun company-echo-show (&optional getter)
   (when getter
     (setq company-echo-last-msg (funcall getter)))
-  (let ((message-log-max nil))
+  (let ((message-log-max nil)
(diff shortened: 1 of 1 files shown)


dgutov:
I'm slightly worried that affecting `company-echo-frontend` and `company-echo-strip-common-frontend` is not TRT, but let's see if that actually bothers anyone.

Thanks!

Issue #201, opened 2014-10-05
Support C#/Java generics in templating


nosami:
I'm trying to get support for C# generics working with templating, but company-mode seems to support only parentheses to mark parameters.

I'd like type parameters to be templated too such that

``` cs
new List<T>(T item)
```

could offer the first `T` inside the angle brackets as a parameter, then `T item` would be the 2nd parameter.

I tried to do this myself, but alas, my elisp skills are still too weak.

I would be eternally grateful if you could make this work :) /cc @bbbscarter

PR #660, opened 2017-04-10
Add function and binding to delete template fields and move to next

### LINKED ISSUE #657: Quickly delete unwanted arguments
As mentioned here abingham/emacs-ycmd#408 , how can I quickly delete unwanted arguments that auto completed by some backends.
 A completion of python's print function looks like this,
```
    print(value, ..., sep=' ', end='\n', file=sys.stdout, flush=False)
```
afaik, yasnippet supports `ctrl-d` to delete them one by one.

ptrv:
Add binding to `company-template-nav-map` to delete template and move to next.

Resolves #657

### DIFF
--- a/company-template.el
+++ b/company-template.el
@@ -65,24 +65,30 @@
 (defun company-template-clear-field ()
   "Clear the field at point."
   (interactive)
-  (company-template-remove-field (company-template-field-at (point)) t))
+  (let ((ovl (company-template-field-at (point))))
+    (when ovl
+      (company-template-remove-field ovl t)
+      (let ((after-clear-fn (overlay-get ovl 'after-clear-function)))
+        (when (functionp after-clear-fn)
+          (funcall after-clear-fn))))))
 
-(defun company-template--clear-c-like-field (start end)
-  "Clear the field at point for c like templates.
-If there is a next field, delete everything up to it.  If the field at
-point is the last field, remove preceding comma if present.  START and END
-are the bounds of the current field."
-  (let* ((next-field-start (company-template-find-next-field))
+(defun company-template--after-clear-c-like-field ()
+  "Function that can be called after deleting a field of a c-like template.
+For c-like templates it is set as `after-post-fn' property on fields in
+`company-template-add-field'.  If there is a next field, delete everything
+from point to it.  If there is no field after point, remove preceding comma
+if present."
+  (let* ((pos (point))
+         (next-field-start (company-template-find-next-field))
          (last-field-p (not (company-template-field-at next-field-start))))
     (cond ((and (not last-field-p)
-                (< end next-field-start)
+                (< pos next-field-start)
                 (string-match "^[ ]*,+[ ]*$" (buffer-substring-no-properties
-                                              end next-field-start)))
-           (delete-region start next-field-start))
+                                              pos next-field-start)))
+           (delete-region pos next-field-start))
           ((and last-field-p
                 (looking-back ",+[ ]*" (line-beginning-position)))
-           (delete-region (match-beginning 0) end))
-          (t (delete-region start end)))))
+           (delete-region (match-beginning 0) pos)))))
 
 (defun company-template-find-next-field ()
   (let* ((start (point))
@@ -120,13 +126,12 @@ are the bounds of the current field."
         (delq templ company-template--buffer-templates))
   (delete-overlay templ))
 
-(defun company-template-add-field (templ beg end &optional display clear-function)
+(defun company-template-add-field (templ beg end &optional display after-clear-fn)
   "Add new field to template TEMPL spanning from BEG to END.
 When DISPLAY is non-nil, set the respective property on the overlay.
 Leave point at the end of the field.
-CLEAR-FUNCTION is a function that can be used to apply custom behavior when
-deleting a field in `company-template-remove-field'.  If nil,
-`delete-region' is used."
+AFTER-CLEAR-FN is a function that can be used to apply custom behavior
+after deleting a field in `company-template-remove-field'."
   (cl-assert templ)
   (when (> end (overlay-end templ))
     (move-overlay templ (overlay-start templ) end))
@@ -139,8 +144,8 @@ deleting a field in `company-template-remove-field'.  If nil,
       (overlay-put ov 'display display))
     (overlay-put ov 'company-template-parent templ)
     (overlay-put ov 'insert-in-front-hooks '(company-template-insert-hook))
-    (when clear-function
-      (overlay-put ov 'clear-function clear-function))
+    (when after-clear-fn
+      (overlay-put ov 'after-clear-function after-clear-fn))
     (push ov siblings)
     (overlay-put templ 'company-template-fields siblings)))
 
@@ -148,13 +153,7 @@ deleting a field in `company-template-remove-field'.  If nil,
   (when (overlayp ovl)
     (when (overlay-buffer ovl)
       (when clear
-        (let ((clear-function (overlay-get ovl 'clear-function))
-              (start (overlay-start ovl))
-              (end (overlay-end ovl)))
-          (if (and (not (eq clear 'clear-bounds-only))
-                   (functionp clear-function))
-              (funcall clear-function start end)
-            (delete-region start end))))
+        (delete-region (overlay-start ovl) (overlay-end ovl)))
       (delete-overlay ovl))
     (let* ((templ (overlay-get ovl 'company-template-parent))
            (siblings (overlay-get templ 'company-template-fields)))
@@ -173,7 +172,7 @@ deleting a field in `company-template-remove-field'.  If nil,
 (defun company-template-insert-hook (ovl after-p &rest _ignore)
   "Called when a snippet input prompt is modified."
   (unless after-p
-    (company-template-remove-field ovl 'clear-bounds-only)))
+    (company-template-remove-field ovl t)))
 
 (defun company-template-post-command ()
   (company-template-clean-up)
@@ -218,7 +217,7 @@ deleting a field in `company-template-remove-field'.  If nil,
     (while (re-search-forward "\\([^,]+\\),?" end 'move)
       (when (zerop (car (parse-partial-sexp last-pos (point))))
         (company-template-add-field templ last-pos (match-end 1) nil
-                                    #'company-template--clear-c-like-field)
+                                    #'company-template--after-clear-c-like-field)
         (skip-chars-forward " ")
         (setq last-pos (point))))))
 
--- a/company-template.el
+++ b/company-template.el
@@ -72,7 +72,8 @@
   (let ((ovl (company-template-field-at (point))))
     (when ovl
       (company-template-remove-field ovl t)
-      (let ((after-clear-fn (overlay-get ovl 'after-clear-function)))
+      (let ((after-clear-fn
+             (overlay-get ovl 'company-template-after-clear)))
         (when (functionp after-clear-fn)
           (funcall after-clear-fn))))))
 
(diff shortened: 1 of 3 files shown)


company-template.el
```diff
@@ -65,15 +66,27 @@
            (> pos (point))
            (< pos minimum)
            (setq minimum pos)))
-    (push-mark)
-    (goto-char minimum)
-    (company-template-remove-field (company-template-field-at start))))
+    (let ((start-ovl (company-template-field-at start)))
+      (if delete-template
```
dgutov: Instead of having two different behaviors inside a function, I think I'd rather the first part of this function was extracted to a utility one, and both `company-template-forward-field` and `company-template-forward-field-and-delete` called it.

dgutov:
Thanks. Any ideas about the comma cleaup?

I suppose it can be a function that is assigned to the `company-template-after-delete` property of the overlay. `company-template-c-like-templatify` would add it via `company-template-add-field`, and `company-template-remove-field` would use it.

dgutov:
BTW, please remember that this project follows the same copyright assignment guidelines as Emacs itself.

ptrv:
> Any ideas about the comma cleaup?

This way the comma is also removed, because we delete the region from the template begin to begin of next template.

company-template.el
ptrv: Sure, this was just a quick way to have something to discuss based on the feature request in #657 .

ptrv: > Instead of having two different behaviors inside a function, I think I'd rather the first part of this function was extracted to a utility one, and both company-template-forward-field and company-template-forward-field-and-delete called it

Sounds good. I'll update the patch with your suggestions

dgutov:
> we delete the region from the template begin to begin of next template

Good point. I've been worried about the cases with generics like https://github.com/company-mode/company-mode/issues/201#issuecomment-57953969 where the second and third fields are separated by something other than comma (and that text isn't supposed to be deleted).

But maybe we can just decide that the user won't press `C-d` there (or if they do that by accident, they'll use undo). That's certainly simpler.

ptrv:
> I've been worried about the cases with generics like #201 (comment) where the second and third fields are separated by something other than comma (and that text isn't supposed to be deleted).

I see. Haven't thought about that. Maybe we could check if there is a comma between the templates and only delete from beginning of template to next template when there is a comma. In other cases just delete the template?

dgutov:
Yeah, but `company-template-forward-field-and-delete` isn't supposed to be syntax-specific.

Anyway, I've yet to hear of optional generic parameters.

dgutov:
By the way, if the parameter we're deleting is the last one, shouldn't we delete the command-space that went before it?

ptrv:
> By the way, if the parameter we're deleting is the last one, shouldn't we delete the command-space that went before it?

Ah yes, I guess the comma should be removed too.

Ok, covering all cases makes this really intense. I am wondering whether adding so much logic is even worth it. `yasnippet` only deletes the placeholder and if the placeholder contains the comma then it gets deleted and if the comma is outside the placeholder it also stays there.

ptrv:
I think it would already be an improvement if `C-d` deletes "just" the template. Then one can continue using `C-d` to delete the remaining characters (comma).

What do you think?

dgutov:
I'm fine with this as a preliminary solution. I can merge this PR, as long as you make the requested change, *and* satisfy the copyright assignment requirements.

Have you contributed to Emacs and/or GNU ELPA projects before? If the 15 lines limit is noticeably exceeded, we'll need you to sign the argeement. Good news is, German citizens can do it pretty quickly online.

ptrv:
Ok, I'll add the requested changes.

I haven't signed the copyright assignment yet for Emacs/Elpa. Can you point me to url where I can do it online? Thx

dgutov:
> Can you point me to the url where I can do it online?

I don't know how exactly it happens. Maybe not online, but electronically. The first step is the same for everybody, though. Please fill out and send this form by email: http://orgmode.org/cgit.cgi/org-mode.git/plain/request-assign-future.txt

"the name of the program" should remain unchanged (Emacs).

ptrv:
Ok, I've sent the email with the form. Let's see how long it will take.

I've also update the PR and separated the function so that we don't have two functionalities within one function.

I added also checks to the remove function, so that only commas are deleted and not `>(` as with template arguments when deleting a field. Also when removing the last field I think it makes a lot sense to remove the preceeding comma.

I will also add tests, but pushed for now for discussion.

company-template.el
```diff
@@ -55,19 +56,44 @@
...
 (defun company-template-forward-field ()
   (interactive)
+  (let ((start (point))
+        (next-field-start (company-template-get-next-field-start)))
+    (push-mark)
+    (goto-char next-field-start)
+    (company-template-remove-field (company-template-field-at start))))
+
+(defun company-template-clear-field ()
+  "Clear field at point.
```
dgutov: Some copyedits:

_the_ field at point

company-template.el
```diff
@@ -55,19 +56,44 @@
...
+If there is a next field, delete everything up to next field.  If field at
```
dgutov: ...up to *it*.  If *the* field at

company-template.el
```diff
@@ -55,19 +56,44 @@
...
+point is last field, remove preceeding comma if present."
```
dgutov: ...is *the* last field, remove the *preceding* comma...

company-template.el
```diff
@@ -55,19 +56,44 @@
...
+  (interactive)
   (let* ((start (point))
-         (templates (company-template-templates-at (point)))
+         (current-ovl (company-template-field-at start)))
+    (when current-ovl
+      (let ((this-field-end (overlay-end current-ovl))
+            (next-field-start (company-template-get-next-field-start)))
+        (cond ((and (company-template-field-at next-field-start)
+                    (< this-field-end next-field-start)
+                    (string-match "^\\s-*,+\\s-*$" (buffer-substring-no-properties
```
dgutov: This should work fine, but I wonder if the regexps are more complex than the have to. Like, the template probably won't ever contain tabs, and there probably won't be any spaces before the comma.

Feel free to disregard this comment.

company-template.el
```diff
@@ -55,19 +56,44 @@
...
+                                                    this-field-end next-field-start)))
+               (delete-region start next-field-start))
+              ((and (not (company-template-templates-at
+                          (company-template-get-next-field-start)))
+                    (looking-back ",+\\s-*" nil t))
```
dgutov: The second argument shouldn't be nil, let's use `(line-beginning-position)` instead.

company-template.el
```diff
@@ -55,19 +56,44 @@
...
+               (delete-region (match-beginning 0) this-field-end))
+              (t (delete-region start this-field-end)))))))
+
+(defun company-template-get-next-field-start ()
```
dgutov: Let's call it `company-template-find-next-field`.

dgutov:
`company-template.el` needs an `M-x copyright-update`.

dgutov:
The new logic in `company-teplate-clear-field` looks reasonable, @amosbird please try it out.

Ideally, though, it should be decoupled from the function, because the templates are not necessarily tied to code or C-like languages.

ptrv:
@dgutov I've applied all your suggestions. The custom clearing is now done with a function in `company-template-remove-field` which is added in `company-template-add-field` as you suggested in your first comment, in order to decouple the custom logic from `company-template-clear-field`. Added also a couple of tests.

company-template.el
```diff
@@ -33,6 +33,7 @@
   (let ((keymap (make-sparse-keymap)))
     (define-key keymap [tab] 'company-template-forward-field)
     (define-key keymap (kbd "TAB") 'company-template-forward-field)
+    (define-key keymap (kbd "C-d") 'company-template-clear-field)
```
ptrv: I think we should fall back to delete-char when there is no active overlay. It should be possible to use the default shortcut for deleting a character when editing the variable name. Would be also the same as yasnippets behavior.

company-template.el
```diff
@@ -179,7 +217,8 @@ Leave point at the end of the field."
   (let ((last-pos (point)))
     (while (re-search-forward "\\([^,]+\\),?" end 'move)
       (when (zerop (car (parse-partial-sexp last-pos (point))))
-        (company-template-add-field templ last-pos (match-end 1))
+        (company-template-add-field templ last-pos (match-end 1) nil
+                                    #'company-template--clear-c-like-field)
```
ptrv: Should we add the clear function also to the objc templatifier?

dgutov: Probably. I'm not sure if there are additional cases to be handled here. Maybe not.

company-template.el
dgutov: When we're editing a variable name, the template field is gone, right?

A template field is an overlay. We should probably create a separate keymap for them, with just this one binding.

company-template.el
```diff
@@ -109,14 +139,22 @@ Leave point at the end of the field."
...
 
 (defun company-template-remove-field (ovl &optional clear)
   (when (overlayp ovl)
     (when (overlay-buffer ovl)
       (when clear
-        (delete-region (overlay-start ovl) (overlay-end ovl)))
+        (let ((clear-function (overlay-get ovl 'clear-function))
+              (start (overlay-start ovl))
+              (end (overlay-end ovl)))
+          (if (and (not (eq clear 'clear-bounds-only))
```
dgutov: This is a minor complaint (don't bother if you're tired of refining this), but maybe the clear-function should just receive the beginning of the next field? Then it could be called simply at the end of `company-template-clear-field`.

Even the beginning of the next field is not necessary for our current purpose, since we can be sure the following comma and spaces are outside of it.

company-template.el
ptrv: Ok, let's keep it as it is

company-template.el
ptrv: Nice! I like it. Done

company-template.el
ptrv: Nice! Done!

ptrv:
@dgutov Thanks for your suggestions! I applied all of them. I have now several commits, but will squash as soon as the PR is ready to be merged.

company-template.el
```diff
@@ -55,19 +60,51 @@
...
+    (goto-char next-field-start)
+    (company-template-remove-field (company-template-field-at start))))
+
+(defun company-template-clear-field ()
+  "Clear the field at point."
+  (interactive)
+  (let ((ovl (company-template-field-at (point))))
+    (when ovl
+      (company-template-remove-field ovl t)
+      (let ((after-clear-fn (overlay-get ovl 'after-clear-function)))
```
ptrv: So this works because the overlay still exists even after delete-overlay has been called. But when does it get destroyed? When the parent overlay is deleted, because it has the evaporate property? Or is this some kind of leak?

dgutov: `delete-overlay` doesn't "kill" the object, it just makes it point to no buffer, see the doc of `overlay-buffer` for confirmation.

The overlay will be GCd eventually, so it  should be fine.

dgutov:
@ptrv Looking great, thanks! No need to squash, the GitHub UI gives that option when merging anyway (you'll remain the author of the commit).

Please let me know when the assignment goes through.

company-template.el
ptrv: I've read the docs, but was wondering when the object will be deleted.
Ok, then it should be fine to get the property after calling `delete-overlay`

ptrv:
> No need to squash, the GitHub UI gives that option when merging anyway

Fine for me.

> Please let me know when the assignment goes through.

I've received the form from the FSF, need to sign it and send it back and wait for the FSF to send it back with their sign.
I'll ping you here as soon as I get the signed form. 
Yes,

ptrv:
@dgutov assignment process is complete

dgutov:
Great!

Please also add a NEWS entry, under `##Next`.

ptrv:
> Please also add a NEWS entry, under ##Next.

Done

dgutov:
Thanks a lot!

PR #674, opened 2017-06-13
Set company-template-nav-map as parent of company-template-field-map

ptrv:
In order to be able to forward a field while company-template-field-map is
active, we need to set company-template-nav-map as parent.

This is a regression introduced in #660 
I thought I tested this, but it slipped through.

### DIFF
--- a/company-template.el
+++ b/company-template.el
@@ -37,6 +37,7 @@
 
 (defvar company-template-field-map
   (let ((keymap (make-sparse-keymap)))
+    (set-keymap-parent keymap company-template-nav-map)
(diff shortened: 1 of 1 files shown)


dgutov:
Huh, indeed. Thanks!

Maybe this bug was not so noticeable because we rarely leave the default value in place.

ptrv:
> Maybe this bug was not so noticeable because we rarely leave the default value in place.
 
Yes, exactly. I ran only into this when I started to write a function definition in a cpp file.

Thanks for the quick merge.

Issue #526, opened 2016-05-29
YouCompleteMe-style tab completion?


Valloric:
[Full disclosure: I'm the author of [YCM](https://github.com/Valloric/YouCompleteMe) so I'm biased. :)]

I've been trying out emacs lately to see can I replicate my uber-customized Vim setup on the dark side :). I'm using company-mode with emacs-ycmd and it's great! Thanks for making company-mode awesome and for continuing to maintain it; OSS dev work is often thankless.

Other than one missing feature, company-mode is doing everything I want it to do.

The best way to describe the feature is with the following gif. (Wait until the part where `LongestCommonSubsequenceLength` is being completed.)

![gif](https://camo.githubusercontent.com/1f3f922431d5363224b20e99467ff28b04e810e2/687474703a2f2f692e696d6775722e636f6d2f304f50346f6f642e676966)

In the gif, only tab is being used, nothing else (no enter).

In YCM, tab is used to _both_ select a completion candidate from the list _and_ to insert it in the buffer. So the user types and gets completions. Pressing tab selects the first item in the completion menu and inserts it in the buffer. Pressing tab again selects the second item and replaces the inserted item with the second one. This can continue as long as the user wishes to cycle through the menu.

The benefits are the following:
1. Only one shortcut key to interact with instead of two.
2. One fewer button press (no need to confirm an entry).

Any chance company-mode can get an option of some sort to behave like this? Something like `complete-and-cycle` for `auto-completion-tab-key-behavior`. It would be greatly appreciated, if for any reason then to lessen the cognitive load on those Vim users using Emacs.

dgutov:
Hi Val,

I'm glad you're happy with company-mode overall.

Thanks for the great explanation; we've discussed this before in the issue #343, and there's a gist linked there that implements this, albeit in a not very clean fashion.

aaronjensen:
Here's the gist for reference: https://gist.github.com/aaronjensen/a46f88dbd1ab9bb3aa22

I haven't had time to pull request it properly. It has been working well for me, however.

dgutov:
There's still work to do on the design here. In particular:

* Advising `company-set-selection` is not great. Advising the frontends would be better. Having two new frontends would be best; they can wrap the two simpler ones, and do things depending on whether the user has already typed tab. Things like showing the completion inline only after tab, and binding `company-selection` to `-1` locally around `company-pseudo-tooltip-frontend` when the user hasn't typed tab yet.
[...]

Issue #517, opened 2016-05-22
Expand completions as you cycle through them?


ZzAntares:
Hello, new company user here, so far the experience has been great, however I was wondering if is there a way to make company completion expand the completion candidate as I cycle through the completion candidates, if I keep pressing `TAB` it will replace current word with the next completion candidate, I never have to choose a completion using `RET` since the completion has been expanded earlier I just keep typing.

A gif will explain it better:
![autocompletion](https://cloud.githubusercontent.com/assets/3653680/15452231/facfc30e-1fac-11e6-961e-9c99ec4d7497.gif)

I am so used to this that selecting the completion with `RET` feels like a waste of time :grin:

Is there any setting I could add to make it behave like this? I did try `company-complete-common-or-cycle` but only completes to the common part (as the name implies duh).

Thanks for the help.

dgutov:
Thanks for the nice gif. We've discussed this before, see #343. It also contains a gist implementing this, or something close.

ZzAntares:
Thanks I'll stay tuned to when this will be doable thing.

Issue #343, opened 2015-04-28
Invoke with tab


ephexeve:
I am using Company mode with Emacs-ycmd, instead of enter, is there anyway I can complete it with tab? So it completes the suggestion?
Currently I have my tab binded to:
```
; bind company-select-next to tab
(eval-after-load 'company
  '(progn
     (define-key company-active-map (kbd "TAB") 'company-select-next)
     (define-key company-active-map [tab] 'company-select-next)))
;; (setq company-selection t) <-- this doesn't make sense -- Dmitry
(setq company-idle-delay 0)
```
This works for selecting the next completion, I would like to bind it so it also selects the completion suggestion.

For example:
```
int some_random_var = 0
some_<tab> = some_random_var
```

dgutov:
Try binding `company-complete-common-or-cycle` instead.

ephexeve:
@dgutov it worked like expected, thanks. 
Is there anyway to achieve the same action as this one? https://asciinema.org/a/19319 (vim using YouCompleteMe)
Using tab to navigate the completion and enter to select it or just type another character and it will keep the last selection?

dgutov:
> type another character and it will keep the last selection

dgutov:
> type another character and it will keep the last selection

Any character at all? No, we don't have this option. Maybe we should, but it looks unusual.

unknownzerx:
@dgutov 
+1 for the action which @ephexeve has metioned. I feel comfortable with that.

It is the behavior same to auto-complete.el that

- press tab first to complete common
- then press tab again to complete all (select the first candidate)
  - but keep the candidates popup open
  - and press more tab to select other candidate

thus all I need to do is press tab several times and continue to edit. No more "RET" needed.

aaronjensen:
I came here to file this issue as well. This is a very handy method of auto-completion in my experience. One you get used to it, having to press enter feels quite unnatural. I wouldn't agree w/ @unknownzerx's first bullet, however, so maybe it makes sense to have separate commands that people can bind on their own. 

My preference:

* as you type, menu shows with nothing selected
* tab selects and completes first but leaves menu open. 
* Repeated tab selects & completes next, shift tab selects and completes previous
* Any other key moves on and closes the window.

![acp](https://cloud.githubusercontent.com/assets/8588/12077700/359a9a96-b1a7-11e5-97ac-ff7375291e3a.gif)

dgutov:
I don't understand this idea: "Any other key moves on and closes the window".

[...]

PR #706, opened 2017-09-02
Company Tab and Go

nikital:
This pull request implements the feature discussed extensively in #526, #517, #343 and #216.

It's implemented as a separate package with minimal configuration required. To enable it just add `company-tng-frontend` to the frontend list. The code is modified from my Gist in #526 in order to enable minimum configuration for the user.

Huge thanks to @dgutov and @aaronjensen for helping me iterate on this feature 😃 

I sent a mail to assign@gnu.org, I'll let you know when the assignment bureaucracy is done.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,11 @@
 ## Next
 
 * New user option `company-tooltip-maximum-width`.
+* New feature `company-tng'. When added as a frontend, it allows you to insert
+  the candidates as soon they are selected, eliminating the need to press RET to
+  confirm your selection. See `company-tng.el` and use
+  `(company-tng-configure-default)` to give it a try.
+  ([#526](https://github.com/company-mode/company-mode/issues/526))
 
 ## 2017-07-15 (0.9.4)
 
--- a/company-tng.el
+++ b/company-tng.el
@@ -1,0 +1,73 @@
+;;; company-tng.el --- company-mode frontend that inserts a candidate
+;;; into the buffer as soon as it's selected, Vim style
+
+;; Copyright (C) 2017  Free Software Foundation, Inc.
+
+;; Author: Nikita Leshenko
+
+;; This file is part of GNU Emacs.
+
+;; GNU Emacs is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; GNU Emacs is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with GNU Emacs.  If not, see <http://www.gnu.org/licenses/>.
+
+
+;;; Commentary:
+;;
+;; company-tng (tab and go) allows you to use TAB to both select a
+;; completion candidate from the list and to insert it into the
+;; buffer.
+;;
+;; It cycles the candidates like `yank-pop' or `dabbrev-expand' or
+;; Vim: Pressing TAB selects the first item in the completion menu and
+;; inserts it in the buffer. Pressing TAB again selects the second
+;; item and replaces the inserted item with the second one. This can
+;; continue as long as the user wishes to cycle through the menu.
+;;
+;; The benefits are that there is only one shortcut key to interact
+;; with and there is no need to confirm an entry.
+
+;;; Code:
+
+(require 'company)
+(require 'cl-lib)
+
+(defvar-local company-tng--overlay nil)
+
+;;;###autoload
+(defun company-tng-frontend (command)
+  "When the user changes the selection at least once, this
+frontend will display the candidate in the buffer as if it's
+already there and any key outside of `company-active-map' will
+confirm the selection and finish the completion."
+  (cl-case command
+    (show
+     (let ((ov (make-overlay (point) (point))))
+       (setq company-tng--overlay ov)
+       (overlay-put ov 'priority 2)))
+    (update
+     (let ((ov company-tng--overlay)
+           (selected (nth company-selection company-candidates))
+           (prefix (length company-prefix)))
+       (move-overlay ov (- (point) prefix) (point))
+       (overlay-put ov 'display (and company-selection-changed selected))))
+    (hide
+     (when company-tng--overlay
+       (delete-overlay company-tng--overlay)))
+    (pre-command
+     (when (and company-selection-changed
+                (not (company--company-command-p (this-command-keys))))
+       (company--unread-this-command-keys)
+       (setq this-command 'company-complete-selection)))))
+
+(provide 'company-tng)
+;;; company-tng.el ends here
--- a/company-tng.el
+++ b/company-tng.el
@@ -83,6 +83,9 @@ confirm the selection and finish the completion."
            (selected (nth company-selection company-candidates))
            (prefix (length company-prefix)))
        (move-overlay ov (- (point) prefix) (point))
+       (overlay-put ov
+                    (if (= prefix 0) 'after-string 'display)
+                    (and company-selection-changed selected))
        (overlay-put ov 'display (and company-selection-changed selected))))
     (hide
      (when company-tng--overlay
--- a/company-tng.el
+++ b/company-tng.el
@@ -86,7 +86,8 @@ confirm the selection and finish the completion."
        (overlay-put ov 'display (and company-selection-changed selected))))
     (hide
      (when company-tng--overlay
-       (delete-overlay company-tng--overlay))
+       (delete-overlay company-tng--overlay)
+       (kill-local-variable 'company-tng--overlay))
      (advice-remove 'company-select-next 'company-tng--allow-unselected)
      (advice-remove 'company-fill-propertize 'company-tng--adjust-tooltip-highlight))
     (pre-command
--- a/company-tng.el
+++ b/company-tng.el
@@ -54,7 +54,8 @@ confirm the selection and finish the completion."
      (let ((ov (make-overlay (point) (point))))
        (setq company-tng--overlay ov)
        (overlay-put ov 'priority 2))
-     (advice-add 'company-select-next :before-until 'company-tng--allow-unselected))
+     (advice-add 'company-select-next :before-until 'company-tng--allow-unselected)
+     (advice-add 'company-fill-propertize :filter-args 'company-tng--adjust-tooltip-highlight))
     (update
      (let ((ov company-tng--overlay)
            (selected (nth company-selection company-candidates))
@@ -64,7 +65,8 @@ confirm the selection and finish the completion."
     (hide
      (when company-tng--overlay
        (delete-overlay company-tng--overlay))
-     (advice-remove 'company-select-next 'company-tng--allow-unselected))
+     (advice-remove 'company-select-next 'company-tng--allow-unselected)
+     (advice-remove 'company-fill-propertize 'company-tng--adjust-tooltip-highlight))
     (pre-command
      (when (and company-selection-changed
                 (not (company--company-command-p (this-command-keys))))
@@ -102,6 +104,14 @@ There is no need to advice `company-select-previous' because it calls
       (company-call-frontends 'update)
       t)
     )))
+
+(defun company-tng--adjust-tooltip-highlight (args)
+  "Prevent the tooltip from highlighting the current selection if it wasn't
+made explicitly (i.e. `company-selection-changed' is true)"
+  (unless company-selection-changed
+    ;; The 4th arg of `company-fill-propertize' is selected
+    (setf (nth 3 args) nil))
+  args)
 
 (provide 'company-tng)
 ;;; company-tng.el ends here
--- a/company-tng.el
+++ b/company-tng.el
@@ -35,6 +35,28 @@
 ;;
 ;; The benefits are that there is only one shortcut key to interact
 ;; with and there is no need to confirm an entry.
+;;
+;; Usage:
+;;
+;; To apply the default configuration for company-tng call
+;; `company-tng-configure-default' from your configuration file.
+;;
+;; You can also configure company-tng manually:
+;;
+;; Add `company-tng-frontend' to `company-frontends':
+;; (add-to-list 'company-frontends 'company-tng-frontend)
+;;
+;; We recommend to bind TAB to `company-select-next', S-TAB to
+;; `company-select-previous', and unbind RET and other now-unnecessary
+;; keys from `company-active-map':
+;; (define-key company-active-map (kbd "TAB") 'company-select-next)
+;; (define-key company-active-map (kbd "<backtab>") 'company-select-previous)
+;; (define-key company-active-map (kbd "RET") nil)
+;; Note that it's not necessary to rebind keys to use this frontend,
+;; you can use the arrow keys or M-n/M-p to select and insert
+;; candidates. You also need to decide which keys to unbind, depending
+;; on whether you want them to do the Company action or the default
+;; Emacs action (for example C-s or C-w).
 
 ;;; Code:
 
--- a/company-tng.el
+++ b/company-tng.el
@@ -73,6 +73,18 @@ confirm the selection and finish the completion."
        (company--unread-this-command-keys)
        (setq this-command 'company-complete-selection)))))
 
+;;;###autoload
+(defun company-tng-configure-default ()
+  "Applies the default configuration to enable company-tng."
+  (add-to-list 'company-frontends 'company-tng-frontend)
+  (let ((keymap company-active-map))
+    (define-key keymap [return] nil)
+    (define-key keymap (kbd "RET") nil)
+    (define-key keymap [tab] 'company-select-next)
+    (define-key keymap (kbd "TAB") 'company-select-next)
+    (define-key keymap [backtab] 'company-select-previous)
+    (define-key keymap (kbd "S-TAB") 'company-select-previous)))
+
 (defun company-tng--allow-unselected (&optional arg)
   "Advice `company-select-next' to allow for an 'unselected'
 state. Unselected means that no user interaction took place on the
--- a/company-tng.el
+++ b/company-tng.el
@@ -53,7 +53,8 @@ confirm the selection and finish the completion."
     (show
      (let ((ov (make-overlay (point) (point))))
        (setq company-tng--overlay ov)
-       (overlay-put ov 'priority 2)))
+       (overlay-put ov 'priority 2))
+     (advice-add 'company-select-next :before-until 'company-tng--allow-unselected))
     (update
      (let ((ov company-tng--overlay)
            (selected (nth company-selection company-candidates))
@@ -62,12 +63,45 @@ confirm the selection and finish the completion."
        (overlay-put ov 'display (and company-selection-changed selected))))
     (hide
      (when company-tng--overlay
-       (delete-overlay company-tng--overlay)))
+       (delete-overlay company-tng--overlay))
+     (advice-remove 'company-select-next 'company-tng--allow-unselected))
     (pre-command
      (when (and company-selection-changed
                 (not (company--company-command-p (this-command-keys))))
        (company--unread-this-command-keys)
        (setq this-command 'company-complete-selection)))))
+
+(defun company-tng--allow-unselected (&optional arg)
+  "Advice `company-select-next' to allow for an 'unselected'
+state. Unselected means that no user interaction took place on the
+completion candidates and it's marked by setting
+`company-selection-changed' to nil. This advice will call the underlying
+`company-select-next' unless we need to transition to or from an unselected
+state.
+
+Possible state transitions:
+- (arg > 0) unselected -> first candidate selected
+- (arg < 0) first candidate selected -> unselected
+- (arg < 0 wrap-round) unselected -> last candidate selected
+- (arg < 0 no wrap-round) unselected -> unselected
+
+There is no need to advice `company-select-previous' because it calls
+`company-select-next' internally."
+  (cond
+   ;; Selecting next
+   ((or (not arg) (> arg 0))
+    (unless company-selection-changed
+      (company-set-selection (1- (or arg 1)) 'force-update)
+      t))
+   ;; Selecting previous
+   ((< arg 0)
+    (when (and company-selection-changed
+               (< (+ company-selection arg) 0))
+      (company-set-selection 0)
+      (setq company-selection-changed nil)
+      (company-call-frontends 'update)
+      t)
+    )))
 
 (provide 'company-tng)
 ;;; company-tng.el ends here
--- a/company.el
+++ b/company.el
@@ -1840,7 +1840,7 @@ each one wraps a part of the input string."
   (interactive)
   (company--search-assert-enabled)
   (company-search-mode 0)
-  (company--unread-last-input))
+  (company--unread-this-command-keys))
 
 (defun company-search-delete-char ()
   (interactive)
@@ -1984,7 +1984,7 @@ With ARG, move by that many elements."
   (if (> company-candidates-length 1)
       (company-select-next arg)
     (company-abort)
-    (company--unread-last-input)))
+    (company--unread-this-command-keys)))
 
 (defun company-select-previous-or-abort (&optional arg)
   "Select the previous candidate if more than one, else abort
@@ -1995,7 +1995,7 @@ With ARG, move by that many elements."
   (if (> company-candidates-length 1)
       (company-select-previous arg)
     (company-abort)
-    (company--unread-last-input)))
+    (company--unread-this-command-keys)))
 
 (defun company-next-page ()
   "Select the candidate one page further."
@@ -2063,7 +2063,7 @@ With ARG, move by that many elements."
                                       0)))
           t)
       (company-abort)
-      (company--unread-last-input)
+      (company--unread-this-command-keys)
       nil)))
 
 (defun company-complete-mouse (event)
@@ -2240,10 +2240,12 @@ character, stripping the modifiers.  That character must be a digit."
             (< (- (window-height) row 2) company-tooltip-limit)
             (recenter (- (window-height) row 2))))))
 
-(defun company--unread-last-input ()
-  (when last-input-event
-    (clear-this-command-keys t)
-    (setq unread-command-events (list last-input-event))))
+(defun company--unread-this-command-keys ()
+  (when (> (length (this-command-keys)) 0)
+    (setq unread-command-events (nconc
+                                 (listify-key-sequence (this-command-keys))
+                                 unread-command-events))
+    (clear-this-command-keys t)))
 
 (defun company-show-doc-buffer ()
   "Temporarily show the documentation buffer for the selection."
--- a/company.el
+++ b/company.el
@@ -815,6 +815,11 @@ means that `company-mode' is always turned on except in `message-mode' buffers."
 (defun company-uninstall-map ()
   (setf (cdar company-emulation-alist) nil))
 
+(defun company--company-command-p (keys)
+  "Checks if the keys are part of company's overriding keymap"
+  (or (equal [company-dummy-event] keys)
+      (lookup-key company-my-keymap keys)))
+
 ;; Hack:
 ;; Emacs calculates the active keymaps before reading the event.  That means we
 ;; cannot change the keymap from a timer.  So we send a bogus command.


company-tng.el
```diff
@@ -0,0 +1,128 @@
...
+;; inserts it in the buffer. Pressing TAB again selects the second
+;; item and replaces the inserted item with the second one. This can
+;; continue as long as the user wishes to cycle through the menu.
+;;
+;; The benefits are that there is only one shortcut key to interact
+;; with and there is no need to confirm an entry.
+;;
+;; Usage:
+;;
+;; Add `company-tng-frontent' to `company-frontends':
```
aaronjensen: frontend

company-tng.el
```diff
@@ -0,0 +1,128 @@
+;;; company-tng.el --- company-mode frontend that inserts selected candidates
```
dgutov: We'll also need a NEWS entry.

nikital:
I fixed your comments and force-pushed them to nikital:company-tng
Cheers 🥂

NEWS.md
```diff
@@ -3,6 +3,7 @@
 ## Next
 
 * New user option `company-tooltip-maximum-width`.
+* New frontend `company-tng'.
```
dgutov: It's more than just a frontend, right? Let's call it "new subpackage", or a "new feature" (with an explanation).

I think NEWS is one of the main sources of users finding out about new stuff in company-mode.

dgutov:
Thanks, but this comment is still not addresses, I think:

> Could we maybe have some other summary? "frontend that inserts" doesn't make a lot of sense to me.

aaronjensen:
It might be nice to add an auto loaded command that would install company-tng with the recommended bindings.

nikital:
@aaronjensen Nice idea, I added (company-tng-configure-default) to help new users get started

@dgutov I edited the NEWS file and added more details. Regarding the comment at the top of the file: Last time I changed it to "company-mode frontend that inserts a candidate into the buffer as soon as it's selected, Vim style", but I have no further ideas to improve it. If you have anything in mind please share and I'll put it in.
Additionally, FYI I finished the copyright assignment for GNU Emacs.

dgutov:
> If you have anything in mind please share and I'll put it in.

Maybe something like "collection of settings for tab-only interaction"? It's a bit vague, though.

> Additionally, FYI I finished the copyright assignment for GNU Emacs.

Great! I'll merge this soon, then.

nikital:
> Maybe something like "collection of settings for tab-only interaction"? It's a bit vague, though.

I like it less, I think that describing a package by it's default keybinding is not very informative and I prefer to describe it by it's functionality (inserts a candidate into the buffer as soon as it's selected).

In any case, I'm really not into bikeshedding over this 😄, so if you prefer "collection of settings for tab-only interaction" I don't mind at all, just let me know and I'll update the pull request to:

    ;;; company-tng.el --- collection of settings for tab-only interaction

Have a good weekend!

dgutov:
@nikital 

> just let me know and I'll update the pull request to

Eh, it's okay. Still undecided, but I can update the summary later anyway.

Could you give me your copyright assignment number? And/or the email corresponding to it. I don't see it in the internal list, but maybe it just needs updating.

nikital:
Sorry for the delay, the assignment is finally complete. (I misunderstood the FSF process so I thought it was done before they mailed me the documents back)

nikital:
Hi, I pushed a bug fix. The completion wasn't displayed in the buffer when `company-prefix` was empty...

dgutov:
Thanks! Merging; I'll go over the documentation changes now.

nikital:
@dgutov Great to see it merged! It's my first non-trivial contribution to an open source project, so thanks for helping me take this step :)

dgutov:
@nikital You're welcome, and thanks again ;)

jojojames:
Thanks @nikital and @dgutov.

Leaving some snips here from trying it out in case it's useful for someone else.

Noticed there's an extra preview added when combining tng with preview frontend.
```
(setq company-frontends
      '(company-tng-frontend
        company-pseudo-tooltip-unless-just-one-frontend
        company-echo-metadata-frontend
        company-preview-frontend
        company-quickhelp-frontend))

;; Redefining like this works for me.
(defun company-preview-frontend (command)
  "`company-mode' frontend showing the selection as if it had been inserted."
  (pcase command
    (`pre-command (company-preview-hide))
    (`post-command
     (unless (and company-selection-changed
                  (memq 'company-tng-frontend company-frontends))
       (company-preview-show-at-point (point)
                                      (nth company-selection company-candidates))))
    (`hide (company-preview-hide))))
```

Also became tricky to expand snippets when there's more than one. Using RET for that seems like a good idea.

```
(defun +company-expand-yasnippet-or-return ()
    "Expand yas template or call RET normally."
    (interactive)
    (if (and company-active-map
             company-selection-changed
             (car company-candidates)
             (get-text-property 0 'yas-template (car company-candidates)))
        (call-interactively #'company-complete-selection)
      (when company-selection-changed
        (company-complete-selection))
      (let ((company-active-map))
        (call-interactively (key-binding (kbd "RET"))))))

  (define-key company-active-map [return] '+company-expand-yasnippet-or-return)
  (define-key company-active-map (kbd "RET") '+company-expand-yasnippet-or-return)
```

dgutov:
@jojojames

> Noticed there's an extra preview added when combining tng with preview frontend

What did you do for this to happen?

But that reminds me... @nikital, try this scenario: in an `emacs-lisp-mode` buffer, type `defal` (it should get you a single completion), then press TAB. What do you expect to happen, and what do you see?

dgutov:
@jojojames 

> Also became tricky to expand snippets when there's more than one

Less obvious maybe, but tricky how? You select the desired snippet using TAB the same way as you pick a non-snippet completion.

jojojames:
>> What did you do for this to happen?
I type a few characters (with those frontends set), a popup appears with list of completions. I press tab once, tng draws the completion into view. Preview frontend draws the preview after the current completion. Maybe it's my set up also.

>> Less obvious maybe, but tricky how? You select the desired snippet using TAB the same way as you pick a non-snippet completion.
Hmnn, this is the scenario I was thinking of.

Popup
[Completion 1
Completion 2
Completion 3 (yasnippet)]

I can select the snippet by tabbing to completion 3 but not sure how to expand it? If I press tab again, it'll loop the selection. If I press enter, it seems to expand and insert a new line into the snippet.

aaronjensen:
> I can select the snippet by tabbing to completion 3 but not sure how to expand it?

You'll need a specific binding to expand it, maybe `M-/`. The concept of tng and snippet expansion are a bit incompatible, but adding an additional binding just for the snippet expansion should work.

jojojames:
> You'll need a specific binding to expand it, maybe M-/. The concept of tng and snippet expansion are a bit incompatible, but adding an additional binding just for the snippet expansion should work.
Yup! Using the RET key and setting up a fallback keybind seems to be DWIM instead of using something like M-/.

dgutov:
> I can select the snippet by tabbing to completion 3 but not sure how to expand it?

Why not start typing whatever you would have been typing after the snippet is expanded? If it's expands into a method call, start typing the first argument. I agree it's not intuitive (and I've love to hear the ideas from the Vim/ycmd world), but it should work and be efficient enough, as soon as you get used to it.

Ultimately, though, `company-tng-frontend` might even automatically "expand" the snippets, just like it "inserts" completions now.

drot:
> Preview frontend draws the preview after the current completion.

Can confirm. Starting with emacs -Q and evaluating the following in order:


company

Last line is what happens when you press TAB while the preview frontend is active.

dgutov:
@nikital In short, I think when there is only one candidate in the list, `company-tng--allow-unselected` should call `company-complete-selection` somehow (not sure how to implement this best with the current advice approach). That would solve the problem with the frontends automatically.

nikital:
### TNG + Preview
I see the problem. I think that we should change the default configuration for TNG:

    (defun company-tng-configure-default ()
      "Applies the default configuration to enable company-tng."
      (setq company-frontends '(company-tng-frontend
                                company-pseudo-tooltip-frontend
      ...

1. It doesn't include the preview frontend. I think that preview is not needed when TNG is active because TNG already does the same thing (previews selected candidates in-buffer)
2. It uses `company-pseudo-tooltip-frontend` instead of `company-pseudo-tooltip-unless-just-one-frontend`. Vim's/ycm's default behavior is to show the tooltip even if there is one candidate and the user can just hit TAB to select it. I think that it makes sense because one candidate and many candidates are presented in the same way and TAB work consistently between them.

### Snippets
I'm in @aaronjensen's camp on snippet expansion and I prefer to use a dedicated hotkey instead of multiplexing it on another key, but I'm glad that @jojojames's `+company-expand-yasnippet-or-return` works as well!

> Ultimately, though, company-tng-frontend might even automatically "expand" the snippets, just like it "inserts" completions now.

That's a nice concept, I like the general idea but I'm not familiar with yasnippet at all so I have no feedback on possible implementation... I also wonder what should happen during regular completion if the user TABs over a really big snippet. Will we just temporarily preview the entire expanded snippet in the buffer? Could create an unpleasant visual flash...

jojojames:
> Why not start typing whatever you would have been typing after the snippet is expanded?

Hmnn, not sure I understand this workflow. I do like the RET to confirm the snippet expansion though. So I'll just stick with that.

> Ultimately, though, company-tng-frontend might even automatically "expand" the snippets, just like it "inserts" completions now.

This seems cool but I think yasnippet also takes control of the tab key when it's active so it might be a little awkward cycling through completions as well as going to the next part of the snippet.

> It doesn't include the preview frontend. I think that preview is not needed when TNG is active because TNG already does the same thing (previews selected candidates in-buffer)

I like the preview while typing and I think TNG only comes into play after the first selection. It'd be nice if the frontends played nice with each other when combining them.

>  It uses company-pseudo-tooltip-frontend instead of company-pseudo-tooltip-unless-just-one-frontend. Vim's/ycm's default behavior is to show the tooltip even if there is one candidate and the user can just hit TAB to select it. I think that it makes sense because one candidate and many candidates are presented in the same way and TAB work consistently between them.

Changing the defaults is fine but I actually think it's a limitation of vim (IIRC it can't show a preview like Emacs does). This might be a case of reproducing ycmd/vim behavior faithfully or doing what's best (whatever that means).

>  I agree it's not intuitive (and I've love to hear the ideas from the Vim/ycmd world)

Having used vim for a while with ycmd, I've never found ycmd+ultisnips (not sure about new packages) to have great synergy.

dgutov:
@nikital 

> we should change the default configuration for TNG

If you're confident in that solution, please go ahead and make a PR.

> what should happen during regular completion if the user TABs over a really big snippet

We could truncate it to the first line, maybe (with an ellipsis as a hint), as one option.

> I'm not familiar with yasnippet at all so I have no feedback on possible implementation

Neither am I familiar with it to the extent required.

dgutov:
@jojojames

> Hmnn, not sure I understand this workflow.

Select the completion that comes from `company-yasnippet` and press some key like SPC. But I agree that it's awkward, not seeing where that space will end up.

> it might be a little awkward cycling through completions as well as going to the next part of the snippet

Good point. Ideas welcome here as well.

But using some extra key (RET or otherwise) seems inevitable indeed.

nikital:
@jojojames 
> Changing the defaults is fine but I actually think it's a limitation of vim (IIRC it can't show a preview like Emacs does). This might be a case of reproducing ycmd/vim behavior faithfully or doing what's best (whatever that means).

I'm don't think we need to reproduce Vim, but for now sticking to Vim default is the easy fix that I PRed in #712.

If you want to integrate preview into the mix, I would start with binding tab to something like this and see if it behaves intuitively (didn't test it, just wrote it off the top of my head):
```
(defun company-select-next-or-complete (&optional arg)
  "Select the next candidate if more than one, else complete.

With ARG, move by that many elements."
  (interactive "p")
  (if (> company-candidates-length 1)
      (company-select-next arg)
    (company-complete-selection)))

```

rswgnu:
Although C-g can be used to abort company completion, often one's muscle memory (from finishing prompted for arguments or entry of forms) will reach for RETURN to finish things up.  Since the default configuration of tng does not bind RETURN to anything, why not make it run company-abort.  Then if you don't want to select a completion from the popup, a quick RETURN
makes it disappear.  Also this requires only a one finger press instead of two.

jojojames:
I believe that's one of the biggest benefits of TNG, that RET isn't bound to anything.

You can bind conpany-abort to company active map if you want though.

rswgnu:
Actually, I don't want RETURN to do a company-abort (because that removes any selected completion).  I just want it to leave whatever completion has been added to the buffer and remove the company completions popup and not do anything else.  Is there an existing command to use for that?

aaronjensen:
`keyboard-quit`?

rswgnu:
No, that does not cleanly terminate from the company popup overlay.  We need a command that leaves any previously selected completion in place, does not select one unless the user has manually selected one and then cleanly terminates from company completion mode.

rswgnu:
Really the company-finish function should be a command which inserts into the buffer any partial or full completion that has been selected (I think tng just adds an overlay with the completion and doesn't do an insertion) but should not select the first completion entry if the user has not made such a selection.

rswgnu:
This is what I had to write to get the above behavior.  Feel free to add it and make any modifications you may require.  I have not yet found any problems with it, though.

```
(defun company-quit ()
  "Insert any selected completion and quit completing."
  (interactive)
  (when (and company-selection-changed company--manual-action
	     (boundp 'company-tng--overlay) company-tng--overlay)
    (company--insert-candidate
     (nth company-selection company-candidates)))
  (company-cancel))
```

rswgnu:
This would be simpler if company-selection were not set to the 0th item in the candidates list prior to anything being selected.  Maybe it could be nil at that point.  Then it could be tested as to whether anything is selected or not.  Or is there a simpler test for that?  And how does one de-select any item but leave the selection popup on screen?

rswgnu:
Wait for an update on this code; I did find an issue.

rswgnu:
The code does seem to work fine here.

nikital:
I think it's a good idea to C-g's behavior consistent with ESC in Vim/Evil (leave the inserted text). I didn't test it extensively, but I think @rswgnu's code can be simplified to:
```elisp
(defun company-tng-complete ()
  (interactive)
  (if company-selection-changed
      (company-complete-selection)
    (company-abort)))

(define-key company-active-map (kbd "C-g") 'company-tng-complete)
```
I will test it for some time on my own config to catch possible problems and then open pull request. Comments are welcome :)

jojojames:
@nikital I've always thought of C-g as 'get out of jail' and not 'execute some custom code related to the current package/keymap that's related to aborting'.

I'm not sure it's so important to match C-g with vim/evil's in the first place since vim/evil don't use C-g as an escape key anyways.

nikital:
I understand what you mean.

However, because in company-tng the candidate gets inserted into the buffer as soon as it's selected, it seems to me that leaving the candidate inside the buffer is the better "get out of jail" experience than deleting a previously inserted candidate.

jojojames:
I've ben using C-g to be more of a 'cancel all completions *even* if I've selected one'.

Sometimes I scroll through the list just to see what's in it (not uncommon) but aren't interested in completing anything.

dgutov:
> leaving the candidate inside the buffer is the better "get out of jail" experience

I disagree.

You can bind that to RET, though, if that's the consensus.

nikital:
I see... I didn't think about your use case (just scrolling through
candidates and then aborting if you don't find any), it makes sense to
leave the current default behavior then.

Regarding RET, I agree with Aaron that having it unbound is very important,
especially in languages such as lisp or python that have no semicolons at
the end of the line, which makes it likely to have the tooltip active when
you are at the end of the line and you want RET to actually insert a new
line.

I think the consensus is to leave things as is for now :)

rswgnu:
On Tue, Dec 5, 2017 at 10:26 AM, Nikita Leshenko <notifications@github.com>
wrote:

> I see... I didn't think about your use case (just scrolling through
> candidates and then aborting if you don't find any), it makes sense to
> leave the current default behavior then.
>

​Yes, C-g works well as a completion abort.​

> ​​
>
> ​​
> Regarding RET, I agree with Aaron that having it unbound is very important,
> ​​
> especially in languages such as lisp or python that have no semicolons at
> ​​
> the end of the line, which makes it likely to have the tooltip active when
> ​​
> you are at the end of the line and you want RET to actually insert a new
> ​​
> line.
>

​That is certainly one possibility.  But generally when a selectable
popup is on-screen​, one must deal with it before going on to do
other editing.  Since RET selects an item when over one and removes
the popup, it makes sense that it would remove the popup without selecting
one when not over and item and then like all keys bound in
company-active-mode
return to its normal binding afterwards.  I use Python and Lisp all the time
and have no problem having RET set to this behavior.

Similarly, C-n and C-p are much more natural for touch typists to use to
move through
the items and more consistent with typical Emacs bindings than using M-n
and M-p (I am
suggesting these bindings be made).  You don't move around in a buffer in
the middle of
completing because then completion won't work properly.  With this setup,
C-n, C-p, C-g
and RET do everything you need without use of any more hard-to-reach keys
and that would
improve speed more than what their current uses do during company mode.

​
>
> ​​
> I think the co
> ​​
> nsensus is to leave t
> ​​
> hings as is for now :)


​If so, so be it, I just want to stimulate discussion and thought about
possible improvements.

​

aaronjensen:
> it makes sense that it would remove the popup without selecting
> one when not over and item and then like all keys bound in
> company-active-mode

@rswgnu I believe you're describing (more or less) the original company behavior. company-tng was requested and built by people who want exactly what you are saying you don't want. We want to be able to have the selection of the item (via tab/shift-tab) be the confirmation. The extra keypress to select is specifically what we don't want. 

It sounds like you maybe want some hybrid, but I'm not sure exactly what it is that you want that the original company behavior does not support.

rswgnu:
On Tue, Dec 5, 2017 at 11:35 AM, Aaron Jensen <notifications@github.com>
wrote:

> it makes sense that it would remove the popup without selecting
> one when not over and item and then like all keys bound in
> company-active-mode
>
> @rswgnu <https://github.com/rswgnu> I believe you're describing (more or
> less) the original company behavior. company-tng was requested and built by
> people who want exactly what you are saying you don't want. We want to be
> able to have the selection of the item (via tab/shift-tab) be the
> confirmation. The extra keypress to select is specifically what we don't
> want.
>
> It sounds like you maybe want some hybrid, but I'm not sure exactly what
> it is that you want that the original company behavior does not support.
>
​The issues I am discussing are:
  What does RET do when no popup completion is visibly selected?  I want it
to quit from company-mode and leave any partial completion that was done
already in the buffer; nothing more or less.
  If RET is pressed on a popup completion, it should be inserted into the
buffer as the completion and company-mode quit.  I believe the default
behavior of company-mode is to invisibly select
  the first potential completion item, so an immediate RET adds that item
as the completion in the buffer.  I never want that behavior and I find it
wastes a lot of time correcting things.
  Often I may want to look at completions without doing any completion.  I
find RET much simpler to use than C-g in such cases.  I want to use C-g
only to abort after seeing some completion
  added to the buffer.

  I also suggested C-p and C-n be bound to move through popup completion
items.

  I like having TAB perform completions as well, so I use company-tng.  But
TAB just moves between popup items and displays each potential completion
in the buffer; it doesn't
  terminate the popup or company-mode.  How could it do that and still move
through the popup selections?  You still have to press RET or some
non-company-mode key to have the
  selection permanently added to the buffer, right?

Bob

aaronjensen:
> What does RET do when no popup completion is visibly selected?

Under company-tng, it closes the popup and inserts a return. This is the behavior that other editors/completion engines have that company-tng is emulating. 

If you want different behavior you'll probably need to write it. It doesn't make sense to change company-tng on this and it'd be up to @dgutov if he wants to change the default behavior of RET, which, at the moment, does not "invisibly select the first potential candidate" as that candidate is already selected. Company selects it immediately when it pops up. company-tng doesn't do this, which is perhaps why you're thinking that it invisibly selects it.



> I also suggested C-p and C-n be bound to move through popup completion
> items.

You can bind these in your init. I have them bound to this as well, I don't know where that's coming from--perhaps spacemacs, but I use tab/shift-tab generally any way. I may actually unbind them because I want to free those up to move around.




> I like having TAB perform completions as well, so I use company-tng.  But
> TAB just moves between popup items and displays each potential completion
> in the buffer; it doesn't
>   terminate the popup or company-mode.  How could it do that and still move
> through the popup selections?  You still have to press RET or some
> non-company-mode key to have the
>   selection permanently added to the buffer, right?

That's right, and this is exactly the behavior that all of who asked for and contributed to company-tng wanted.

Again, it sounds like you want a hybrid, specifically you want to be able to press RET when nothing has been selected to dismiss company-mode or to (actually) insert the candidate if one is selected without also inserting a newline. It looks like you wrote a function to do what you want. Perhaps if you want to pull request that function @dgutov may be interested in having it built-in so others could bind it to whatever they wanted, if they wanted to be used in conjunction with `company-tng`. I would say it should not be the default, however.

rswgnu:
On Tue, Dec 5, 2017 at 2:15 PM, Aaron Jensen <notifications@github.com>
wrote:

> Company selects
> ​[the first completion candidate]​
> immediately when it pops up. company-tng doesn't do this, which is perhaps
> why you're thinking that it invisibly selects it.
>
> ​With company-tng loaded, before I customized anything, once a completions
list popped up, nothing was selected, but if I pressed return, the first
candidate was inserted into the buffer.
So that did appear to be as an invisible selection of the first candidate,
as you noted.

> Aga
> ​​
> in, it sounds like you want a hybrid, specifically you want to be able to
> press RET when nothing has been selected to dismiss company-mode or to
> (actually) insert the candidate if one is selected without also inserting a
> newline.
>
> ​Yes, said more simply, I want consistent behavior which is that RET
inserts any selected candidate and quits from company-mode.  company-tng is
not consistent since it inserts a non-selected candidate when RET is
pressed prior to selecting any candidate.
​​
It lo
​​
ok
​​
s like you wrote a function to do what you want.

Yes, I am only writing to encourage a change in defaults.

> Perhaps if you want to pull request that function @dgutov
> <https://github.com/dgutov> may be intereste
> ​​
> d in having it built-in so others could bind it to whatever they wanted,
> if they wanted to be used in conjunction with company-tng. I would say it
> should not be the default, however.
>
> ​That's a reasonable idea but based on past experience I would ask Dimitri
to express some interest in this before I do anything further with it.  If
he is interested, I will do it.

Thanks for your thoughts on this.
​​

aaronjensen:
> With company-tng loaded, before I customized anything, once a completions
> list popped up, nothing was selected, but if I pressed return, the first
> candidate was inserted into the buffer.
> So that did appear to be as an invisible selection of the first candidate,
> as you noted.

did you invoke `(company-tng-configure-default)`? It unmaps RET and it should not do what you're describing.

rswgnu:
On Tue, Dec 5, 2017 at 4:57 PM, Aaron Jensen <notifications@github.com>
wrote:

> With company-tng loaded, before I customized anything, once a completions
> list popped up, nothing was selected, but if I pressed return, the first
> candidate was inserted into the buffer.
> So that did appear to be as an invisible selection of the first candidate,
> as you noted.
>
> did you invoke (company-tng-configure-default)?
>
​Yes.​

> It
> ​​
> u
> ​​
> nmaps RET and it should not do what you're describing.
>
> ​I saw this behavior many times before I fixed it.  I should note that
sometimes I am using the jedi backend for Python and its use of deferred
function calls leads to all sorts of oddness at times.

But I basically loaded company and company-tng out-of-the-box, set them up
as recommended, turned on global company mode, and bound the following keys
like so:

  (global-set-key [remap completion-at-point] 'company-complete)
  (global-set-key [remap complete-symbol] 'company-complete)
  (global-set-key [remap indent-for-tab-command]
'company-indent-or-complete-common)


With that I observed the problematic RET behavior.  To fix it, in
company-active-map, I had to bind RET to this function that I wrote:

  (defun company-quit ()
    "Insert any selected completion and quit completing."
    (interactive)
    (when (and company-selection-changed company--manual-action
       (boundp 'company-tng--overlay) company-tng--overlay)
      (company--insert-candidate
       (nth company-selection company-candidates)))
    (company-cancel))

Bob

jojojames:
company-active-map

company-backends

company-frontends

company-diag (this is a function)

It might be more productive to give values of those^ variables.

That global-set-key for the indent-for-tab-command looks suspicious though. Why do you need to rebind those keys? Part of the tng-setup is rebinding keys for you.

rswgnu:
On Tue, Dec 5, 2017 at 6:58 PM, James N <notifications@github.com> wrote:

> That global-set-key for the indent-for-tab-command looks suspicious
> though. Why do you need to rebind those keys? Part of the tng-setup is
> rebinding keys for you.
>
​I just tested without indent-for-tab-command rebound and in that case, I
don't get a completions popup when I press tab after:
​

jojojames:
I'm coming to the conclusion that the behavior you're seeing might not be how company-tng is expected to behave?

It might be better to debug with emacs -q and/or provide those ^values from above and/or bisect your config and see what's going on.

>  I think the consensus is to leave things as is for now :)

+1

rswgnu:
On Tue, Dec 5, 2017 at 8:09 PM, James N <notifications@github.com> wrote:

> I'm coming to the conclusion that the behavior you're seeing might not be
> how company-tng is expected to behave?
>
​So by default with tng enabled, if no completion is yet selected, pressing
RET is supposed to simply quit from company-mode and add a newline to the
buffer, right?​

​​

That is not the behavior I remember seeing though.

> It migh
> ​​
> t be better to debug with emacs -q and/or provide those ^values from above
> and/or bisect your config and
> ​​
> see what's going on.
> ​​
>
> ​Yes, I had started doing that but haven't had time to finish tracing
through things.

rswgnu:
On Tue, Dec 5, 2017 at 6:58 PM, James N <notifications@github.com> wrote:

> That global-set-key for the indent-for-tab-command looks suspicious
> though. Why do you need to rebind those keys? Part of the tng-setup is
> rebinding keys for you.
>

I j
​​
ust
​​
tested without indent-for-tab-command rebound and in that case, I don't get
a completions popup when I press tab
​in python-mode ​
after:
​
sys.​
​With that key remapped, I see the expected popup showing features of the
sys module.​
​
Bob

nikital:
@rswgnu I couldn't understand your last message because your email client messed up the formatting. Could you fix it or use Github's web interface?

> So by default with tng enabled, if no completion is yet selected, pressing RET is supposed to simply quit from company-mode and add a newline to the buffer, right?​

Yes. Since the idea of company-tng is to make TAB/S-TAB cycle through candidates and insert them as they are selected, you can, theoretically, unbind all company-mode keys except TAB and it would still work. RET isn't a special key in company-tng, that's why it's unbound by default.

As other have mentioned, apparently you have another package that binds RET back and creates this weird behavior. The "unselected" state is something specific to company-tng, other packages will get confused by it and might insert the first candidate even if it's not selected. You should unbind RET if you want a good company-tng experience.

Also, if you can open a new issue with a minimal config that reproduces your out-of-the-box experience I might be able to modify company-tng to provide a better integration (For example maybe we should advice `company-complete-selection` to make it aware of the unselected state when company-tng is active). Please @ mention me on the new issue.

terlar:
First of all really insightful discussions, and it had lead me into refining my setup and really understand what I would want.

I almost have what I want with `company-tng` and `company-preview`, I want to use `company-preview` before anything is selected by `company-pseudo-tooltip`. I think it is most easily explainable by a short gif.

My current configuration:
```
(defun company-preview-if-none-selected-frontend (command)
  "`company-preview-frontend', but only when nothing is selected."
  (message "%s" company-selection)
  (unless (and (eq command 'post-command)
               (> company-selection 0))
    (company-preview-frontend command)))
(setq company-frontends
      '(company-tng-frontend
        company-pseudo-tooltip-unless-just-one-frontend
        company-preview-if-none-selected-frontend
        company-echo-metadata-frontend)
```

company-demo

The reason why I want to have the preview is that I find it easier to see, ah this is the text I want and then I just accept directly, however when I start selecting from the menu I am fine with that insertion. So I wonder how to solve a preview that only is shown before you enter the pseudo-tooltip box. As shown by the video there are two different states where both have `company-selection` = `0`. You can see it as I select-previous out of the box and nothing is selected, then techincally it is still `0` and that is the case when I would want `company-preview` to be active.

Is it possible to differentiate by those states, or is it only internal to the pseudo tooltip?

jojojames:
@terlar What theme/font are you using?

Also, did you try my workaround (mentioned above.) That should fix your problem.

I do think we should try to get company-preview working with company-tng. I've seen this asked a couple times now and it's definitely a source of confusion.

terlar:
@jojojames Thank you, that works very well. I just ignored it since there was a lot of discussion of the problem after your function post with other solutions, so I assumed it didn't work.

I rewrote it as such (to not having to reproduce logic):
```elisp
  (defun company-preview-if-not-tng-frontend (command)
    "`company-preview-frontend', but not when tng is active."
    (unless (and (eq command 'post-command)
                 company-selection-changed
                 (memq 'company-tng-frontend company-frontends))
      (company-preview-frontend command)))
```

@jojojames The theme I am using is [tao-theme](https://github.com/11111000000/tao-theme-emacs) using the `tao-yang` version with some [custom overrides](https://github.com/terlar/emacs.d/blob/dcf455288ea834beb71a43db3d29c540c91af97e/lisp/local-theme.el). The font is [Fira Mono](https://github.com/mozilla/Fira). :smile_cat:

Issue #717, opened 2017-10-04
Emacs 26 threads?


aaronjensen:
Hi, I was wondering if any consideration has been put into whether or not the new lisp thread system would be able to help to make lisp backends async? I'm thinking of things like company-dabbrev that can sometimes cause a slight stutter. In an ideal implementation I'd be able to set `company-idle-delay` to 0 for all backend combinations and not notice any stutter.

dgutov:
Most likely not.

Emacs threads are not concurrent (only one Lisp threads runs at a time), so I don't see how it can speed up dabbrev in any way. Using an external process maybe, but that's a different architecture.

aaronjensen:
Perhaps not speed it up, but could it yield to input in any of its looping?

dgutov:
That's an interesting question. But what would it do when control returns to it after input? The collected results are already invalid, strictly speaking.

On the other hand, even without threading, we could add an `input-pending-p` condition to `company-dabbrev--search-buffer`.

aaronjensen:
They may not be completely invalid, they may be a superset of the valid results. For example, if I type "exclamation", and company kicks in at "ex" and finishes after I type "c", filtering it down further would be faster. I know that different fuzzy finders may need to handle that differently, but maybe it'd work in many cases.

Also, I think that the stutter I was seeing recently was coming from backends other than `company-dabbrev`, or from a combination of many backends running. This is what I had:

```lisp
'((company-dabbrev-code company-gtags company-etags company-keywords)
    company-files 
    company-dabbrev)
```

By removing `(company-dabbrev-code company-gtags company-etags company-keywords)` it sped things up. I'll run at `company-idle-delay` 0 and for a bit and see how it goes. 

That said, I have a fast computer and I think that this would generally be a good improvement that could be applied to many of the lisp backends.

dgutov:
> They may not be completely invalid, they may be a superset of the valid results

That maybe be true. So if you want to try your hand at a threads-based backend, please go ahead. Without threads, I think we can also use per-buffer caches for a close-enough improvement.

> the stutter I was seeing recently was coming from backends other than company-dabbrev

From `company-dabbrev-code`, then? It shares the implementation and has the same tradeoffs.

aaronjensen:
> From company-dabbrev-code, then? It shares the implementation and has the same tradeoffs.

Possibly yeah, I wasn't sure if it was one of the others (or the combination of all of them working at once).

[...]

PR #720, opened 2017-10-06
Abort dabbrev search on input pending

aaronjensen:
Unlike `while-no-input`, this appears to not consume input and it actually helps with the lag.


aaronjensen:
Related to #717

dgutov:
Thanks. Let's try this then, and please report any unforeseen effects.

aaronjensen:
Will do, thanks for the merge.

PR #737, opened 2017-11-11
Change to container-based Travis-CI build

iquiw:
Supports emacs24, emacs25 and emacs-snapshot.

I am not sure about Pros. and Cons. of [container-based build](https://docs.travis-ci.com/user/reference/trusty/#Container-based-with-sudo%3A-false).

If you don't like the change, we can just add emacs25.
https://github.com/iquiw/company-mode/commit/5ff576581bd2d7adbc178a8ba761dafd5c385769

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,24 +1,26 @@
-# https://github.com/rolandwalker/emacs-travis
+sudo: false
 
 language: generic
 
-env:
-  matrix:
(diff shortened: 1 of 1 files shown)


dgutov:
Finally :)

When I tried container-based builds before, these APT sources were not allowed. Thanks!

PR #740, opened 2017-11-22
Don't restart semantics after ::

Havner:


dgutov:
Makes sense, I guess. Thanks!

Issue #731, opened 2017-10-19
Company tng map should address clash with templates map


rski:
I've added (company-tng-configure-default) to my init like the docs in company-tng.el recommend, but the way TAB is bound, it clashes with TAB inside templates.

Consider this really contrived case:

![2017-10-19-175507_316x49_scrot](https://user-images.githubusercontent.com/2960312/31783590-aa91807c-b4f6-11e7-9a3a-793f095fff32.png)

Here, my instinctive reaction is to press tab to select fmt, but because the point is inside a template, it jumps to the next field. I've solved it for go by telling the backend not to insert a template, but the docs should probably be updated with a recommendation to work around this and tng-configure-default should do something about it too.

I would solve it by unbinding TAB from the template map, but what should forward-field be bound to instead?

dgutov:
It's an old problem, but it's a lot more noticeable with tng, of course.

Yasnippet's keymap takes priority because it's set using an overlay. Last I checked, setting it this way doesn't work at the last buffer position, so that's why we still use the current option. Keeping both approaches and switching between them dynamically could work, though.

Maybe some rebinding would satisfy most people, though. Problem is, we'd have to do it for both `company-template` and `yasnippet`, which would be pretty invasive. Paging @nikital.

nikital:
Hi, good morning :)

I'm not using snippets at the moment so I'm not sure what the desired default behavior should be...

1. TAB will select the candidate—`company-tng-configure-default` will unbind the "next template field" keybindings.
2. TAB will move to the next template field—`company-tng` will deactivate itself when a template is active.
3. TAB will do some sophisticated DWIM action? I can't think of a reasonable DWIM behavior.
4. `company-tng-configure-default` will disable templates entirely by default to avoid out-of-the-box confusion for new users.

I lean towards option 1 and asking the user to bind some other key for "next field". In Vim I bind C-j for snippet expansion and next template field, while keeping TAB exclusively for completions.

Which option sounds the most reasonable to you? Do you see other options?

rski:
Heya @nikital 

I'm not using the yas backend either, so it's good that @dgutov brought it up because it would've completely slipped my mind.

Putting aside yas for a second, I think that (1) is the most reasonably consistent behaviour where you don't have to do the mental shift from "TAB completes" to "i'm in a template now, TAB does another thing".

[...]

PR #764, opened 2018-03-10
company-tng: Suppress post-completion

nikital:
See commit messages and the discussions at #731


dgutov:
I see you've decided to go with advice again (I figured it would be a custom var which `company-tng-configure-default` would set). Anyway, it should work fine.

Let's see if somebody reports this to be not to their liking.

nikital:
@dgutov Yes I went with an advice because I prefer not to refactor the generic company code until we iron out all the potential compatibility issues between company and company-tng and make sure that the two can coexist peacefully :)

PR #769, opened 2018-03-15
Always get candidates with prefix string in `company-diag`

### LINKED ISSUE #768: company-diag call backend with raw prefix
If the backend returns prefix as ("foo" . t), company should use "foo" to fetch the candicates, isn't it? But in company-diag, it is not, which make backend report error.

tigersoldier:
Fixes #768


dgutov:
Thanks!

Issue #142, opened 2014-06-19
company-capf does not pass control to the next back-end


luciferasm:
**Version:**
0.8.1-cvs (downloaded from MELPA as folder called "company-20140609.753")

**To reproduce bug (in terminal):**

> emacs -Q -nw
> (package-initialize)
> (require 'company)
> (global-company-mode)
> foobarqux
> foob<- point here

**Expected result:**
foobarqux should be offered as the only candidate.

**Actual result:**
No candidates are offered.

**Speculation of why:**
`company-backends` has entry `company-capf` in the middle, and ends with entry `company-dabbrev`. `company-capf` is apparently working, since it offers the candidates "footnote" and "footnote-mode <f>" when typing "foo". But when typing a "b" after "foo" in the effort to expand to "foobarqux" in the example above, `company-capf` does not seem to pass control to the next back-end though. In this case it should fall through several back-ends until finally reaching `company-dabbrev`, that could successfully take care of the desired expansion.

At closer inspection, the responsible code could maybe be found in the `company-capf`'s prefix clause, and specifically in the behavior of `company--capf-data`. `company-capf`'s prefix clause asks `company--capf-data` if there are any candidates. `company--capf-data` should return nil in our case, which would then pass control to the next back-end. But it seems like `company--capf-data` only return nil after characters that does not belong to the word or symbol constituent syntax classes. To illustrate, evaluate:

> (global-set-key [(f5)] (lambda () (interactive) (message "%s" (company--capf-data))))

When pressing f5 after any word or symbol, it returns a long list (i.e. a truthy value). It only returns nil after e.g. ; or a blank line.

Wilfred:
For what it's worth, I've created [company-try-hard](https://github.com/Wilfred/company-try-hard) to explore this idea. This enables you to exhaustively try each backend for candidates.

dgutov:
@Wilfred That's terrific. I'm glad you were able to implement it with not too much effort.

NightMachinary:
> For what it's worth, I've created [company-try-hard](https://github.com/Wilfred/company-try-hard) to explore this idea. This enables you to exhaustively try each backend for candidates.

It seems this still doesn't work when the first backends encounter an error?

dgutov:
@NightMachinary Do you mean encounters a bug? Backends are not supposed to signal errors.

NightMachinary:
[...]

PR #799, opened 2018-06-13
company-yasnippet return nil as prefix when company-grab-symbol returns empty string

wyuenho:
Checks all the trigger keys for all the templates enabled for the current major mode to, if one matches `company-grab-symbol`, return it. This is reasonably fast because the snippet tables are cached, so this is all done in memory.

This implementation returns much more accurate completions and since now prefix may be nil, other backends that come after company-yasnippet will have a chance at completion.

### DIFF
--- a/company-yasnippet.el
+++ b/company-yasnippet.el
@@ -62,6 +62,11 @@
             (push prefix prefixes))))
       prefixes)))
 
+(defun company-yasnippet--prefix ()
+  (cl-find (company-grab-symbol)
+           (cl-mapcan 'yas--table-all-keys (yas--get-snippet-tables))
+           :test 'string=))
+
 (defun company-yasnippet--candidates (prefix)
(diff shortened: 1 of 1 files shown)


wyuenho:
Hmmm, this is still not ideal, let me try matching the return value of `(company-grab-symbol)` with actual trigger keys available under the mode.

wyuenho:
@dgutov this should be ready now.

wyuenho:
Ahhh I just noticed #142 , @dgutov what do you think about this PR?

dgutov:
I'm not sure this is a definite improvement. How long does `company-yasnippet--prefix` take? The prefix command is supposed to be fast.

Aside from that, so what if the current prefix is one of the snippet keys? The user might want to have some other backend to be used anyway.

This also probably breaks the case where the prefix is a substring of one of the keys.

PR #802, opened 2018-06-14
Update Javascript and Python keywords

wyuenho:
I've also aliases rjsx-mode to javascript-mode


dgutov:
Any idea why the build fails?

wyuenho:
It doesn't anymore. There was a duplicate keyword in the python list.

PR #830, opened 2018-09-13
Define variable alias before referent

jabranham:
To avoid Emacs bug #5950: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=5950


dgutov:
Thanks!

PR #858, opened 2019-01-09
company: company--company-command-p: Fix lookup-key

### LINKED ISSUE #857: company-tng updates the buffer too late/gets cancelled?
I'm occasionally caught out by using company-tng to complete something as the last thing and invoking a command, and the completion getting lost. For example:

Completing the last word in a magit commit message and typing `C-c C-c` to commit.
Completing a variable name and typing `C-x C-s` to save.

nikital:
Make sure it returns an actual command.

`lookup-key` returns a number when the key sequence is longer than 1 but it
isn't a valid sequence of prefix characters. The number evaluates to true but
the key isn't in the keymap.

When that happens the completion gets lost. To reproduce, complete a variable
name and type C-x C-s to save.

Fixes #857

Reported-by: Thomas Fini Hansen <https://github.com/xendk>

### DIFF
--- a/company.el
+++ b/company.el
@@ -832,7 +832,7 @@ means that `company-mode' is always turned on except in `message-mode' buffers."
 (defun company--company-command-p (keys)
   "Checks if the keys are part of company's overriding keymap"
   (or (equal [company-dummy-event] keys)
-      (lookup-key company-my-keymap keys)))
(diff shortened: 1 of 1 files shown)


dgutov:
Very quick. Thank you!

PR #873, opened 2019-02-17
Add tallies of and echo frontend for backend candidate counts

rswgnu:
I really wanted to be able to see which backends triggered during completion and how many candidates they each contributed to the candidate pool.  The existing code did not provide a simple way to generate this information so I adapted it to do so and added an echo frontend feature so you can dynamically see what is being generated at each step, if desired.

To use, either:
   1. add company-echo-completion-counts-frontend to company-frontends for dynamic display
or
   2. complete or attempt to complete something in a buffer and while completion is active, press
       C-# to see the backend candidate counts
or
   3. Finish or abort completion and then use M-x company-show-completion-counts RET

-----
Although company-diag exists already, it doesn't provide the same kind of insight into what is happening especially when multiple backends generate one set of candidates.  Maybe this can also be used to extend the output of company-diag.

-----
I neglected to see you had automated tests for company-mode, so I did not run them prior to submitting this.  **After a quick look, I cannot follow why the tests are failing on many lambda expressions that I did not touch.  Someone with more experience with ert should run the tests and debug if possible.**

Thanks for company-mode.


dgutov:
I'm not sure about merging this.

Who would use such a frontend?

If you wanted to expand the output of `company-diag`, though, be my guest. I think the patch would also be smaller in that case.

rswgnu:
I think the description explains how this helps.  Any developer or user who wants to be able to be able to see or explain which backends are generating most or least completions could easily use it.  I did not see a simple way to extend company-diag with this information or I would have done that but now that it is written maybe it could be integrated.

dgutov:
>  I did not see a simple way to extend company-diag with this information or I would have done that

Actually, the completions list by itself already contains that info (the `company-backend` text property on the strings). When the property is absent, that means that the completion came from the first backend in the group.

So if you just want the counts, adding them to `company-diag` should be pretty easy.

PR #877, opened 2019-03-02
add variable companny-etags-update-interval

redguardtoo:
`company-etags-update-interval` is a minimum interval to read tags files.
Emacs could be less responsive when a huge tags file is parsed by funciton `tags-completion-table'.

You could use ctags to create one big tags file for linux kernel source (https://www.kernel.org/) to reproduce  the issue. The old version reads from phyiscal tags file too frequenly.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,13 @@
 # History of user-visible changes
 
+## Next
+
+* Snippet/template field interaction is inhibited while completion is active
+  (where by default `TAB` calls `company-complete-common`, clashing with snippet
+  map binding `TAB` to "jump to the next field"). Affects both
+  `company-template` and `yasnippet` (requires changes from 2019-04-21,
+  currently unreleased).
+
 ## 2019-04-15 (0.9.10)
 
 * `company-clang`: better compatibility with Clang 8
--- a/company-template.el
+++ b/company-template.el
@@ -29,16 +29,24 @@
   "Face used for editable text in template fields."
   :group 'company)
 
+(defvar company-template-forward-field-item
+  '(menu-item "" company-template-forward-field
+              :filter company-template--keymap-filter))
+
 (defvar company-template-nav-map
   (let ((keymap (make-sparse-keymap)))
-    (define-key keymap [tab] 'company-template-forward-field)
-    (define-key keymap (kbd "TAB") 'company-template-forward-field)
+    (define-key keymap [tab] company-template-forward-field-item)
+    (define-key keymap (kbd "TAB") company-template-forward-field-item)
     keymap))
+
+(defvar company-template-clear-field-item
+  '(menu-item "" company-template-clear-field
+              :filter company-template--keymap-filter))
 
 (defvar company-template-field-map
   (let ((keymap (make-sparse-keymap)))
     (set-keymap-parent keymap company-template-nav-map)
-    (define-key keymap (kbd "C-d") 'company-template-clear-field)
+    (define-key keymap (kbd "C-d") company-template-clear-field-item)
     keymap))
 
 (defvar-local company-template--buffer-templates nil)
@@ -77,6 +85,10 @@
              (overlay-get ovl 'company-template-after-clear)))
         (when (functionp after-clear-fn)
           (funcall after-clear-fn))))))
+
+(defun company-template--keymap-filter (cmd)
+  (unless (run-hook-with-args-until-success 'yas-keymap-disable-hook)
+    cmd))
 
 (defun company-template--after-clear-c-like-field ()
   "Function that can be called after deleting a field of a c-like template.
--- a/company.el
+++ b/company.el
@@ -774,9 +774,11 @@ keymap during active completions (`company-active-map'):
       (progn
         (add-hook 'pre-command-hook 'company-pre-command nil t)
         (add-hook 'post-command-hook 'company-post-command nil t)
+        (add-hook 'yas-keymap-disable-hook 'company--active-p nil t)
         (mapc 'company-init-backend company-backends))
     (remove-hook 'pre-command-hook 'company-pre-command t)
     (remove-hook 'post-command-hook 'company-post-command t)
+    (remove-hook 'yas-keymap-disable-hook 'company--active-p t)
     (company-cancel)
     (kill-local-variable 'company-point)))
 
@@ -1668,6 +1670,9 @@ prefix match (same case) will be prioritized."
 
 (defsubst company-keep (command)
   (and (symbolp command) (get command 'company-keep)))
+
+(defun company--active-p ()
+  company-candidates)
 
 (defun company-pre-command ()
   (company--electric-restore-window-configuration)


redguardtoo:
Looks th failed check is not caused by my code.
When I checked out previous version and run `make test-batch`, I got same message `SKIPPED  company-non-prefix-fancy-capf-highlighting` w

dgutov:
The failed check was there before, it's all right.

dgutov:
But why is calling tags-completion-table repeatedly a problem? In my testing, it doesn't re-read the tag files unnecessarily.

Do you see the message "Making tags completion table" on subsequent calls?

redguardtoo:
@dgutov , I updated the pull request to fix some error. Compare my latest code with original code:
- `(visit-tags-table-buffer)` will reset cache if tags file `mtime` is changed
- If cache is reset, the whole reloading process is triggered from  function `tags-completion-table`

Now let me explain why this cache update interval is required.

Say I'm editing `fns.c` from Emacs source. Any change of `fns.c` will trigger another process to update tags file. Because tags file `mtime` is updated. The whole cache reloading process is also triggered.

As I tested, cache reloading could take 42 seconds. So when I input one line code and save the file. I need wait 42 second to input second line.

As I know, many developers prefer setting up some backend process to automatically updating tags file. They also prefer manually save the code from frequently. So the use case is typical.

My interval won't give user less candidates because `company-dabbrev-code` in the group is still running.

dgutov:
> Any change of fns.c will trigger another process to update tags file

How does this happen?

Have you tried debouncing this event instead? You would save on CPU time this way.

redguardtoo:
> > Any change of fns.c will trigger another process to update tags file
> 
> How does this happen?
> 
> Have you tried debouncing this event instead? You would save on CPU time this way.

For example, some people installed https://github.com/jixiuf/ctags-update,
```
if ctags-auto-update-mode
      (add-hook 'after-save-hook 'ctags-update nil t)
(remove-hook 'after-save-hook 'ctags-update t)))
```
Actually, I have already tried the debouncing idea. It doesn't work in some projects.

The reason is tags files is also used for code navigation. When navigating code, I expect the tags file reflect the latest code change. So the tags file has to be updated frequently.

dgutov:
> But It doesn't work.

Meaning, you don't like that it make the tags outdated even during navigation?

> Say I update the tags file in every 3 minutes.

How long does that take, BTW?

redguardtoo:
> > But It doesn't work.
> 
> Meaning, you don't like that it makes the tags outdated even during navigation?
I understand `company-etags` is only responsible for code completion.

The key point is reading tags file content into a buffer is always fast. It just takes 2 or 3 second. That buffer is enough for `find-tag` algorithm. No extra data structure need be constructed.

`tags-completion-table` will parse the buffer and build a list of strings. This step is slow. Could take 35~40 seconds. The list of strings is NOT used for code navigation.

I don't think calling `visit-tags-table-buffer` inside `company-etags.el` has anything to do with code navigation.

If you read the code `find-tag-noselect` (use by `find-tag`), it will call `visit-tags-table-buffer` anyway.

As author of `counsel-etag`, I guess algorithm should be similar:
- when user run `whatever-find-tag` command, I load the tags file into a buffer
- search the buffer and find matches
> 
> > Say I update the tags file in every 3 minutes.
> 
> How long does that take, BTW?
It takes 112 seconds to scan 45000 files. The tags file size is about 300M. About 40 seconds to load tags file into Emacs cache.

dgutov:
> That buffer is enough for find-tag algorithm

That's a fair point. `find-tag` doesn't need to build the full completion table.

And if scanning takes 112 seconds, and there's no easy way to speed that up considerably, delaying the scan until you need the up-to-date table (e.g. for `M-x find-tag`) would be too slow for the user.

All right.

I think a better approach is to add that delay feature to `tags-completion-table` in the core. Which means `M-x report-emacs-bug`. After all, this function is used by different completion frontends (we know about `company-etags`, you mentioned `counsel-etags`, and `xref-find-definitions` would be one other example), and it would be better to improve it in one place.

In the meantime, adding an advice on it which would memoize the returned value for xx amount of seconds seems like the best alternative. I wouldn't look up the `tags-completion-table` variable directly, though, that seems like relying on undefined behavior.

redguardtoo:
@dgutov ,
The function `tags-completion-table` set the variable `tags-completion-table`. That variable is not used by `find-tag`, `xref-find-definitions` or `counsel-etags-find-tag`.

The bottleneck is the algorithm of `tags-completion-table` is too slow (could be 40 seconds). So I try to use interval to workaround this bottleneck only.

Any command like `xxx-find-tag` is already fast enough with or without any interval.

`company-etags`:
TAGS => buffer => variable `tags-completion-table`

`find-tag`:
TAGS => buffer => `find-tag` own algorithm (see `find-tag-in-order` which is used by `find-tag-noselect`).

The first step `TAGS => buffer` is always fast. `find-tag` and `company-etags` might use same API in this step. But this step is trivial to the generic performance.

I'm trying to optimize the second step of `company-etags`. It has no effect on `find-tag` or other tag file command.

dgutov:
I get all that. One minor correction: `xref-find-definitions` does use `tags-completion-table` if you call it with a prefix (so it prompts for the identifier).

Now, please read my previous message after "All right".

redguardtoo:
OK, I will try to submit the patch to emacs team at first.

I understand when the cursor is not over any word, `xref-find-definition` will try to call `(tags-completion-table)` to get possible tagname list. I'm not sure Emacs tean will allow me to use interval at this scenario. It's not like `company-mode` which provides other backends to provide missing candidates.

dgutov:
> I'm not sure Emacs tean will allow me to use interval at this scenario.

We'll try to work something out.

> I did grep the emacs source, that's the only scenario tags-completion-table is used inside Emacs

You seem to have missed `tags-lazy-completion-table`, which is used, in turn, in `tags-completion-at-point-function` and `find-tag-tag`.

redguardtoo:
I got a new idea to improve the performance of `company-etags`.

This solution is much better because it solved  the problem from the root cause. Code change is minimum.

`tags-completion-table` need build a lisp list from a big string. But lisp is too damn slow.

So the solution is we don't need destroy and re-create that list at all. We only need update the initial list.

Before refresh the content of tags file, we save current cached content into a temporary file, then use `diff` to get the difference, use regex to extract new tag names from diff output and inserted into the lisp list.

If you think this solution is doable, I can update pull request.

dgutov:
@redguardtoo That's sounds like an intriguing idea (if the changes are small, diff should be as well; as long as the diff program is itself much faster). So I'd happily take a look at the patch and the measurements.

I think there should be a way to plug that feature in without changing `company-etags` itself, though. We can explore that option as the next step.

dgutov:
FWIW, for me it's just a piece of the puzzle. Ideally, the etags program wouldn't re-generate the whole file and do the update faster (because it's retaining the index in its memory). It could also output the changed lines to stdout, so that `diff` is not needed.

redguardtoo:
I tested in Linux kernel project. It's really slow. It's because etags DID read the whole content of a file named `TAGS` and re-generate the tag list from the content.

See `tags-completion-table`, check this line `(push (funcall tags-completion-table-function) tables)`.

`tags-completion-table-function` is `etags-tags-completion-table`,

As you can see, it scans from `(point-mix)` to `(point-max)`,
```lisp
(defun etags-tags-completion-table () ; Doc string?
  (let (table
	(progress-reporter
	 (make-progress-reporter
	  (format "Making tags completion table for %s..." buffer-file-name)
	  (point-min) (point-max))))
    (save-excursion
      (goto-char (point-min))
      ;; This regexp matches an explicit tag name or the place where
      ;; it would start.
      (while (re-search-forward
              "[\f\t\n\r()=,; ]?\177\\\(?:\\([^\n\001]+\\)\001\\)?"
	      nil t)
	(push	(prog1 (if (match-beginning 1)
			   ;; There is an explicit tag name.
			   (buffer-substring (match-beginning 1) (match-end 1))
			 ;; No explicit tag name.  Backtrack a little,
                         ;; and look for the implicit one.
                         (goto-char (match-beginning 0))
                         (skip-chars-backward "^\f\t\n\r()=,; ")
                         (prog1
                             (buffer-substring (point) (match-beginning 0))
                           (goto-char (match-end 0))))
		  (progress-reporter-update progress-reporter (point)))
		table)))
    table))
```

dgutov:
> I tested in Linux kernel project. It's really slow.

I thought that much would be obvious. What is the point you are trying to make?

redguardtoo:
etags program did regenerate the list from the scratch. That's the root cause. So my new suggestion is to make "do the update faster" happen.

Just clarify, I understand that "etags program" means the lisp function `etags-tags-completion-table`. cli program like `etags.el` or `ctags` is not the bottle neck.

dgutov:
Are you saying that the `etags` program invocation finishes quickly even when scanning the Linux kernel?

Anyway, as I said already:

> I'd happily take a look at the patch and the [resulting] measurements

redguardtoo:
Most people use `universal-ctags` or `Exuberant Ctags` these days. But even `etags` is fast enough. The root cause is the Lisp is 100 times slow as C (I did some measurement before). 

I will send you the pull request asap.

Issue #903, opened 2019-07-26
improve the performance `company-etags`


redguardtoo:
Per discussion https://github.com/company-mode/company-mode/pull/877

There is much room for improvement:

1. Use producer/consumer pattern, consumer only searches matches in candidate list. It should know nothing about tags file. The producer is responsible to convert tags file into candidate list (maybe in a different thread if possible)

2. All the tag names are assigned to different partitions, each partition only contains the string with same first character

3. As I mentioned before, we could use CLI program `diff` to get the minimum update patch for the candidate list.

4. More efficient APIs and 3rd party CLI tools should be used. For example, if candidate list is just a plain string containing lines, cli program `sort` could be used to sort lines, which might be faster than sort a Lisp list of strings.

5. build a cache system. Like cache for CPU. Say user inputs "get" to get all the tags starting with "get". It's very possible she still needs the same tags next time.

6. cache the candidates. For example, when user input `test1` to search all tag names start with `test1`, it's very possible candidates start with `tes` is already returned, why not cache the `tes` result in a variable. So search `test1` could re-use `tes` result.

@dgutov , if you are interested, I can send you a new pull request asap.

redguardtoo:
Good news. Looks `(tags-completion-table)`, which is used by `company-etags` right now, is actually not inefficient at all.

Tested with Linux kernel, My new simple string match algorithm take 6 seconds (read the file and build the list) ,while `tags-completion-table` takes 120 seconds.

dgutov:
> is actually not inefficient at all

Not inefficient, or not efficient? If it's the latter, do you have a smaller patch for `tags-completion-table` which can be contributed to the core? I'm sure other users of this function would like to benefit as well.

I'm a bit skeptical, though, considering I've looked into its performance a couple of years ago.

redguardtoo:
Check my pull request, when I started coding, I changed the design a little bit.

Some guy already tried my patched version and the feedback is very positive.

See https://emacs-china.org/t/c-emacs/10068/11 in Chrome, and use google translator to translate the page.

Here is user's translated comment,
![image](https://user-images.githubusercontent.com/184553/62818278-b77ed600-bb88-11e9-9a22-f1c00df31707.png)

dgutov:
> See my pull request

That's not something that can be applied to Emacs, though.

redguardtoo:
only point 2, 3, 5 is used. Other ideas are are too not implemented.

redguardtoo:
[...]

PR #904, opened 2019-07-27
speed up company-etags

redguardtoo:
See #903 

I rewrite the algorithm from the scratch, now the code is much more efficient.

Test with latest linux kernel code.

- The initial loading takes 9 seconds to read tags file and prepare all the candidate. The old algorithm takes more than 120 seconds.

- After initial loading, if CLI program `diff` exists, the new tag name is extracted from diff output. It's less than 0.5 second. The old code need 10~20 seconds at best, or 120 seconds at worst

- `(all-completion prefix all-tags)` is as 50 times fast as old version.

### DIFF
--- a/company-etags.el
+++ b/company-etags.el
@@ -45,6 +45,12 @@ buffer automatically."
   :type 'boolean
   :package-version '(company . "0.7.3"))
 
+(defcustom company-etags-support-ctags-only nil
+  "Nil to support tags file created by both etags and ctags.
+Non-nil to support tags file created only by ctags.
+Please note nil slows down tags file loading time."
+  :type 'boolean)
+
 (defcustom company-etags-everywhere nil
   "Non-nil to offer completions in comments and strings.
 Set it to t or to a list of major modes."
@@ -54,17 +60,42 @@ Set it to t or to a list of major modes."
                          (symbol :tag "Major mode")))
   :package-version '(company . "0.9.0"))
 
+(defcustom company-etags-check-tags-file-interval 30
+  "The interval (seconds) to check tags file.
+Default value is 30 seconds."
+  :type 'integer)
+
+
+(defcustom company-etags-tags-file-name "TAGS"
+  "The name of tags file."
+  :type 'string)
+
 (defvar company-etags-modes '(prog-mode c-mode objc-mode c++-mode java-mode
                               jde-mode pascal-mode perl-mode python-mode))
 
 (defvar-local company-etags-buffer-table 'unknown)
 
+(defvar company-etags-tags-file-caches nil
+  "The cached tags files.")
+
+(defvar company-etags-cached-candidates nil
+  "The cached candidates searched with certain prefix.
+It's like (prefix . candidates).")
+
+(defconst company-etags-fast-pattern
+  "\177\\([^\177\001\n]+\\)\001"
+  "Pattern to extract tag name created by Ctags only.")
+
+(defconst company-etags-slow-pattern
+  "\\([^\f\t\n\r()=,; ]*\\)[\f\t\n\r()=,; ]*\177\\\(?:\\([^\n\001]+\\)\001\\)?"
+  "Pattern to extract tag name created by Ctags/Etags.")
+
 (defun company-etags-find-table ()
   (let ((file (expand-file-name
-               "TAGS"
+               company-etags-tags-file-name
                (locate-dominating-file (or buffer-file-name
                                            default-directory)
-                                       "TAGS"))))
+                                       company-etags-tags-file-name))))
     (when (and file (file-regular-p file))
       (list file))))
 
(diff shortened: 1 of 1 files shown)


redguardtoo:
`make test` passed at my computer. Don't know why CI failed.

cireu:
One test failed on Archlinux, with Emacs 27.0.5 https://github.com/emacs-mirror/emacs/commit/8e0ebb9a3cb9beef2f5ff50436fef1c54a3e3c92

图片

redguardtoo:
> One test failed on Archlinux, with Emacs 27.0.5 [emacs-mirror/emacs@8e0ebb9](https://github.com/emacs-mirror/emacs/commit/8e0ebb9a3cb9beef2f5ff50436fef1c54a3e3c92)
> 
> 图片

This test passed on my computer using Emacs 26.

cireu:
This test also passed on my laptop in Emacs 26.2.

ShuguangSun:
In travis log, the ```emacs --version``` raise error: can't find libguntls.so...

company-etags.el
```diff
@@ -74,15 +89,176 @@ Set it to t or to a list of major modes."
...
+    tagname-dict))
+
+(defun company-etags-append-new-tagname-dict (new-tagnames file-info)
+  "Append NEW-TAGNAMES to FILE-INFO."
+  (dolist (tagname new-tagnames)
+    (company-etags-push-tagname tagname (plist-get file-info :tagname-dict))))
+
+(defun company-etags-all-completions (prefix tagname-dict)
+  "Search for partial matches to PREFIX in TAGNAME-DICT."
+  (let* ((c (elt prefix 0))
```
dgutov: Does this handle of case of prefix being the empty string?

dgutov:
Thank you!

I'm happy to review it. Ideally, though, I'd like to see this improvement split into several steps. If you could provide the performance numbers for every change, that would be fantastic.

redguardtoo:
I just update screencast, start from new Emacs, using linux kernel master. There is timer at right side, so you can see the difference.

Old version:
old

New version:
new

company-etags.el
redguardtoo: fixed

company-etags.el
```diff
@@ -74,15 +89,179 @@ Set it to t or to a list of major modes."
...
+     ((or (and (>= c ?a) (<= c ?z))
+          (and (>= c ?A) (<= c ?Z))
+          (eq c ?$)
+          (eq c ?#)
+          (eq c ?@)
+          (eq c ?%)
+          (eq c ?_)
+          (eq c ?!)
+          (eq c ?*)
+          (eq c ?&)
```
JJPandari: would `memq` slow things down here if used?

redguardtoo: I don't know. Good question.

redguardtoo:
ping @dgutov

redguardtoo:
Hi, @dgutov, what else I need to do to get this pull request merged?

dgutov:
Sorry for the late reply.

To merge this I need to understand these things:

- Why these changes can't go into Emacs core instead, to benefit everyone.

- If they can't, then for each significant piece of complexity, the performance benefits and tradeoffs that come with it.

redguardtoo:
Sure. I can explain.

The reason is simple.  The new algorithm is designed for `company-mode` ONLY. So Emacs core can't use my code.

For example `etags-tags-completion-table` is expected to return a list of strings. But my function returns a complex lisp data structure because of the partition thing.

I could send a new patch to Emacs core team. It's just ten lines of code and is different from my current algorithm.

But in my code, similar function should return

dgutov:
> But my function returns a complex lisp data structure because of the partition thing

So let's tackle this one first. What performance benefits does this bring? It surely can't speed up parsing. It might even slow it down somewhat. But it might speed up completion. So what are the numbers for both?

> The difference is I use string manipulation API

Did we compare the two directly? I imagine the difference might be something else: like the regular expressions used, or the lack of updating the progress bar.

> I avoid parsing function at all when tags file is updated. I use cli program diff. So the update time decreases from 2 minutes to 0.1 second.

Like discussed, it's an interesting idea, but it should fit within Emacs core, if the other blockers don't stop this from happening.

redguardtoo:
> > But my function returns a complex lisp data structure because of the partition thing
> 
> So let's tackle this one first. What performance benefits does this bring? It surely can't speed up parsing. It might even slow it down somewhat. But it might speed up completion. So what are the numbers for both?

The complex data object is to speed up searching completion candidates. The data structure returned has no impact on the performance of parsing function. You didn't ask me to test the "searching completion candidates" performance. So all my tests are testing the the process of extract all tag names from tags file content.

See https://github.com/company-mode/company-mode/issues/903#issuecomment-521234754 

All the measurement is to analyzed the performance of extract the completion list from tags file content. Original time is 2 minutes, not it's about 9~10 seconds.

> 
> > The difference is I use string manipulation API
> 
> Did we compare the two directly? I imagine the difference might be something else: like the regular expressions used, or the lack of updating the progress bar.
> 
Yes, see the link above. There is also another test below the link which compares the different regex. The etags-compatible regex costs about 30 seconds while ctags-compatible regex costs 10 seconds. Using string api only.

The original code in Emacs costs about 2 minutes, using buffer API, etags compatible regex.

I did some measurement on progress bar before. It turns out progress bar is not the key element of performance. If you want to I can show the test code and result.

> > I avoid parsing function at all when tags file is updated. I use cli program diff. So the update time decreases from 2 minutes to 0.1 second.
> 
> Like discussed, it's an interesting idea, but it should fit within Emacs core, if the other blockers don't stop this from happening.
The code is written for company-etags only. It has no effect on Emacs core or any other commands using Emacs core API. What does "fit within Emacs core" mean? What are the other "blockers"? Sorry, I'm not native speaker, so I'm a bit confused.

redguardtoo:
@dgutov, let me explain the original company-etags and what's the bottle neck.

Possible actions:
- a1, read tags file (2s)
- a2, extract all the tag name (2m)
- a3, filter the tag names using prefix (2s)

Scenario a, when company-etags is used for the first time, below actions happen,
- a1
- a2

Scenario b, when user input some prefix, below actions are triggered
- a1 (optional, when tags file is updated)
- a2 (optional, when tags file is updated)
- a3

My optimization:
- I reduce the time of a2 from 2m to 10s in Scenario1
- In Scenario 2, I make a1 and a2 unnecessary by using `diff` (Diff only takes about 1~2 seconds)
- In Scenario 2, I reduce time of a3 from 2s to 0.1s

dgutov:
> The data structure returned has no impact on the performance of parsing function. You didn't ask me to test the "searching completion candidates" performance.

I kind of did. I asked to split this change into steps, each of which added the next piece of implementation complexity, and justify each, with numbers. And also show the trade-offs, ideally.

Regarding partititoning, it must also at least have some downside for completion for empty prefix. Does it work at all? Is it slower by much?

> Original time is 2 minutes, not it's about 9~10 seconds.

That's very good. But that's also achievable without using the partitioning stuff, and thus should be quite compatible with contributing it upstream.

> Yes, see the link above.

It's unexpected because the common wisdom with Emacs is that it's better to use buffers for any operations with large strings, performance-wise. So I'd really like to dig into this comparison later when I have more time. If you post this change to Emacs's bug tracker, though, you could get some reviews from other Emacs core contributors (provided the patch is against progmodes/etags.el).

> The etags-compatible regex costs about 30 seconds while ctags compatible regex costs 10 seconds.

And here, I wonder if it's possible to improve the performance of the former case.

> I did some measurement on progress bar before. It turns out progress bar is not the key element of performance. I can show the test code and result.

That would be welcome, yes.

> What does "fit within Emacs core" mean? What are the other "blockers"? Sorry, I'm not native speaker, so I'm a bit confused.

Let me try to explain.

The current company-etags backend uses a piece of code from core Emacs. Thus delegating any concerns about this code (feature, performence, etc) to Emacs maintainers.

If we decide to reimplement TAGS parsing in this package, we'll have to maintain it later ourselves, which I'm not hugely enthusiastic about. The other downside is that any other callers of tags-completion-table won't benefit from your improvements. So ideally we contribute this kind of improvement upstream, if it's not too hard to integrate. This way we can also get some valuable feedback on the implementation choices.

I have other, less fundamental concerns about the code, but we can discuss those later.

dgutov:
> In Scenaris 2, a1 and a2 is bottle neck initially. After removing these two actions, a3 itself is the bottle neck.

If partitioning is the only way to make completion usable for this large numbers of items, I feel we should bring this up on emacs-devel either way. Stefan might have some insights on that. Or that would inform some better design for future improvement of completion tables.

redguardtoo:
I suggest changing company-etags first. I can send the separate patch to emacs devel team. It's just ten lines of emacs lisp code.

According to my experience, it could take 2 years for company-mode to use latest api from core Emacs (https://lists.gnu.org/archive/html/emacs-devel/2018-04/msg00366.html). It's too long time to wait to avoid 10 lines of duplicacy.

My partition algorithm covers all the corner cases.

Progress bar test result,  (189.05791491 0 0.0) vs (142.075124941 0 0.0)   :
```
(defun my-etags-tags-completion-table () ; Doc string?
  (let (table)
    (save-excursion
      (goto-char (point-min))
      ;; This regexp matches an explicit tag name or the place where
      ;; it would start.
      (while (re-search-forward
              "[\f\t\n\r()=,; ]?\177\\\(?:\\([^\n\001]+\\)\001\\)?"
	      nil t)
	(push	(prog1 (if (match-beginning 1)
			   ;; There is an explicit tag name.
			   (buffer-substring (match-beginning 1) (match-end 1))
			 ;; No explicit tag name.  Backtrack a little,
                         ;; and look for the implicit one.
                         (goto-char (match-beginning 0))
                         (skip-chars-backward "^\f\t\n\r()=,; ")
                         (prog1
                             (buffer-substring (point) (match-beginning 0))
                           (goto-char (match-end 0)))))
		table)))
    table))

(let* ((gc-cons-threshold most-positive-fixnum))
  (message "%S vs %S"
           (benchmark-run-compiled 1
             (setq tags-file-name "~/projs/linux-master/TAGS")
             (visit-tags-table-buffer)
             (etags-tags-completion-table))
           (benchmark-run-compiled 1
             (setq tags-file-name "~/projs/linux-master/TAGS.copy")
             (visit-tags-table-buffer)
             (my-etags-tags-completion-table))))

```

monnier:
Hi,

[ I tried to read the whole exchange of comments, to get up to speed,
  but I may have missed some details.  ]

First, I must say this sounds really cool, and from what I can tell,
I can't see any technical reason why this can't be applied directly
to `etags.el`.

I'm not familiar with `company-etags.el`, so I'm not completely sure what
the patch does, but I'll discuss the parts I saw separately:

1- company-etags-check-tags-file-interval: it seems this isn't just
   speeding things up but introduces a new feature to automatically reload
   data when a TAGS file was changed.  This sounds very useful and would
   be very welcome in `etags.el.

2- ctags vs etags speed: I wasn't aware that there's a large speed
   difference between the two.  Looking at the respective regexps, I can
   see why the ctags regexp is faster.  I think we can significantly
   speed up the etags case in the following way:
   eliminate the subgroup 1 (i.e. use "\177\\\(?:\\([^\n\001]+\\)\001\\)?")
   and use things like (skip-chars-backward "\f\t\n\r()=,; ") to extract
   the tag when the remaining subgroup doesn't match.

3- Use of company-etags-tags-file-name instead of hardcoding "TAGS".
   Sounds like a good idea (and of course fits just fine in etags.el).

4- Regarding "string API" vs "buffer API", I'm wondering why using
   strings is so much faster: when the string contains non-ascii chars,
   operations like (string-match ... start) and substring can take
   a non-trivial amount of time simply finding the corresponding
   position in the string (converting between char positions and byte
   positions), whereas when working in a buffer there's a lot more
   caching going on.  Also working in the buffer avoids the
   `buffer-string` call which copies the whole buffer's contents into
   the string.  So the fact that it's faster for you indicates
   that there are other costs linked (maybe to moving point within
   a buffer) and maybe we can fix those to get something even faster.

5- Regarding the partitioning, it seems perfectly applicable to etags.el
   (and some parts of the code can be simplified, e.g. replacing (or (eq
   ..) (eq ...) ...) with (memq ...), or not initializing the hash-table
   at all).  Also I'm not sure why you need the "rubbish bin".
   And more importantly, it needs a bit more work to handle
   case-insensitive completion.
   I'm also curious what kind of speed up you get from this: Clearly,
   I can imagine cases where you can get a significant speed up (for
   very large tags tables and non-empty pefix) but depending on the
   completion style you use, `prefix` may almost always be "", so we
   still want that case to be "fast enough" (if possible) so it'd good
   to know if it can't be "fast enough".

Any chance you could help us port your patch to etags.el (I'll let
Dmitry figure out how to deal with using the faster code in
company-etags.el when running on older Emacsen)?

redguardtoo:
@monnier ,.

4. string api has less overhead than buffer api. In the algorithm I designed for company-mode, it's only called when parsing tags file content for once. So the initial speed is important. When tags files content is changed, I use `diff`program to extract new tag names (please note I don't delete obsolete tag names). This is also optimization for company-etags only, company-etags only completes function name.

I didn't use `memq` because it's not fast enough. `rubish bin` because some tag name's first character might not fall back into any parititoin. It's just sanity check code.

monnier:
> 4. string api has less overhead than buffer api.

You said so a few times already, so I believe it is your experience, but
I'm quite familiar with the implementation, and I know that the reality
is more complex than that.  I can create cases where this is true and
case where the reverse is true as well: it's not "more" or "less"
overhead, it's just a different overhead.

For the problem at hand, I'm surprised you found a noticeable difference
(I'd expect the use of a buffer to be (if anything) slightly more
efficient rather than less), so I think it would be very beneficial to
try and figure out where the slow behavior come from when we use the buffer.

> In the algorithm I designed for company-mode, it's only called when
> parsing tags file content for once.

Yes, I know.

> So the initial speed is important. When tags files content is changed,
> I use `diff`program to extract new tag names (please note I don't delete
> obsolete tag names).

Yes, I also saw that.

> This is also optimization for company-etags only,
> company-etags only completes function name.

The same optimization would apply to all uses of etags if applied to
etags.el (e.g. for xref), which would be much better.

> I didn't use `memq` because it's not fast enough.

(memq x '(...)) is faster than a bunch of (or (eq ...) (eq ...) ...)

> `rubish bin` because some tag name's first character might not fall
> back into any parititoin.

You don't need to "create" any of the partitions, so you can start as if
you had "all characters" already initialized, so you don't need any
"rubbish bin": (push x (gethash c ht)) will work just fine even if
there's no existing entry for `c` since (gethash c ht) will return nil
in that case.


        Stefan

PR #930, opened 2019-11-08
Use line-wrap in company doc buffer

Zetagon:
I had issues where long lines did not wrap in company-doc-buffer. I think this fixes it but this is my first time looking at company-modes code so feel free to criticize this all you want.

### DIFF
--- a/company.el
+++ b/company.el
@@ -2271,7 +2271,8 @@ character, stripping the modifiers.  That character must be a digit."
     (erase-buffer)
     (when string
       (save-excursion
-        (insert string)))
+        (insert string)
+        (visual-line-mode)))
     (current-buffer)))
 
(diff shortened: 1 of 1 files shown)


dgutov:
It might be a good change. I wonder, though, which circumstances is this required in? Some particular backend and/or some window width?

Zetagon:
Oh yeah. I am using doom-emacs which puts all "popup buffers" at the bottom(see the screenshot) and company-lsp for java.

This is with line-wrapping on.
image

Will there be a performance hit? I don't know how often this is an issue but I definitely would not want sentences to run of screen when reading docs.

dgutov:
OK, let's do this. It shouldn't hurt, performance or otherwise.

Issue #889, opened 2019-04-27
Any way to show numbers on the left?


AmaiKinono:
Having `M-<digit>` to select a candidate is nice, but when one of the candidates are too long, it's hard to find out which one corresponds to which number:

![2019-04-28-014054_693x190_scrot](https://user-images.githubusercontent.com/28714352/56853200-11032200-6957-11e9-8fb3-4b9e503c2a94.png)

But having them on the left would make it so convenient.

I looked into the code, and the position seems to be hardcoded? Could we have an option to decide which side should the numbers go?

dgutov:
There is one now!

PR #941, opened 2019-12-28
add option to show number on the left

amosbird:
https://github.com/company-mode/company-mode/issues/889

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -13,6 +13,7 @@
   map binding `TAB` to "jump to the next field"). Affects both
   `company-template` and `yasnippet` (requires changes from 2019-04-21,
   currently unreleased).
+* New values for option `company-show-numbers` to show number on the left.
 
 ## 2019-04-15 (0.9.10)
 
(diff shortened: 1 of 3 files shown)


dgutov:
Hi! Thanks for the patch.

Two notes:

I think we can get by with just one new value of `company-show-numbers` instead of a new variable (the value would be symbol `left`).

This feature needs at least one test.

amosbird:
ok, updated

NEWS.md
```diff
@@ -13,6 +13,7 @@
   map binding `TAB` to "jump to the next field"). Affects both
   `company-template` and `yasnippet` (requires changes from 2019-04-21,
   currently unreleased).
+* New values for option `company-show-numbers` to show number on the left.
```
dgutov: "New value"

company.el
```diff
@@ -641,7 +641,9 @@ commands in the `company-' namespace, abort completion."
 (defcustom company-show-numbers nil
   "If enabled, show quick-access numbers for the first ten candidates."
   :type '(choice (const :tag "off" nil)
-                 (const :tag "on" t)))
+                 (const :tag "left" 'left)
+                 (const :tag "right" 'right)
```
dgutov: Why `right`? Just use `t` for this.

amosbird: It's for some user who imagines symmetric configurations... I'll remove it

dgutov:
OK, let's try it. Thank you!

PR #946, opened 2020-01-27
Stop overriding post-completion

### LINKED ISSUE #881: Enable both company-yasnippet and company-tng and no expansion on snippet
Hi, there,
I've enabled add `company-yasnippet` to `company-lsp` and with `company-backends` as following and also enable the `company-tng`:
```
((company-lsp . #1=(:with company-yasnippet))                                                                                                                                                               
 (company-bbdb . #1#)                                                                                                                                                                                       
 (company-eclim . #1#)                                                                                                                                                                                      
 (company-semantic . #1#)                                                                                                                                                                                   
 (company-clang . #1#)                                                                                                                                                                                      
 (company-xcode . #1#)                                                                                                                                                                                      
 (company-cmake . #1#)                                                                                                                                                                                      
 (company-capf . #1#)                                                                                                                                                                                       
 (company-files . #1#)                                                                                                                                                                                      
 (company-dabbrev-code company-gtags company-etags company-keywords . #1#)                                                                                                                                  
 (company-oddmuse . #1#)                                                                                                                                                                                    
 (company-dabbrev . #1#)) 
```


```
(setq company-idle-delay 0)
(setq company-minimum-prefix-length 1)
(setq company-selection-wrap-around t)
(company-tng-configure-default)
```

And when I type `for` in python-mode, I get following completation:

<img width="278" alt="屏幕快照 2019-03-18 下午10 38 16" src="https://user-images.githubusercontent.com/1575247/54538669-9cd87480-49cf-11e9-9a96-b99585003ebd.png">

But candidates 0 (`for (Keyword)`) and 1 `(for -> for ... in ... : ...)` give the same result (no snippet expansion on candidate 1 at all).

What shall I do to make the candidate 1 to expand to a correct snippet?

**Thanks for your discussion! For my current solution is:**

1. enable `tng` post-completion by:
```
(advice-add 'company-tng--supress-post-completion :override #'ignore)
```
2. bind yas's exand to other key instead of `tab`
```
604(define-key yas-minor-mode-map "\C-j" 'yas-expand)                                                                                                                                                                                                                         
605(define-key yas-keymap "\C-j" 'yas-next-field-or-maybe-expand)                                                                                                                                                                                                             
606(dolist (keymap (list yas-minor-mode-map yas-keymap))                                                                                                                                                                                                                      
607  (define-key keymap (kbd "TAB") nil)                                                                                                                                                                                                                                      
608  (define-key keymap [(tab)] nil))
```

### LINKED ISSUE #943: company-tng completes whole argument definition on completion with lsp-mode
I've noticed a strange behavior specific to using company and lsp (specifically in C++) where company inserts any arguments declared for a function exactly, with all commas, names and types are inserted into the buffer and adds a space after a full stop.
image
image
Now, this isn't very useful to me because clearly I have to go back and edit, well, delete the parameter list if I do not want an error. I believe this likely has some fault in lsp-mode but I wondered what could be done on company side, all I want is in situations like this is a little workaround that doesn't have the extra space and stops at the end of the function name so I can fill in the function manually. Especially since normal company is fine and stops completion at the first bracket.
This is all the company configuration I have 
image

dgutov:
Disable args insertion in individual backends instead.

Fixes #943.
Fixes #881.

### DIFF
--- a/company-tng.el
+++ b/company-tng.el
@@ -1,6 +1,6 @@
 ;;; company-tng.el --- company-mode configuration for single-button interaction
 
-;; Copyright (C) 2017  Free Software Foundation, Inc.
+;; Copyright (C) 2017-2019  Free Software Foundation, Inc.
 
 ;; Author: Nikita Leshenko
 
@@ -64,15 +64,19 @@
 ;; We recommend to disable `company-require-match' to allow free typing at any
 ;; point.
 ;;
-;; By default, company-tng doesn't work well with backends that use
-;; `post-completion' (for actions such as expanding snippets in
-;; company-yasnippet or company-template). In company-tng, completion candidates
+;; By default, company-tng doesn't work well with backends that insert function
+;; arguments into the buffer and (optionally) expand them into a snippet
+;; (usually performed in `post-completion' using yasnippet or company-template).
+;; In company-tng, completion candidates
 ;; are inserted into the buffer as the user selects them and the completion is
 ;; finished implicitly when the user continues typing after selecting a
 ;; candidate. Modifying the buffer (by expanding a snippet) when the user
 ;; continues typing would be surprising and undesirable, since the candidate was
-;; already inserted into the buffer. For this reason company-tng disables
-;; `post-completion' in all backends.
+;; already inserted into the buffer.
+;;
+;; For this reason `company-tng-configure-default' disables arguments insertion
+;; for a number of popular backends.  If the backend you are using is not among
+;; them, you might have to configure it not to do that yourself.
 ;;
 ;; YASnippet and company-tng both use TAB, which causes conflicts. The
 ;; recommended way to use YASnippet with company-tng is to choose a different
@@ -122,8 +126,12 @@ confirm the selection and finish the completion."
      (when (and company-selection-changed
                 (not (company--company-command-p (this-command-keys))))
        (company--unread-this-command-keys)
-       (setq this-command 'company-complete-selection)
-       (advice-add 'company-call-backend :before-until 'company-tng--supress-post-completion)))))
+       (setq this-command 'company-complete-selection)))))
+
+(defvar company-clang-insert-arguments)
+(defvar company-semantic-insert-arguments)
+(defvar company-rtags-insert-arguments)
+(defvar lsp-enable-snippet)
 
 ;;;###autoload
 (defun company-tng-configure-default ()
@@ -132,6 +140,11 @@ confirm the selection and finish the completion."
   (setq company-frontends '(company-tng-frontend
                             company-pseudo-tooltip-frontend
                             company-echo-metadata-frontend))
+  (setq company-clang-insert-arguments nil
+        company-semantic-insert-arguments nil
+        company-rtags-insert-arguments nil
+        lsp-enable-snippet nil)
+  (advice-add #'eglot--snippet-expansion-fn :override #'ignore)
   (let ((keymap company-active-map))
     (define-key keymap [return] nil)
     (define-key keymap (kbd "RET") nil)
@@ -179,16 +192,6 @@ made explicitly (i.e. `company-selection-changed' is true)"
     ;; The 4th arg of `company-fill-propertize' is selected
     (setf (nth 3 args) nil))
   args)
-
-(defun company-tng--supress-post-completion (command &rest args)
-  "Installed as a :before-until advice on `company-call-backend' and
-prevents the 'post-completion command from being delivered to the backend
-for the next iteration. post-completion do things like expand snippets
-which are undesirable because completions are implicit in company-tng and
-visible side-effects after the completion are surprising."
-  (when (eq command 'post-completion)
-    (advice-remove 'company-call-backend 'company-tng--supress-post-completion)
-    t))
 
 (provide 'company-tng)
 ;;; company-tng.el ends here
--- a/company-tng.el
+++ b/company-tng.el
@@ -1,6 +1,6 @@
 ;;; company-tng.el --- company-mode configuration for single-button interaction
 
-;; Copyright (C) 2017-2019  Free Software Foundation, Inc.
+;; Copyright (C) 2017-2020  Free Software Foundation, Inc.
 
 ;; Author: Nikita Leshenko
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,15 @@
 # History of user-visible changes
 
+## Next
+
+* `company-tng` stops disabling `post-completion` in backends
+  ([#946](https://github.com/company-mode/company-mode/pull/946)). Instead,
+  `company-tng-configure-default` disables snippet expansion in most popular
+  backends. If a backend you use needs this and is not covered, and you use
+  `company-tng`, disable snippet insertion by customizing a relevant option
+  provided by the backend. The result is better compatibility with LSP backends
+  because they currently depend on `post-completion` in all cases.
+
 ## 2020-02-07 (0.9.12)
 
 * Tooltip rendering bugfix.


company-tng.el
```diff
@@ -132,6 +140,11 @@ confirm the selection and finish the completion."
   (setq company-frontends '(company-tng-frontend
                             company-pseudo-tooltip-frontend
                             company-echo-metadata-frontend))
+  (setq company-clang-insert-arguments nil
+        company-semantic-insert-arguments nil
+        company-rtags-insert-arguments nil
+        lsp-enable-snippet nil)
```
yyoncho: `lsp-enable-snippet` won't have any effect since the snippets are already enabled on server startup.

Mallchad:
This works as I wanted, thanks guys. 
@dgutov 
@yyoncho 
As mentioned in issue #943 it's not perfect since the inline preview does not match up with what is actually inserted exactly, one character to the right, with argument list.
image
but that's ok for me personally, it's a workable solution until everything else is fixed.
Additional information and notes is mentioned in issue #943

company-tng.el
dgutov: Wouldn't it? This function is supposed to be called from the user's init script, so a restart should fix the problem, at least.

dgutov:
> If we enable post-completion: there is no point to disable snippets unless we force lsp-mode to insert the target text (doing what YCM was doing)

If I understood Mallchad's reaction right, the crucial difference is there: when the completion is finally inserted, it's inserted without the arguments list (thanks to the textedit in post-completion). The inline preview shows something different, though (it's also a problem, though not as serious).

Of course, if you manage to make the backend work closer to how most others work (e.g. company-clang), that should fix both problems. Possibly making this PR unnecessary (I'm not sure).

>  There is no key to just insert the item at the focus(e. g. pressing ret inserts new line).

To "just insert the item at the focus" (after pressing TAB however many times to select it), continue typing as if the completion is already inserted.

> This does not feel natural if the inserted text is not the target text.

Could you rephrase that?

>  When company popup shows up there is nothing selected/inserted and pressing ret just inserts ret.

That's the idea behind this backend: only TAB performs completion or selects among the completions (and you have to press it at least once). RET just inserts RET.

yyoncho:
> Of course, if you manage to make the backend work closer to how most others work (e.g. company-clang), that should fix both problems. Possibly making this PR unnecessary (I'm not sure).

IMO it will add one more option - use company-tng with post-completion(in any form). 

> To "just insert the item at the focus" (after pressing TAB however many times to select it), continue typing as if the completion is already inserted.

I was referring the case when you want to insert without typing anything but something like `C-f` also inserts so this case is handled too.
 
> Could you rephrase that?

If you have `foo(String param, String param2)` as a preview and then when you start typing it changes to `foo(|,);` which feels unnatural to me. You may disregard this comment since it is purely subjective and we will be solved when we change lsp-mode.

dgutov:
> IMO it will add one more option - use company-tng with post-completion(in any form)

Seems like a good thing, yes.

> I was referring the case when you want to insert without typing anything but something like C-f also inserts so this case is handled too.

Right.

> If you have foo(String param, String param2) as a preview and then when you start typing it changes to foo(|,); which feels unnatural to me. You may disregard this comment since it is purely subjective and we will be solved when we change lsp-mode.

I agree it's a problem. We've already mentioned it, I think.

Yet it's probably better behavior than the current state of affairs. And indeed, if lsp-mode fixes it to behave closer to other backends (with argslist being a part of annotation), it would be better in several respects. About that, though... it seems like you have some other strings serve as annotations already (return types, I think). Would it work well to just concatenate them?

ratking82:
Hi, I have the exact issue that has been addressed in this PR.  I was wondering if this will be merged upstream or should I just patch my copy of company-tng.el ? Thank you!

dgutov:
@ratking82 Hi!

Have you tried this patch? Do you like the new behavior?

dgutov:
And which backends have you used it with?

ratking82:
> And which backends have you used it with?

@dgutov  I used it mainly with company-lsp with clangd and pyls as lsp servers.  With the `lsp-mode` config variable `lsp-enable-snippet` set to `nil`, in both `cc-mode` and `python-mode`, the result of completion was correct. Strangely, `RET` does not insert and exit completion, putting point after the completion text, but additionally enters a newline. In `cc-mode` while visualising the completions have a whitespace. From the clangd verbose output it is apparent that the `label` field of the completion LSP message has a whitespace. I donot know exactly why this is so. Note, the whitespace is not present in completions when using clangd as lsp server for VSCode.  

Setting the variable to `t` results, in python mode, the insertion of functions arguments or placeholder like `$0`.  

After your message I also tried using company-clang and company-capf; none of which were able to provide a completion. Though this most probably is an issue arising out of my current build configuration and is not necessarily an issue with company mode.

company-tng.el
nikital: Maybe you should also disable `company-lsp-enable-snippet`? This way even if LSP sends a snippet, `company-lsp` won't insert it.

nikital:
@ratking82 Could you provide me your configuration? I'd like to look into the trailing whitespace issue you've mentioned and I don't personally use LSP.

Regarding `RET`, you don't need to type `RET` to finish the completion in `company-tng`. The idea in `comany-tng` is that you select the completion candidate and just continue typing as if the candidate has already been inserted. So if you press `RET` after selecting a candidate the expected behaviour is that a newline is inserted.

company-tng.el
nikital, line 146: Maybe you should also disable company-lsp-enable-snippet? This way even if LSP sends a snippet, company-lsp won't insert it.

ratking82:
> @ratking82 Could you provide me your configuration? I'd like to look into the trailing whitespace issue you've mentioned and I don't personally use LSP.

@nikital  Please note, that the whitespace is prepended to the `label:` field. I far as I can tell, this is the response from clangd and is not an issue with company. As par your request, please find below my personal configuration for `lsp-mode` and `company`.  FYI, I am using version 10 of clangd. 

```
(use-package company
  :config
  (setq company-minimum-prefix-length 1
        company-idle-delay 0.0
        company-require-match nil
        company-tooltip-align-annotations t
        company-dabbrev-downcase case-replace
        company-show-numbers t
        company-transformers '(company-sort-by-occurrence))
  (global-company-mode))

(use-package lsp-mode
  :init (setq lsp-keymap-prefix "s-j")
  :commands (lsp lsp-deferred)
  :config (setq lsp-auto-configure t
                lsp-auto-guess-root nil
                lsp-log-io t
                ;; Do not use flymake
                lsp-prefer-flymake nil
                lsp-enable-xref t
                lsp-enable-on-type-formatting t
                lsp-imenu-show-container-name t
                lsp-enable-snippet nil
                lsp-prefer-capf nil)
  (setq lsp-clients-clangd-args '("-j=6" "--background-index"
                                  "--log=verbose"
                                  "--header-insertion=iwyu"
                                  "--header-insertion-decorators"
                                  "--all-scopes-completion"
                                  "--clang-tidy"
                                  "--pretty"))
  :hook
  (c++-mode . lsp-deferred)
  (python-mode . lsp-deferred)
  (lsp-mode . lsp-enable-which-key-integration
            ))

(use-package company-lsp
  :after (lsp))
```

> 
> Regarding `RET`, you don't need to type `RET` to finish the completion in `company-tng`. The idea in `comany-tng` is that you select the completion candidate and just continue typing as if the candidate has already been inserted. So if you press `RET` after selecting a candidate the expected behaviour is that a newline is inserted.

Thank you for the clarification.

company-tng.el
dgutov: According to Ivan, this variable is no longer used. But if you notice a problem that is fixed with setting it to nil, please let me know.

dgutov:
All right, looks like this is probably a good idea. Let's run with it.

@nikital Please let us know if the investigation leads to something interesting.

nikital:
@ratking82 I tried to reproduce the whitespace issue using your config, `clangd` 10, clean Emacs 26.3 and the latest packages. I couldn't trigger a spurious whitespace :(
Could you give me a concrete text example that triggers the issue?

PR #965, opened 2020-03-15
Add the docbuffer in company-yasnippet

taigacute:
Like this,

image

### DIFF
--- a/company-yasnippet.el
+++ b/company-yasnippet.el
@@ -100,7 +100,8 @@
 (defun company-yasnippet--doc (arg)
   (let ((template (get-text-property 0 'yas-template arg)))
     (with-current-buffer (company-doc-buffer)
-      (insert (yas--template-content template))
+      (yas-minor-mode 1)
+      (yas-expand-snippet (yas--template-content template))
       (goto-char (point-min))
       (current-buffer))))
 
--- a/company-yasnippet.el
+++ b/company-yasnippet.el
@@ -99,17 +99,17 @@
 
 (defun company-yasnippet--doc (arg)
   (let ((template (get-text-property 0 'yas-template arg))
-        (mode (format "%s" major-mode)))
+        (mode (symbol-value 'major-mode)))
     (with-current-buffer (company-doc-buffer)
       (yas-minor-mode 1)
       (yas-expand-snippet (yas--template-content template))
       (delay-mode-hooks
         (let ((inhibit-message t))
-          (if (string-equal mode "web-mode")
+          (if (string-equal mode 'web-mode)
                 (progn
-                  (setq mode "html-mode")
-                  (funcall (intern mode)))
-            (funcall (intern mode))))
+                  (setq mode 'html-mode)
+                  (funcall mode))
+            (funcall mode)))
          (ignore-errors (font-lock-ensure)))
       (current-buffer))))
 
--- a/company-yasnippet.el
+++ b/company-yasnippet.el
@@ -97,6 +97,13 @@
        res))
    tables))
 
+(defun company-yasnippet--doc (arg)
+  (let ((template (get-text-property 0 'yas-template arg)))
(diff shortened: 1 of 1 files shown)


company-yasnippet.el
```diff
@@ -97,6 +97,22 @@
        res))
    tables))
 
+(defun company-yasnippet--doc (arg)
+  (let ((template (get-text-property 0 'yas-template arg))
+        (mode (format "%s" major-mode)))
```
dgutov: Why not keep `mode` a symbol value? Then, below you will compare it with `'web-mode` and set to `'html-mode` if true.

taigacute: thanks.I'm not very good at writing elisp.

dgutov: How about:

```
(mode major-mode)
```

company-yasnippet.el
```diff
@@ -97,6 +97,22 @@
...
 
+(defun company-yasnippet--doc (arg)
+  (let ((template (get-text-property 0 'yas-template arg))
+        (mode (symbol-value 'major-mode)))
+    (with-current-buffer (company-doc-buffer)
+      (yas-minor-mode 1)
+      (yas-expand-snippet (yas--template-content template))
+      (delay-mode-hooks
+        (let ((inhibit-message t))
+          (if (string-equal mode 'web-mode)
```
dgutov: Just `(eq` would suffice.

taigacute: ok  i rewrite it .

company-yasnippet.el
taigacute: got it! thanks.

taigacute:
A tip for this pr,use these code  wroks lsp ,set  company-backends `(company-capf company-yasnippet)`, Will get a nice complementary experience。
```elisp
  (defun company-yasnippet/disable-after-dot (fun command &optional arg &rest _ignore)
    (if (eq command 'prefix)
        (when-let ((prefix (funcall fun 'prefix)))
          (unless (memq (char-before (- (point) (length prefix))) '(?. ?> ?\())
            prefix))
      (progn
        (when (and arg (not (get-text-property 0 'yas-annotation-patch arg)))
          (let* ((name (get-text-property 0 'yas-annotation arg))
                 (snip (format "%s (Snippet)" name))
                 (len (length arg)))
            (put-text-property 0 len 'yas-annotation snip arg)
            (put-text-property 0 len 'yas-annotation-patch t arg)))
        (funcall fun command arg))))

  (advice-add #'company-yasnippet :around #'company-yasnippet/disable-after-dot)
```

dgutov:
Thanks!

dgutov:
I have a couple of questions regarding your snippet:

* Why `?\(`? Are snippets after an opening paren unlikely to help either?
* Is the `yas-annotation-patch` thing there to clarify the annotation? Is the ` -> ` arrow not enough?

dgutov:
Try this, maybe:

```diff
diff --git a/company-yasnippet.el b/company-yasnippet.el
index e5fded4..f773522 100644
--- a/company-yasnippet.el
+++ b/company-yasnippet.el
@@ -131,7 +131,7 @@ shadow backends that come after it.  Recommended usages:
           (company-grab-symbol)))
     (annotation
      (concat
-      (unless company-tooltip-align-annotations " -> ")
+      " -> "
       (get-text-property 0 'yas-annotation arg)))
     (candidates (company-yasnippet--candidates arg))
     (no-cache t)
```

taigacute:
> Why ?\(? Are snippets after an opening paren unlikely to help either?  
  
..seems like is wrong, maybe it should be removed. I modified it.

> Is the yas-annotation-patch thing there to clarify the annotation? Is the -> arrow not enough?  

config it will look better when use lsp.
image

dgutov:
OK.

I create a var: `company-yasnippet-annotation-fn`. You can change this behavior more easily through it.

PR #968, opened 2020-03-24
Keywords complete for c++-mode and c-mode

condy0919:


dgutov:
Thanks!

Could you make sure the keywords are sorted alphabetically? The build is failing otherwise: https://travis-ci.org/github/company-mode/company-mode/pull_requests

condy0919:
`char16_t` should be place before `char8_t` although the `sizeof(char8_t)` is less than `sizeof(char16_t)`...

PR #971, opened 2020-03-29
create company-extract-common function

redguardtoo:
Extract code into a new function `company-extract-common`. Logic is still same.

`company-extract-common` could be changed int the future to support fuzzy matching of candidates. Current logic assume matching with prefix.

### DIFF
--- a/company.el
+++ b/company.el
@@ -1195,6 +1195,13 @@ can retrieve meta-data for them."
                                             (symbol-name backend))))
         (format "%s-<%s>" base name)))))
 
+(defun company-extract-common (prefix candidates)
+  "Use PREFIX to extract common string of CANDIDATES."
+  (let ((common (try-completion "" candidates)))
+    (when (string-prefix-p prefix common
+                           completion-ignore-case)
+      common)))
+
 (defun company-update-candidates (candidates)
   (setq company-candidates-length (length candidates))
   (if company-selection-changed
@@ -1223,10 +1230,7 @@ can retrieve meta-data for them."
     ;; `company-complete-common', unless there's only one candidate.
     (setq company-common
           (if (cdr company-candidates)
-              (let ((common (try-completion "" company-candidates)))
-                (when (string-prefix-p company-prefix common
-                                       completion-ignore-case)
-                  common))
+              (company-extract-common company-prefix company-candidates)
             (car company-candidates)))))
 
 (defun company-calculate-candidates (prefix ignore-case)


dgutov:
How would support for fuzzy matching work based on this?

The common part in that case is essentially an array of strings, not really a string.

redguardtoo:
Say there are two tags "helloHell" and "helloWold". The user inputs "hlo" which matches "hello" in fuzzy matching algorithm, since both tags has common "hello", so when user press TAB, the word "hello" should be completed.

One example of the the fuzzy algorithm is convert "hlo" to regex "h.*l.*o.*" to match candidates.

The company-common is a string which is set by `company-extract-common`. As I checked, common is string type,

image

dgutov:
I know what fuzzy matching is.

How would you propose changing the resulting function to support it? Try, see if it works well.

redguardtoo:
Demo:
company-fuzzy

To test,
- use my branch code
- create `company-test.el`
```lisp
(defadvice company-extract-common (around company-extract-common-hack activate)
  (let* ((args (ad-get-args 0))
         (prefix (nth 0 args))
         (candidates (nth 1 args))
         (common (try-completion "" candidates)))
    (when (string-match (company-test-fuzzy-pattern prefix) common)
      (setq ad-return-value common))))

(defun company-test-fuzzy-pattern (prefix)
  (concat "^" (seq-reduce (lambda (r e) (concat r (char-to-string e) ".*")) prefix "")))

(defun company-test-candidates (prefix)
  (let* ((all-items '("helloWorld1"
                      "helloWorld2"
                      "helloHell1"
                      "helloHell2"
                      "byeWorld1"
                      "byeWorld2"
                      "byeHell1"
                      "byeHell2"))
         rlt)
    (dolist (item all-items)
      (when (string-match (company-test-fuzzy-pattern prefix) item)
      ;; (when (string-match (concat "^" prefix)  item)
        (push item rlt)))
    rlt))

(defun company-test (command &optional arg &rest ignored)
  (interactive (list 'interactive))
  (cl-case command
    (interactive (company-begin-backend 'company-test))
    (init t)
    (prefix (company-grab-symbol))
    (candidates (company-test-candidates arg))))
```
- Run `(setq company-backends '(company-test))` in any file
- Input `hll`, then press TAB

dgutov:
Okay. But try this:

* Input `h`, then press TAB.
* Try to input `ll` after that.
* See `Matching input in required` in the echo area instead.

redguardtoo:
Hi, @dgutov ,
The user need input at least three character to trigger auto-complete. That's controlled by `company-minimum-prefix-length` in `company.el`, If I `(setq company-minimum-prefix-length 1)` ,input `h`, and press TAB, everything works fine. 

`ll` is supposed not to match any candidate, as defined in function `company-test-fuzzy-pattern`

dgutov:
No, the full input in that case will be `hll`. And it *should* match.

`company-minimum-prefix-length` does not affect the behavior when completion is triggered with a command explicitly.

My point is, your scenario will work even without the advice. My scenario *should* be fixed by that advice (in theory), but isn't.

Make sure you have the default value of `company-require-match` set before trying it.

Issue #380, opened 2015-07-22
Dark themes in terminal


romankrv:
Need redesign color in dark themes in terminal. 
It's look like this. See attach
![screen shot 2015-07-22 at 1 01 40 pm](https://cloud.githubusercontent.com/assets/42733/8839775/bf991bf6-30e1-11e5-83c9-c63aa6207f08.png)

dgutov:
I suppose; patches welcome.

But does Emacs even know that it's running in a dark-background terminal? The popup looks much better on a light background.

romankrv:
It's problem on iTerm. It's probably conflict 256 color of the palette of iTerm and company.el color pallete.

dgutov:
> 256 color

Right.

But what I mean is, you might not be able to just change the colors to look better on dark background. What does `frame-background-mode` evaluate to, in the above environment?

romankrv:
Can you get me more detail or workflow how to deal with the  “frame-background-mode“ 
M-x frame-background-mode not much in my emacs

dgutov:
`M-: frame-background-mode RET`

romankrv:
not exist, can give me elisp code for that?

dgutov:
What does not exist? This variable has been in Emacs for years.

What kind of code are you looking for?

romankrv:
I try to find function as you said: frame-background-mode
but it only has - frame-set-background-mode

![screen shot 2015-07-24 at 5 45 32 pm](https://cloud.githubusercontent.com/assets/42733/8877075/3f978996-322c-11e5-8b85-29d737243d2c.png)

How to get frame-background-mode ?
Thanks

Silex:
It's a variable.

romankrv:
Yes, you are right, it exist. 
This variable has value 'nil' I checked it in iTerm(256 color app) and Term.app (16 color app) and both of them are equal 'nul'
That is it variable not use in.

Tell me please, Is your project (company-mode) not ready to work with 256 color palette?

Let's start to plan to change it. You agree? Because you are more the expert in your project then I. 

Thanks.

dgutov:
@romankrv Does anything change, visually, if you customize that variable to `dark`?

romankrv:
okay, I try it 
and report then

ned2:
[...]

PR #986, opened 2020-05-24
Set tooltip background color for terminals with less than 88 colors

hanazuki:
Related issue: #380 -- The screenshot on the first comment describes the issue fixed by this patch.

The current implementation specifies the foreground color for tooltips to be black but does not gives the background color on terminals with less color capability, which causes invisible text when the background is black on <88-color terminals.

This patch fixes the problem by giving a default background color. This change does not affect >=88-color terminals.

I believe many people use 256-color terminals or true-color GUI nowadays, but I use an 8-color terminal and want every Emacs package to have a usable default color on it...

### DIFF
--- a/company.el
+++ b/company.el
@@ -90,7 +90,7 @@ attention to case differences."
   :group 'faces)
 
 (defface company-tooltip
-  '((default :foreground "black")
+  '((default :foreground "black" :background "yellow")
     (((class color) (min-colors 88) (background light))
      (:background "cornsilk"))
     (((class color) (min-colors 88) (background dark))


dgutov:
Hi!

Could you show a resulting screenshot?

hanazuki:
yes!

## master (dark / light)
Screen Shot 2020-05-24 at 22 17 31
Screen Shot 2020-05-24 at 22 22 22


## master+patch (dark / light)
Screen Shot 2020-05-24 at 22 19 24
Screen Shot 2020-05-24 at 22 21 20

dgutov:
Thanks!

hanazuki:
Thank you!

isolet:
IMO, it's a bad default for GUI emacs.

dgutov:
@isolet See #987.

Issue #959, opened 2020-03-04
How to contribute a new backend?


c-alpha:
Hi, I've requested (and received) the copyright assignment form via the `emacs-devel` mailing list. I'll mention `company-mode` as well, and sign the assignment agreement once I've received it.

What next after that?

Make a PR here? Or propose it in an email to the `emacs-devel` mailing list (it's a small file, so could well attach it to the mail)?

Any hints appreciated.

dgutov:
Hi!

Either way is fine.

However, if it's a niche backend (not one that would the majority of company-mode users would use), maybe the best way to publish it is a separate package, going through MELPA. Or GNU ELPA.

What kind of backend is it?

> I'll mention company-mode as well, and sign the assignment agreement once I've received it.

Putting 'emacs' as the program name for the assignment should suffice.

c-alpha:
> [...]
> However, if it's a niche backend (not one that would the majority of company-mode users would use), maybe the best way to publish it is a separate package, going through MELPA. Or GNU ELPA.
> 
> What kind of backend is it?
> [...]

It is for [EUDC](https://www.gnu.org/software/emacs/manual/html_mono/eudc.html), he Emacs Unified Directory Client, a common interface to directory servers and contact information. This will allow email addresses coming from LDAP servers to be offered for completion in relevant headers in message buffers.

dgutov:
Sounds a bit niche to me.

But other than that, wouldn't you prefer to have this completion also supported by `M-x completion-at-point`? If Emacs has some EUDC support already, this might fit better as a patch for `message.el`. See `message-completion-alist` and so on.

To go this route, you might want to ask further questions on emacs-devel. Alternatively, to publish this backend as-is as a separate package on GNU ELPA, an email to emacs-devel should work as well.

c-alpha:
> Sounds a bit niche to me.

EUDC is part of core Emacs. Depends how you gauge "niche" then...

> But other than that, wouldn't you prefer to have this completion also supported by `M-x completion-at-point`? If Emacs has some EUDC support already, this might fit better as a patch for `message.el`. See `message-completion-alist` and so on.
> [...]

I presume the same points would equally apply to BBDB (but for which there is a `company-mode` backend, and one for `helm`, too)?

[...]

PR #997, opened 2020-08-05
EUDC Back-End

c-alpha:
This new back-end collects completion candidates for email addresses from EUDC,
and offers them when editing email message header fields that take one or more
email addresses (To, Cc, Bcc, From, and Reply-To) in message modes (configurable
via company-eudc-modes; default are message-mode, notmuch-message-mode, and
mail-mode).

Since EUDC queries may involve multiple remote servers, they can take many
seconds to complete. Hence, two options for using this back-end are provided,
and you must decide which option to use.

The first option adds the back-end to the global list of back-ends, so that EUDC
is queried whenever company-mode attempts completion in a suitable email message
header field. This option is intended for when the additional delay induced by
the EUDC query seems acceptable.

The second option is intended for when the additional delay induced by the EUDC
query does not seem acceptable, and the user therefore will want EUDC to be
queried only upon explicit request. To achieve this, the second option offers a
function that can be bound to a key chord (e.g. C-tab).

### DIFF
--- a/company-eudc.el
+++ b/company-eudc.el
@@ -1,0 +1,155 @@
+;;; company-eudc.el --- company-mode completion backend for EUDC  -*- lexical-binding: t -*-
+
+;; Copyright (c) 2020 condition-alpha.com
+
+;; This program is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+;;
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+;;
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+;;; Commentary:
+;;    This library provides an interface to EUDC. EUDC is the Emacs
+;;    Unified Directory Client, a common interface to directory
+;;    servers and contact information, and part of core Emacs. EUDC
+;;    can consult mutiple servers for each query, providing the
+;;    combined results through a single interface. EUDC has backends
+;;    for LDAP servers, and to the BBDB.
+
+;;; Usage:
+;;    (require 'company-eudc)
+
+;;; Code:
+
+;;{{{      Internal cooking
+
+(require 'company)
+(require 'cl-lib)
+
+(defun company-eudc-email-address-from-alist (rec)
+  "Generate an email address string (\"first last <email>\") from
+  an attribute list."
+  (let* ((name (cdr (assoc 'name rec)))
+	 (firstname (cdr (assoc 'firstname rec)))
+	 (email (cdr (assoc 'email rec)))
+	 (candidate ""))
+    (if firstname
+	(setq candidate (format "%s %s <%s>" firstname name email))
+      (setq candidate (format "%s <%s>" name email)))
+    candidate))
+
+;;}}}
+
+;;{{{      High-level interfaces
+
+(defgroup company-eudc nil
+  "Completion backend for EUDC."
+  :group 'company)
+
+(defcustom company-eudc-modes '(message-mode notmuch-message-mode mail-mode)
+  "Major modes in which `company-eudc' may complete."
+  :type '(repeat (symbol :tag "Major mode")))
+
+;;;###autoload
+(defun company-eudc (command &optional arg &rest ignored)
+  "`company-mode' completion backend for EUDC.
+EUDC is the Emacs Unified Directory Client, a common interface to
+directory servers and contact information.
+
+Completion will be attempted in the To, Cc, Bcc, From, and
+Reply-To header fields of message buffers only. I.e. you can not
+use this backend for completing email addresses outside a message
+header.
+
+To load the `company-eudc' backend, simply do
+
+    (require 'company-eudc)
+
+With this alone, no completions from EUDC will be offered
+yet. You will additionally need to decide how you want to use
+it. Since EUDC can query remote resources (such as e.g. LDAP
+servers), the completion process may take a while. Depending on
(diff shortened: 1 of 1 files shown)


company-eudc.el
```diff
@@ -0,0 +1,155 @@
...
+    (require 'company-eudc)
+    (company-eudc-activate-autocomplete)
+
+If you have configured many and/or slow servers for EUDC, this
+will block Emacs for some time (i.e. until EUDC has delivered its
+results). If this is a frequent issue, and you would like to
+avoid this, use `company-eudc-expand-inline' instead. "
+  (interactive)
+  (add-to-list 'company-backends 'company-eudc)
+  (with-eval-after-load "message"
```
dgutov, line 121: You don't need `with-eval-after-load` to call `add-hook`.

company-eudc.el
```diff
@@ -0,0 +1,155 @@
...
+(defun company-eudc-activate-autocomplete ()
+  "Provide `company-eudc' completions under `company-mode' control.
+This function installs `company-eudc' in the list of
+`company-mode' backends for `message-mode'. Completion candidates
+from EUDC will thus be offered by `company-mode' as any other
+candidate.
+
+To get this behaviour, do
+
+    (require 'company-eudc)
```
dgutov, line 112: If this function is autoloaded, no need to `require`.

company-eudc.el
```diff
@@ -0,0 +1,155 @@
...
+;;;###autoload
+(defun company-eudc-expand-inline ()
+  "Provide `company-eudc' completions in `message-mode' interactively.
+This function triggers `company-mode' completion at point, using
+the `company-eudc' backend only. It is intended for being bound
+to a key chord; for example:
+
+    (require 'company-eudc)
+    (with-eval-after-load \"message\"
+      (define-key message-mode-map (kbd \"<C-tab>\") 'company-eudc-expand-inline))
```
dgutov, line 134: Why not bind `company-eudc` to a key directly?

This function seems unnecessary (though the explanation in the docstring is most likely useful).

company-eudc.el
```diff
@@ -0,0 +1,155 @@
...
+`company-eudc-expand-inline')."
+  (interactive (list 'interactive))
+  (pcase command
+    (`interactive (company-begin-backend 'company-eudc))
+    (`prefix (and (memq major-mode company-eudc-modes)
+		  (let ((case-fold-search t))
+		    (looking-back
+		     "^\\([^ :]*-\\)?\\(To\\|B?Cc\\|From\\|Reply-to\\):.*? *\\([^,;]*\\)"
+		     (line-beginning-position)))
+		  (company-grab-symbol)))
```
alexmurray, line 94: `company-grab-symbol` will only grab say `Murray` when the line is say as follows with point annotated as `|`:

```
To: Alex Murray|
```

I think a better implementation would be to re-use a lot of the bits from `eudc-expand-inline` with something like the following which I think is more robust an similar to the functionality which `eudc` users would be used to:

```
(defun company-eudc (command &optional arg &rest ignored)
  "`company-mode' completion backend for EUDC.
EUDC is the Emacs Unified Directory Client, a common interface to
directory servers and contact information.
Completion will be attempted in the To, Cc, Bcc, From, and
Reply-To header fields of message buffers only. I.e. you can not
use this backend for completing email addresses outside a message
header.
To load the `company-eudc' backend, simply do
    (require 'company-eudc)
With this alone, no completions from EUDC will be offered
yet. You will additionally need to decide how you want to use
it. Since EUDC can query remote resources (such as e.g. LDAP
servers), the completion process may take a while. Depending on
how you have configured `company-mode', this may result in Emacs
being blocked for extensive periods of time. To accommodate this,
the `company-eudc' backend can either install itself into the
list of `company-mode' backends for `message-mode' (see
`company-eudc-activate-autocomplete'), or provide autocompletion
through `company-mode' only when bound to a key (see
`company-eudc-expand-inline')."
  (interactive (list 'interactive))
  (let ((prefix (save-excursion
		    (buffer-substring-no-properties
                     (save-excursion
		       (if (re-search-backward "\\([:,]\\|^\\)[ \t]*"
					       (point-at-bol) 'move)
		           (goto-char (match-end 0)))
		       (point))
                     (point)))))
    (pcase command
      (`interactive (company-begin-backend 'company-eudc))
      (`prefix (and (memq major-mode company-eudc-modes)
		    (let ((case-fold-search t))
		      (looking-back
		       "^\\([^ :]*-\\)?\\(To\\|B?Cc\\|From\\|Reply-to\\):.*? *\\([^,;]*\\)"
		       (line-beginning-position)))
                    prefix))
      (`candidates (eudc-query-with-words
                                     (split-string prefix "[ \t]+")))
      (`sorted t))))
```

company-eudc.el
```diff
@@ -0,0 +1,155 @@
...
+being blocked for extensive periods of time. To accommodate this,
+the `company-eudc' backend can either install itself into the
+list of `company-mode' backends for `message-mode' (see
+`company-eudc-activate-autocomplete'), or provide autocompletion
+through `company-mode' only when bound to a key (see
+`company-eudc-expand-inline')."
+  (interactive (list 'interactive))
+  (pcase command
+    (`interactive (company-begin-backend 'company-eudc))
+    (`prefix (and (memq major-mode company-eudc-modes)
```
alexmurray, line 89: Rather than using `memq` would say `derived-mode-p` be better? ie. just `(derived-mode-p message-mode)` so that say users of `mu4e` don't have to manually add say `mu4e-compose-mode` to `company-eudc-modes`?

yugaego:
Hi @c-alpha, I got to this PR while triaging and reviewing issue #959. Quick search showed that this package wasn't published to any package repository eventually, but stays publicly available [at GitHub](https://github.com/condition-alpha/company-eudc), is that right? Any further plans with it? Thanks!

c-alpha:
> Hi @c-alpha, I got to this PR while triaging and reviewing issue #959. Quick search showed that this package wasn't published to any package repository eventually, but stays publicly available [at GitHub](https://github.com/condition-alpha/company-eudc), is that right?

Yes, that's correct.

> Any further plans with it?

Nope.

yugaego:
Thank you for the confirmation.

Issue #998, opened 2020-08-10
Easy way to determine when there's no selection


kiennq:
Currently `company-tng` is advice against `company-fill-propertize` so that the selection will not be shown.

Doing that affect third party frontends that needs to know about currently highlighted selection.
For example:
- to adding front icon for output from `company--create-lines` and adjust icon background to match with selection.
- Provide different key binding depends on if there's selection or not. Like, press `Enter` when there's no selection will just enter new line, but will enter selected item when there is.

Is there an easy way to determine if there's no selection in company popup?

dgutov:
> to adding front icon for output from company--create-lines and adjust icon background to match with selection.

This one should be doable already. If `company-fill-propertize` is advised, and it calls the icon-generating logic, it could pass a boolean arg (to indicate the current item is selected.

> Provide different key binding depends on if there's selection or not. Like, press Enter when there's no selection will just enter new line, but will enter selected item when there is.

Sounds like you want to bring in that feature from `company-tng` more organically into company-mode proper. It shouldn't be too hard, but I don't have a list of places to change. It could probably start with a user option to choose whether the default selection is `0` or `nil`.

> Is there an easy way to determine if there's no selection in company popup?

Currently, no. This possibility lives purely inside `company-tng`.

dgutov:
> It could probably start with a user option to choose whether the default selection is 0 or nil.

Or it would determine whether most uses of `company-selection` should check `company-selection-changed` is non-nil as well.

kiennq:
> It could probably start with a user option to choose whether the default selection is `0` or `nil`.

Oh, I was thinking the same thing and kind of got it working in my private change.



> Or it would determine whether most uses of `company-selection` should check `company-selection-changed` is non-nil as well.

I'm not really understand this. Do you suggest that I should check `company-selection` and `company-selection-changed` together?
That may not work since:
- Without `company-tng`, `company-selection == 0` and `company-selection-changed = nil` mean first item is **selected**
- With `company-tng`, the same condition mean **no item** is selected



> This one should be doable already. If `company-fill-propertize` is advised, and it calls the icon-generating logic, it could pass a boolean arg (to indicate the current item is selected.

Advising `company-fill-propertize` is a little bit tricky, since one has to advise again that function with lowest depth, else it will not work. Anyway, I don't think advising a function should be official way :)

dgutov:
> I'm not really understand this. Do you suggest that I should check company-selection and company-selection-changed together?

[...]

PR #999, opened 2020-08-11
company-selection: add company-selection-default to have better support for non-selection popup

kiennq:
I've implemented what suggested in https://github.com/company-mode/company-mode/issues/998#issuecomment-671651866.

By adding new var `company-selection-default` that front end like `company-tng` can set on `show` and `hide`, We can make the selection state of `company-mode` temporarily become `nil` when `company-tng` is active.
By that, `company-tng` no longer need to advise `company-fill-propertize` anymore.

Also, by checking `company-selection`, we can have kind of multi binding when press a key and it reacts differently depend on if there's selection or not. E.g; <Enter> key will complete selection or just insert new line when there's no selection.

Let me know if you think this's enough.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,9 @@
 
 ## Next
 
+* `company-tng-mode` has been added and replace the manual method of enabling
+  `company-tng-frontend`. Also, `company-selection` can have null value to
+  indicate no selection.
 * `company-auto-complete` and `company-auto-complete-chars` have been renamed to
   `company-auto-commit` and `company-auto-commit-chars` respectively.
 * `company-clang` filters out duplicates
--- a/company-tng.el
+++ b/company-tng.el
@@ -37,15 +37,11 @@
 ;; need to confirm the entry.
 ;;
 ;; Usage:
+;; Enable`company-tng-mode' with:
+;;   (add-hook 'company-mode-hook 'company-tng-mode)
 ;;
 ;; To apply the default configuration for company-tng call
 ;; `company-tng-configure-default' from your init script.
-;;
-;; You can also configure company-tng manually:
-;;
-;; Add `company-tng-frontend' to `company-frontends':
-;;
-;;   (add-to-list 'company-frontends 'company-tng-frontend)
 ;;
 ;; We recommend to bind TAB to `company-select-next', S-TAB to
 ;; `company-select-previous', and unbind RET and other now-unnecessary
@@ -105,25 +101,22 @@ confirm the selection and finish the completion."
     (show
      (let ((ov (make-overlay (point) (point))))
        (setq company-tng--overlay ov)
-       (overlay-put ov 'priority 2))
-     (advice-add 'company-select-next :before-until 'company-tng--allow-unselected)
-     (advice-add 'company-fill-propertize :filter-args 'company-tng--adjust-tooltip-highlight))
+       (overlay-put ov 'priority 2)))
     (update
-     (let ((ov company-tng--overlay)
-           (selected (nth company-selection company-candidates))
-           (prefix (length company-prefix)))
+     (let* ((ov company-tng--overlay)
+            (selected (and company-selection
+                           (nth company-selection company-candidates)))
+            (prefix (length company-prefix)))
        (move-overlay ov (- (point) prefix) (point))
        (overlay-put ov
                     (if (= prefix 0) 'after-string 'display)
-                    (and company-selection-changed selected))))
+                    selected)))
     (hide
      (when company-tng--overlay
        (delete-overlay company-tng--overlay)
-       (kill-local-variable 'company-tng--overlay))
-     (advice-remove 'company-select-next 'company-tng--allow-unselected)
-     (advice-remove 'company-fill-propertize 'company-tng--adjust-tooltip-highlight))
+       (kill-local-variable 'company-tng--overlay)))
     (pre-command
-     (when (and company-selection-changed
+     (when (and company-selection
                 (not (company--company-command-p (this-command-keys))))
        (company--unread-this-command-keys)
        (setq this-command 'company-complete-selection)))))
@@ -133,65 +126,49 @@ confirm the selection and finish the completion."
 (defvar company-rtags-insert-arguments)
 (defvar lsp-enable-snippet)
 
+(defgroup company-tng nil
+  "Company Tab and Go."
+  :group 'company)
+
+(defcustom company-tng-auto-configure t
+  "Automatically apply default configure when enable `company-tng-mode'."
+  :type 'boolean)
+
 ;;;###autoload
-(defun company-tng-configure-default ()
-  "Applies the default configuration to enable company-tng."
-  (setq company-require-match nil)
-  (setq company-frontends '(company-tng-frontend
-                            company-pseudo-tooltip-frontend
-                            company-echo-metadata-frontend))
-  (setq company-clang-insert-arguments nil
-        company-semantic-insert-arguments nil
-        company-rtags-insert-arguments nil
-        lsp-enable-snippet nil)
-  (advice-add #'eglot--snippet-expansion-fn :override #'ignore)
-  (let ((keymap company-active-map))
-    (define-key keymap [return] nil)
-    (define-key keymap (kbd "RET") nil)
-    (define-key keymap [tab] 'company-select-next)
-    (define-key keymap (kbd "TAB") 'company-select-next)
-    (define-key keymap [backtab] 'company-select-previous)
-    (define-key keymap (kbd "S-TAB") 'company-select-previous)))
+(define-obsolete-function-alias 'company-tng-configure-default 'company-tng-mode "0.9.14"
+  "Applies the default configuration to enable company-tng.")
 
-(defun company-tng--allow-unselected (&optional arg)
-  "Advice `company-select-next' to allow for an 'unselected'
-state. Unselected means that no user interaction took place on the
-completion candidates and it's marked by setting
-`company-selection-changed' to nil. This advice will call the underlying
-`company-select-next' unless we need to transition to or from an unselected
-state.
-
-Possible state transitions:
-- (arg > 0) unselected -> first candidate selected
-- (arg < 0) first candidate selected -> unselected
-- (arg < 0 wrap-round) unselected -> last candidate selected
-- (arg < 0 no wrap-round) unselected -> unselected
-
-There is no need to advice `company-select-previous' because it calls
-`company-select-next' internally."
+;;;###autoload
+(define-minor-mode company-tng-mode
+ "This minor mode enables `company-tng-frontend'."
+  :init-value nil
+  :global t
   (cond
-   ;; Selecting next
-   ((or (not arg) (> arg 0))
-    (unless company-selection-changed
-      (company-set-selection (1- (or arg 1)) 'force-update)
-      t))
-   ;; Selecting previous
-   ((< arg 0)
-    (when (and company-selection-changed
-               (< (+ company-selection arg) 0))
-      (company-set-selection 0)
-      (setq company-selection-changed nil)
-      (company-call-frontends 'update)
-      t)
-    )))
-
-(defun company-tng--adjust-tooltip-highlight (args)
-  "Prevent the tooltip from highlighting the current selection if it wasn't
-made explicitly (i.e. `company-selection-changed' is true)"
-  (unless company-selection-changed
-    ;; The 4th arg of `company-fill-propertize' is selected
-    (setf (nth 3 args) nil))
-  args)
+   (company-tng-mode
+    (setq company-frontends
+          (add-to-list 'company-frontends 'company-tng-frontend))
+    (when company-tng-auto-configure
+      (setq company-require-match nil)
+      (setq company-frontends '(company-tng-frontend
+                                company-pseudo-tooltip-frontend
+                                company-echo-metadata-frontend))
+      (setq company-clang-insert-arguments nil
+            company-semantic-insert-arguments nil
+            company-rtags-insert-arguments nil
+            lsp-enable-snippet nil)
+      (advice-add #'eglot--snippet-expansion-fn :override #'ignore)
+      (let ((keymap company-active-map))
+        (define-key keymap [return] nil)
+        (define-key keymap (kbd "RET") nil)
+        (define-key keymap [tab] 'company-select-next)
+        (define-key keymap (kbd "TAB") 'company-select-next)
+        (define-key keymap [backtab] 'company-select-previous)
+        (define-key keymap (kbd "S-TAB") 'company-select-previous)))
+    (setq company-selection-default nil))
+   (t
+    (setq company-frontends
+          (delete 'company-tng-frontend company-frontends))
+    (setq company-selection-default 0))))
 
 (provide 'company-tng)
 ;;; company-tng.el ends here
--- a/company.el
+++ b/company.el
@@ -757,9 +757,10 @@ asynchronous call into synchronous.")
                           (company-candidates
                            (:eval
                             (if (consp company-backend)
-                                (company--group-lighter (nth company-selection
-                                                             company-candidates)
-                                                        company-lighter-base)
+                                (when company-selection
+                                  (company--group-lighter (nth company-selection
+                                                               company-candidates)
+                                                          company-lighter-base))
                               (symbol-name company-backend)))
                            company-lighter-base))
   "Mode line lighter for Company.
@@ -1108,7 +1109,9 @@ matches IDLE-BEGIN-AFTER-RE, return it wrapped in a cons."
 
 (defvar-local company-common nil)
 
-(defvar-local company-selection 0)
+(defvar company-selection-default 0
+  "The default value for `company-selection'.")
+(defvar-local company-selection company-selection-default)
 
 (defvar-local company-selection-changed nil)
 
@@ -1192,10 +1195,20 @@ can retrieve meta-data for them."
                     frontend (error-message-string err) command)))))
 
 (defun company-set-selection (selection &optional force-update)
-  (setq selection
-        (if company-selection-wrap-around
-            (mod selection company-candidates-length)
-          (max 0 (min (1- company-candidates-length) selection))))
+  "Set SELECTION for company candidates.
+This will update `company-selection' and related variable.
+Only update when the current selection is changed, but optionally always
+update if FORCE-UPDATE."
+  (when selection
+    (let* ((offset (if company-selection-default 0 1))
+           (company-candidates-length
+            (+ company-candidates-length offset)))
+      (setq selection
+            (if company-selection-wrap-around
+                (mod selection company-candidates-length)
+              (max 0 (min (1- company-candidates-length) selection))))
+      (setq selection (unless (< selection offset)
+                        (- selection offset)))))
   (when (or force-update (not (equal selection company-selection)))
     (setq company-selection selection
           company-selection-changed t)
@@ -1214,10 +1227,11 @@ can retrieve meta-data for them."
   (setq company-candidates-length (length candidates))
   (if company-selection-changed
       ;; Try to restore the selection
-      (let ((selected (nth company-selection company-candidates)))
-        (setq company-selection 0
-              company-candidates candidates)
+      (let ((selected (and company-selection
+                           (nth company-selection company-candidates))))
+        (setq company-candidates candidates)
         (when selected
+          (setq company-selection 0)
           (catch 'found
             (while candidates
               (let ((candidate (pop candidates)))
@@ -1226,9 +1240,9 @@ can retrieve meta-data for them."
                                   (company-call-backend 'annotation selected)))
                   (throw 'found t)))
               (cl-incf company-selection))
-            (setq company-selection 0
+            (setq company-selection company-selection-default
                   company-selection-changed nil))))
-    (setq company-selection 0
+    (setq company-selection company-selection-default
           company-candidates candidates))
   ;; Calculate common.
   (let ((completion-ignore-case (company-call-backend 'ignore-case)))
@@ -1666,7 +1680,7 @@ prefix match (same case) will be prioritized."
           company-candidates-cache nil
           company-candidates-predicate nil
           company-common nil
-          company-selection 0
+          company-selection company-selection-default
           company-selection-changed nil
           company--manual-action nil
           company--manual-prefix nil
@@ -1871,11 +1885,12 @@ each one wraps a part of the input string."
     (company-update-candidates cc)))
 
 (defun company--search-update-string (new)
-  (let* ((pos (company--search new (nthcdr company-selection company-candidates))))
+  (let* ((selection (or company-selection 0))
+         (pos (company--search new (nthcdr selection company-candidates))))
     (if (null pos)
         (ding)
       (setq company-search-string new)
-      (company-set-selection (+ company-selection pos) t))))
+      (company-set-selection (+ selection pos) t))))
 
 (defun company--search-assert-input ()
   (company--search-assert-enabled)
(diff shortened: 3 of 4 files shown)


company-tng.el
```diff
@@ -106,12 +106,15 @@ confirm the selection and finish the completion."
      (let ((ov (make-overlay (point) (point))))
        (setq company-tng--overlay ov)
        (overlay-put ov 'priority 2))
-     (advice-add 'company-select-next :before-until 'company-tng--allow-unselected)
-     (advice-add 'company-fill-propertize :filter-args 'company-tng--adjust-tooltip-highlight))
+     (setq company-selection-default nil)
+     (unless company-selection-changed
+       (setq company-selection company-selection-default))
```
dgutov, line 111: Do we need to do this here? I imagine setting it to the "correct" value in `company-update-candidates` should be enough.

company-tng.el
```diff
@@ -106,12 +106,15 @@ confirm the selection and finish the completion."
...
+     (advice-add 'company-select-next :before-until 'company-tng--allow-unselected))
```
dgutov, line 112: Perhaps we should merge the `company-tng--allow-unselected` advice into the definition of `company-select-next`?

company-tng.el
```diff
@@ -106,12 +106,15 @@ confirm the selection and finish the completion."
...
     (update
-     (let ((ov company-tng--overlay)
-           (selected (nth company-selection company-candidates))
-           (prefix (length company-prefix)))
+     (let* ((ov company-tng--overlay)
+            (selection (or company-selection 0))
```
dgutov, line 115: I'm guessing the `or` here is actually unnecessary. Or should be, because this action should only be possible once the user has selected something.

We could add an assert here, though. For confidence.

company.el
```diff
@@ -1091,7 +1091,9 @@ matches IDLE-BEGIN-AFTER-RE, return it wrapped in a cons."
 
 (defvar-local company-common nil)
 
-(defvar-local company-selection 0)
+(defvar company-selection-default 0
```
dgutov, line 1094: We could also make this a boolean, since there are only two valid values here, really. That would make initializing `company-selection` just a little bit more complicated.

company.el
```diff
@@ -2141,7 +2143,8 @@ With ARG, move by that many elements."
   "Insert the selected candidate."
   (interactive)
   (when (company-manual-begin)
-    (let ((result (nth company-selection company-candidates)))
+    (let* ((selection (or company-selection 0))
```
dgutov, line 2146: Should we do something with commands like this, too?

Perhaps port even more code from `company-tng`.

dgutov:
> Let me know if you think this's enough.

Possibly yes, but we could do more. I like the direction, either way.

It will need some tests as well, though.

company-tng.el
kiennq: The problem is `company-update-candidates` is called before `(company-call-frontends 'show)`, so the `company-selection-default` is not set to correct value at that time.
How about I introduce new `pre-show` command for front end so it can be used to setup `company-selection-default`?

dgutov: Then it shouldn't be set in the `show` handler. Frontends are a poor vehicle for changing behavior anyway.

It should be set in `company-tng-configure-default`. All tng users seem to have to call it.

Or we could add `company-tng-mode`, and it would be configured there.

company.el
kiennq: Yes, but I think we can put another PR for that. Doing that now will make this PR more complicated.

company-tng.el
kiennq: Done, I've checked the `selected` with `company-selection`

company.el
kiennq: This command should not run when there's no selection. Checked that.

company-tng.el
kiennq: I've merged a part of going to another selection from unselected state. For going back to unselected state, I left that in `company-tng`. Since that should be exclusive behavior of `company-tng`

company-tng.el
kiennq: Add `company-tng-mode`

company.el
dgutov: Ok.

company-tng.el
dgutov: Its docstring now says it's a frontend..?

company-tng.el
dgutov: It would have allowed us to dispense with the advices altogether.

In any case, this doesn't look like a very important feature. Let's maybe ask the author. @nikital, do you ever use it in practice?

dgutov:
Speaking of tests, we'll at least need some for `company-set-selection`. Since the added behavior is not normally executed, there's a significant risk of breaking it without test coverage.

company-tng.el
kiennq: Do you have suggestion?

company-tng.el
kiennq: I sometime do use this though

kiennq:
> Speaking of tests, we'll at least need some for `company-set-selection`. Since the added behavior is not normally executed, there's a significant risk of breaking it without test coverage.

I agree. Also I notice that no CI is run for this PR, is that because travis stopped or do you plan to move to another CI soon?

company.el
```diff
@@ -2268,11 +2280,12 @@ character, stripping the modifiers.  That character must be a digit."
 (defvar-local company-last-metadata nil)
 
 (defun company-fetch-metadata ()
-  (let ((selected (nth company-selection company-candidates)))
-    (unless (eq selected (car company-last-metadata))
-      (setq company-last-metadata
-            (cons selected (company-call-backend 'meta selected))))
-    (cdr company-last-metadata)))
+  (when company-selection
+    (let ((selected (nth company-selection company-candidates)))
```
nikital, line 2284: I think that when nothing is selected, it makes sense to show the metadata for the first candidate anyway.
This way you'll see useful info right away and I don't see any downsides to it.
Therefore I suggest deleting `(when company-selection` and instead doing

    (let ((selected (nth (or company-selection 0) company-candidates)))

company.el
```diff
@@ -2321,16 +2334,17 @@ character, stripping the modifiers.  That character must be a digit."
...
-             (doc-buffer (or (company-call-backend 'doc-buffer selected)
-                             (user-error "No documentation available")))
-             start)
-        (when (consp doc-buffer)
-          (setq start (cdr doc-buffer)
-                doc-buffer (car doc-buffer)))
-        (setq other-window-scroll-buffer (get-buffer doc-buffer))
-        (let ((win (display-buffer doc-buffer t)))
-          (set-window-start win (if start start (point-min))))))))
+      (when company-selection
```
nikital, line 2337: Nitpick, feel free to disregard:
Since calling this function shouldn't do anything without a selection, I suggest moving the check for `company-selection` to the beginning of the function. I think that generally code that exits as early as possible is easier to reason about, so instead of aborting inside `electric-do` it's preferable to abort before.

company.el
```diff
@@ -2775,23 +2789,25 @@ If SHOW-VERSION is non-nil, show the version in the echo area."
...
+        (setq scrollbar-bounds (company--scrollbar-bounds company-tooltip-offset
+                                                          limit len)))
+       ((eq company-tooltip-offset-display 'lines)
+        (when (> company-tooltip-offset 0)
+          (setq previous (format "...(%d)" company-tooltip-offset)))
+        (setq remainder (- len limit company-tooltip-offset)
+              remainder (when (> remainder 0)
+                          (setq remainder (format "...(%d)" remainder))))))
+
+      (cl-decf selection company-tooltip-offset))
```
nikital, line 2809: This change introduces a small regression to scrollbar rendering, since no scrollbar will be displayed unless something is selected. Then, when a user performs a selection, the scrollbar will pop into existence.

For the purpose of scrolling, I would argue that no selection is the same as 0 (because we want to render the start of the candidate list). Therefore I suggest that instead of disabling this code entirely, it's enough to replace `company-selection` with `(or company-selection 0)` in this section.

Then, the last line (the `cl-decf` line) should be wrapped in `(when selection`. I tested my proposal locally and it fixes the rendering.

company.el
```diff
@@ -3169,59 +3187,59 @@ Delay is determined by `company-tooltip-idle-delay'."
...
-                                 '(face company-echo-common) comp))
-        (setq comp (propertize comp 'face 'company-echo))
-        (add-text-properties 0 (string-width company-common)
-                             '(face company-echo-common) comp))
-      (if (>= len limit)
-          (setq candidates nil)
-        (push comp msg)))
-
-    (mapconcat 'identity (nreverse msg) " ")))
+  (when company-selection
```
nikital, line 3190: This causes the `echo-fronted` to behave strangely, since nothing will be rendered unless a selection is performed. (I tried it with `(setq company-frontends '(company-tng-frontend company-echo-frontend))`.)

For the purpose of rendering, I would argue that no selection is the same as 0: We want to render the candidates for the beginning.
Therefore I suggest deleting `(when company-selection` and instead doing `(or company-selection 0)` in this function.

company.el
```diff
@@ -3169,59 +3187,59 @@ Delay is determined by `company-tooltip-idle-delay'."
...
-        (cl-incf len 4)
-        (cl-incf i))
-      (if (>= len limit)
-          (setq candidates nil)
-        (push (propertize comp 'face 'company-echo) msg)))
-
-    (concat (propertize company-prefix 'face 'company-echo-common) "{"
-            (mapconcat 'identity (nreverse msg) ", ")
-            "}")))
+  (when company-selection
```
nikital, line 3220: Same comment as in `company-echo-format`

company-tng.el
nikital: @dgutov Personally, this feature is useful. Sometimes I would select the first candidate, ponder for a moment, unselect it and continue typing. Since there is no way in `tng` to cancel a completion once something is selected, this provides a nice way out (`C-g` would close the popup but leave the completion inserted).

kiennq:
@nikital Thanks. I've updated as your suggestions

company-tng.el
```diff
@@ -105,23 +101,20 @@ confirm the selection and finish the completion."
...
     (hide
      (when company-tng--overlay
        (delete-overlay company-tng--overlay)
-       (kill-local-variable 'company-tng--overlay))
-     (advice-remove 'company-select-next 'company-tng--allow-unselected)
-     (advice-remove 'company-fill-propertize 'company-tng--adjust-tooltip-highlight))
+       (kill-local-variable 'company-tng--overlay)))
     (pre-command
      (when (and company-selection-changed
                 (not (company--company-command-p (this-command-keys))))
```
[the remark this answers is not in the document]
kiennq: Thanks, this is definitely a big bug.

company.el
dgutov: I kinda worry that it might be confusing.

But if it's really useful in your experience, let's keep it that way for now.

company-tng.el
nikital: Regarding the docstring, rather than copying the one from `company-tng-frontend` I think it will be clearer if says something like  "This minor mode enables \`company-tng-frontend' ".
So that users won't try to add `company-tng-mode` as a frontend by mistake.

kiennq: Done

NEWS.md
```diff
@@ -12,6 +12,8 @@
   users these days, with
   ([emacs-eclim](https://github.com/emacs-eclim/emacs-eclim/)) declared obsolete
   in favor of `lsp-java`. Though it used its own backend anyway.
+* `company-tng-mode` has been added and replace the manual method of enabling
+  `company-tng-frontend`.
```
dgutov, line 16: We add NEWS entries at the top here.

company-tng.el
dgutov: Nikita gave one. I'll probably edit the docstring later, too.

More importantly, do we have a reason for `company-tng-mode` not to set up the bindings automatically, like `company-tng-configure-default` did? Do you edit the bindings later yourself?

IIUC, `company-tng-configure-default` is not functional by itself anymore. So it would be a good idea to make it an obsolete alias for `company-tng-mode`, so that users' configurations don't break right away after the upgrade.

dgutov:
> > Speaking of tests, we'll at least need some for `company-set-selection`. Since the added behavior is not normally executed, there's a significant risk of breaking it without test coverage.
> 
> I agree. Also I notice that no CI is run for this PR, is that because travis stopped or do you plan to move to another CI soon?

Builds still continue: https://travis-ci.org/github/company-mode/company-mode/pull_requests

For some reason, they are not shown on GitHub anymore, though. Perhaps I'll have to redo the Travis integration.

company-tng.el
kiennq: I think that `company-tng` needs this as they need to cancel completion once something has been selected.
Other backends don't do this and actually wait for user explicit completion completing action.

NEWS.md
kiennq: Done

company-tng.el
kiennq: I made it obsolete since `0.9.14`, please update the version and docstring as you see fit.

company-tng.el
```diff
@@ -133,25 +126,17 @@ confirm the selection and finish the completion."
 (defvar company-rtags-insert-arguments)
 (defvar lsp-enable-snippet)
 
+(defgroup company-tng nil
+  "Company Tab and Go."
+  :group 'company)
+
+(defcustom company-tng-auto-configure t
+  "Automatically apply default configure when enable `company-tng-mode'."
+  :type 'boolean)
```
nikital, line 135: When `company-tng-auto-configure` is `t`, enabling and then immediately disabling `company-tng-mode` leaves undoable side-effects on the configuration, which affects company in general. Usually I expect the effects of a minor mode to go away after I disable it.
I wonder if it would be better to leave this `nil`, and then set it to `t` if the user calls `company-tng-configure-default`, since that function implies that the user wants the extra configuration.
I don't have a strong opinion (since I also see benefits for leaving it `t` by default), I'm just wondering what you guys think about it.

dgutov: There's also an option of saving the previous configuration somewhere, and restoring it when the mode is disabled.

dgutov: Regarding the bindings specifically, maybe it could replace `company-active-map` wholesale with a tng-specific map (which inherits from it, of course). And restore it, of course, when the mode is disabled.

nikital: Not sure if it's really worth going this far... It's not like `company-tng` is toggled on and off frequently or as part of normal usage. Also, won't a new map break existing users of `company-tng` in case they customized key-bindings?

kiennq: Since we specifically mentioned `company-tng-configure-default` is an obsoleted alias of `company-tng-mode`, I think this should be `t` by default.
If most of `company-tng` user is already use `company-tng-configure-default`, I think this only affect the minimal number of user who manually add `company-tng-frontend` only. Even in that case, setting `company-tng-auto-configure` to `nil` will solve this problem.
Perhaps we can do better document so user aware of the auto configure stuff?

dgutov: But what's the use case of setting it to `nil`? Redoing the whole configuration manually?

If the user is willing to do that, they could as well change `company-frontends` as appropriate and set `company-selection-default` to nil, too. If `company-tng--allow-unselected` becomes unnecessary, not much else will be needed. That would only leave Eglot requiring an advice, but I'd be happy to direct anybody complaining about that to Eglot's issue tracker.

kiennq: > But what's the use case of setting it to `nil`? Redoing the whole configuration manually?

Not necessarily everything, but only needed thing.
For example, I'm more favorable of using snippets with `lsp-mode` than just a few glitch on the overlay preview is not displayed correctly, so I'm not setting any `lsp-enable-snippet` to `nil`.
The end result is that I still got the `tab and go` functionality of selecting a candidate and then can type the next one immediately, with the cost of some visual glitch that can easily ignored.


Also, I think that `company-selection-default` should only be for `company` plugin package to use only, user should not set that. Since setting `company-selection-default` without a frontend that know how deal with it.
For example:
- `company-tng` is showing some cues in form of overlay.
- Or a new front end that can support multiple actions on selecting a candidate, like: pressing <Enter> when there's no selection will just insert new line, but if there's selection, it will show a popup menu asking what user want to do with this candidate. In that case, the new frontend can set `company-selection-default` to `nil` and benefit from that.

The `company-tng--alow-unselected` is tricky, since only `tng` show overlay preview and allow user to select and commit the candidate immediately, thus it will need the `allow-unselected` to go to unselected state from selected. The imagine frontend that I took as example a few lines ago will not need this functionality.

dgutov: > I'm more favorable of using snippets with lsp-mode than just a few glitch on the overlay preview is not displayed correctly, so I'm not setting any lsp-enable-snippet to nil

But if you disable the auto-configuration, you will have to include the rest of the configuration in your config, won't you? That's the "everything" that I meant.

Whereas if `company-tng-mode` always did its whole thing, you could just add a short function to `company-tng-mode-hook` (for example) to tweak the 1-2 values you want to be different.

> I think that company-selection-default should only be for company plugin package to use only

That also weighs in favor of the approach described above.

> but if there's selection, it will show a popup menu

Frontends normally should limit themselves to visualization, at least while there is no good way to implement something like this in a frontend. But OK, let's consider this potentiality as well.

> The imagine frontend that I took as example a few lines ago will not need this functionality.

But will it be hindered by it?

kiennq: > But if you disable the auto-configuration, you will have to include the rest of the configuration in your config, won't you? That's the "everything" that I meant.

The large part of `company-tng-configure-default` are settings external package variables and add new key bindings to `company-active-map`. If I'm happy with the old key bindings and don't want external package variables to be set, isn't it easier to just disable the auto configuration and no need to do anything else?
The `company-tng-frontend` is always added into `company-fronends` list, regardless of auto configuration or not.

Before user can easily ignore all auto configuration stuff by do not call `company-tng-configure-default`. Now I think we should also provide alternative method to do that without much trouble of reset everything in `company-tng-mode-hook`.



> > The imagine frontend that I took as example a few lines ago will not need this functionality.
> 
> But will it be hindered by it?

Probably not, since the new imaginary frontend is not having functionality of auto commit the selected candidate.

To be honest, merging the `company-tng--alow-unselected` with `company-select-next` is tricky and not that clean, since it has to check for `company-selection-default` value and act based on that. Other command no need to do that and only need to check for `company-selection` value to be `nil` or not.

dgutov: > If I'm happy with the old key bindings and don't want external package variables to be set, isn't it easier to just disable the auto configuration and no need to do anything else?

Seems like I misunderstood you, and you not only want to keep the snippet expansion as-is, but keep the default key bindings as well? If we do expect this to be a significant use case for `company-tng-mode` users, then yes, you guys' suggestion to add the `-configure-default` variable and default it to `t` sounds good.

The function that would be called when this variable is `t` should be called something else, though. And `company-tng-configure-default` should be an obsolete alias for `company-tng-mode` to ease the transition.

> To be honest, merging the company-tng--alow-unselected with company-select-next is tricky and not that clean, since it has to check for company-selection-default value and act based on that.

Right, but if `company-selection` being nil is a valid core behavior now, why wouldn't it check that? And we would benefit from having a test for this.

kiennq: > Right, but if `company-selection` being nil is a valid core behavior now, why wouldn't it check that

I mean checking to make sure `company-select-next` is behave correctly when `company-selection` is `nil` is already integrated, and quite straight forward.
However, the part of make `company-selection` to become `nil` again from other value in `company-select-next` is not clean and too dependent on `company-tng`.



> And we would benefit from having a test for this.

I'm working on adding test for this

nikital: @kiennq How is `company-selection` transitioning to `nil` is more dependent on `company-tng` than transitioning from `nil`?
If the latter is implemented inside core company, I agree with @dgutov that it makes sense to implement the former, for symmetry. And then get rid of the advice altogether.

kiennq: @nikital It's because while `company-selection` from `nil` is quite straight forward, the transition to `nil` required checking value of `company-selection-default`, `company-selection-wrap-around` to determine the outcome.
For the details:
- Currently `company-selection-next` just calculate the selection and pass to `company-set-selection` which will calculating `company-selection-wrap-around` to correct index and set it. (1)
- If we want to support transition to `nil`, we will have to check the value of `company-selection-default` (which is not checked in any function, we only check value of `company-selection`), calculating the desired selection regarding to `company-selection-wrap-around` and pass it to `company-set-selection`. (2)

Trying to force (2) in `company-selection-next` make the function become more bulky and not clean at all.

Since `company-selection` can be `nil` we have to handle the case of transition from `nil` anyway, but the case of transition to `nil` is kind of `company-tng` specialty, since without it, `company-tng` cannot easily cancel the completion once selected.
The other front end may have `company-selection-default` to `nil` but if it doesn't have same behavior of committing selection immediately like `company-tng`, it will not require transition to `nil`.

dgutov: @kiennq I think I see your point, if you're saying that `company-tng` currently doesn't obey `company-selection-wrap-around`, and it seems messy to fix that. I'm not going to insist you do that as part of this PR.

But the mess is here already (again: `company-tng` doesn't obey a documented user option), and the conflicts will need to be resolved anyway. Keeping the behavior in an advice doesn't make the whole body of code less "bulky" and "not clean".

So, up to you, but it would be great to choose a behavior that honors both tng and wrap-around. Yes, it will have to check the value of `company-selection-default`. But it's a variable from the same package, so there's no ugliness in that.

kiennq: Okay, I've added test and also remove the `advice-add`. It should be better now

test/core-tests.el
```diff
@@ -591,3 +591,128 @@
...
+
+    (company-select-next -2)
+    (should (eq company-selection 7))
+
+    ;; Nil selection
+    (setq company-selection nil)
+    (company-select-next 5)
+    (should (eq company-selection 4))
+
+    (setq company-selection nil)
```
dgutov, line 650: When the default is `nil`, this case seem unreachable in practice.

company-tng.el
dgutov: Thanks!

What's your situation with copyright assignment to FSF?

kiennq: I've completed it.

test/core-tests.el
kiennq: Yes, but no hurt in testing it, to make sure if someone somehow be able to force it.

company.el
```diff
@@ -2034,10 +2048,27 @@ followed by `company-search-toggle-filtering'."
...
+- (arg < 0 no wrap-round) unselected -> unselected"
   (interactive "p")
   (when (company-manual-begin)
-    (company-set-selection (+ (or arg 1) company-selection))))
+    (let ((selection (if company-selection
+                         (+ (or arg 1) company-selection)
+                       (+ (or arg 1)
+                          (cond ((or (not arg) (> arg 0)) -1)
+                                (t 0))))))
+      ;; special case
```
dgutov, line 2065: This indeed turned out to be pretty complex.

If we allowed `(arg = 1) last candidate selected -> default selection`, would it make things a bit more uniform and, as such, simpler?

company.el
```diff
@@ -2325,18 +2356,19 @@ character, stripping the modifiers.  That character must be a digit."
...
-             (doc-buffer (or (company-call-backend 'doc-buffer selected)
-                             (user-error "No documentation available")))
-             start)
-        (when (consp doc-buffer)
-          (setq start (cdr doc-buffer)
-                doc-buffer (car doc-buffer)))
-        (setq other-window-scroll-buffer (get-buffer doc-buffer))
-        (let ((win (display-buffer doc-buffer t)))
-          (set-window-start win (if start start (point-min))))))))
+  (when company-selection
```
dgutov, line 2359: Perhaps we should add some user-error to commands which are explicitly invoked by the user, and which really need a selected completion to perform.

Or, failing that, default to 0, like with the `meta` case.

company.el
dgutov: Sorry for dawdling on the review. In other situation I'd merge and then fiddle myself later, but here we have two users of the feature in question in the discussion. And I am not one of them.

company.el
kiennq: Changed to default to `0` like `meta` case.

company.el
nikital: I tried to simplify this code so that the special handling of selected <-> unselected is handled in one function (`company--clamp-selection`) in a way that leaves `company-select-next` almost unmodified.
Please have a look and let me know if you like it better.
https://github.com/nikital/company-mode/commit/c19fa33076c1230761be5f44634b7bc11e1170e0

kiennq: Thank you @nikital. Sorry I didn't notice this comment earlier, I went to kind of a same approach to you and push a new commit. Allow forward selection to nil selection state does make thing easier.
@dgutov Can you help reviewing?

dgutov: Thank you both.

@nikital It's hard for me to directly compare both patches, but there are definitely things to like in yours (in particular, the generous commentary). If you believe @kiennq's latest version can be improved still, please submit a new PR on top of the new master, and we'll continue there. Thank you either way.

It's time to merge, let's get this in and collect user feedback, if any. Thanks to @kiennq for the work and patience.

Issue #1013, opened 2020-09-28
Visual bugs with completion menu


ludwj61:
![image](https://user-images.githubusercontent.com/42984254/94436314-4d4d3a00-016a-11eb-9021-75dcbc6b09ba.png)
![image](https://user-images.githubusercontent.com/42984254/94436177-20008c00-016a-11eb-800c-4580702c8ca4.png)

As in the images, the completion menu does a few weird things. 

First, it makes line numbers disappear (1st image) (and I couldn't get it to happen but I feel like I've seen it do other weird things with line numbers, like make extra show up if the completion menu appears at the end of a file.

Second, the background of the completion menu extends the entirety of the buffer (both images).

Third, the background of the completion menu extends past the end of completion menus with few options (2nd image).

I am on Arch Linux (btw) with i3gaps.
Emacs 27.1 (occurs with native comp and pgtk as well)
I am using DOOM emacs config

dgutov:
Hi!

The line numbers thing is expected, it's a limitation of overlay-based rendering. Use `company-posframe` or `company-box` to avoid it.

The background things, however, shouldn't be. Can you reproduce it with the default theme?

ludwj61:
Thanks for the help. It actually _doesnt_ happen with the default theme, I never thought it would be theme related as it happens with all of doom themes. I guess this is something that repo has to fix, then. 

Would you happen to know if there's any known tweak I can make to the offending themes to fix this myself?

dgutov:
Could you give an example of such theme? I'll take a look then.

ludwj61:
sure, here is one https://github.com/ianpan870102/vscode-dark-plus-emacs-theme

dgutov:
Hi again!

I've installed the theme from MELPA (20200926.900) and tried to reproduce this, with company-mode from master, and both Emacs 27.1 and Emacs 28 (master). And I couldn't.

First, are you sure that no alternative package for popup is used, such as `company-box`? There was a recent report about that.

If not, could you provide a repro step-by-step starting with vanilla Emacs (`emacs -Q`)? Both to verify yourself that no extra settings in Doom cause this, or if they do, to include them in the scenario.

And BTW, is that really the right theme? This one I installed uses different colors than those that I see on your screenshots. E.g. light yellow for method names (which are purple on the screenshots). Another difference is I don't see a special background under the current line's number. And the color of that background (on your screenshots) suspiciously matches the color of the glitch (and not the popup's background color).

[...]

PR #1020, opened 2020-10-06
Fix tooltip background stretching to EOL

### LINKED ISSUE #1019: Tooltip background is applied to EOL when default face is remapped
https://github.com/company-mode/company-mode/commit/e1d58370cf1f6b42b929376eab2193bfc0914381 (the fix for #996) introduced [an issue](https://github.com/hlissner/emacs-solaire-mode/issues/37) that is present when the `default` face is remapped using the `face-remap` library:

image

The overlay's background is fontified to EOL.

This can be fixed by either:

1. Removing the hard-coded `:background`:

   ```diff
        (let* ((nl-face (list
                         :extend t
   -                     :inverse-video nil
   -                     :background (face-attribute 'default :background)))
   +                     :inverse-video nil)
               (str (apply #'concat
                           (when nl " \n")
                           (cl-mapcan
   ```

2. Or looking up `default`'s remap:

   ```diff
        (let* ((nl-face (list
                         :extend t
                         :inverse-video nil
   -                     :background (face-attribute 'default :background)
   +                     :background
   +                     (face-attribute
   +                      (if-let (remap (cadr (assq 'default face-remapping-alist)))
   +                          (if (keywordp (car-safe remap))
   +                              (plist-get remap :background)
   +                            remap)
   +                        'default)
   +                      :background
   +                      nil t)))
               (str (apply #'concat
                           (when nl " \n")
                           (cl-mapcan
   ```

Would you be open to a PR for one of these solutions? (I'll factor out the `if-let` for backwards compatibility if I do)

----

Just realized this is the same issue as https://github.com/company-mode/company-mode/issues/1013.

hlissner:
Adds `company--face-attribute FACE ATTR`, which resolves any remappings of FACE (with face-remap) recursively. This solution is slightly different to what was suggested in #1019, but handles more edge cases.

This fixes #1019, #1013, https://github.com/hlissner/emacs-solaire-mode/issues/37 and https://github.com/hlissner/doom-emacs/issues/4013

----

The function is 15L, but altogether 17L are modified. Whether or not I need to do the FSF CA for this, it's probably a good idea I do (won't be the last time I contribute to GPL'd code). How would I go about doing that?

### DIFF
--- a/company.el
+++ b/company.el
@@ -2746,6 +2746,22 @@ If SHOW-VERSION is non-nil, show the version in the echo area."
       (cl-decf ww (1- (length (aref buffer-display-table ?\n)))))
     ww))
 
+(defun company--face-attribute (face attr)
+  ;; Like `face-attribute', but accounts for faces that have been remapped to
+  ;; another face, a list of faces, or a face spec.
+  (cond ((symbolp face)
+         (let ((remap (cadr (assq face face-remapping-alist))))
+           (if remap
+               (company--face-attribute remap attr)
+             (face-attribute face attr nil t))))
+        ((keywordp (car-safe face))
+         (or (plist-get face attr)
+             (company--face-attribute (plist-get face :inherit))))
+        ((listp face)
+         (cl-find-if #'stringp
+                     (mapcar (lambda (f) (company--face-attribute f attr))
+                             face)))))
+
 (defun company--replacement-string (lines old column nl &optional align-top)
   (cl-decf column company-tooltip-margin)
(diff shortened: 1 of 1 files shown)


dgutov:
Excellent! Thank you.

dgutov:
> How would I go about doing that?

Please follow these instructions: https://code.orgmode.org/bzg/org-mode/raw/master/request-assign-future.txt

(Keep "Emacs" as the answer to the first question.)

PR #1023, opened 2020-10-13
Fix company--face-attribute helper

### LINKED ISSUE #1022: Completion fails with more than one candidate: "frontend company-pseudo-tooltip-unless-just-one-frontend error "Wrong type argument: listp, go-mode-default" on command post-command"
A few days ago Company stopped presenting completion popups in a few modes I use regularly—most notably, _go-mode_. Whenever company would normally present such a popup, I see the following messages echoed in the modeline and printed in the _\*Messages\*_ buffer.
```
Company: An error occurred in post-command
Company: frontend company-pseudo-tooltip-unless-just-one-frontend error "Wrong type argument: listp, go-mode-default" on command post-command
```
After that, no popup appears.

If I'm reading that correctly, there's a problem with the Emacs face `go-mode-default`, where we expected it to be a list, but it's not. I've seen the same error in _emacs-lisp-mode_ with the face `emacs-lisp-default`.

I'm using Aquamacs version 3.5 on macOS. When I run <kbd>M-x describe-face RET go-mode-default RET</kbd>, I see the following:
<details>
<summary>go-mode-default</summary>

```
Face: go-mode-default (sample) (customize this face)

Documentation:
mode-specific face activated by ‘aquamacs-autoface-mode’.
The ‘default’ face is remapped (in the appropriate buffers) to this face.

           Family: unspecified
          Foundry: unspecified
            Width: unspecified
           Height: unspecified
           Weight: unspecified
            Slant: unspecified
       Foreground: unspecified
DistantForeground: unspecified
       Background: unspecified
        Underline: unspecified
         Overline: unspecified
   Strike-through: unspecified
              Box: unspecified
          Inverse: unspecified
          Stipple: unspecified
             Font: unspecified
          Fontset: unspecified
          Inherit: prog-mode-default
```
</details>

Given the mention of `aquamacs-autoface-mode`, I figured this might be something unique to Aquamacs that I'll need to work around here.

I tried customizing the `company-frontends` variable to whether changing my choices would avoid tripping over these non-list faces, but to no avail; I couldn't find any choices that restored the expected popup behavior.

Here is the `company-diag` output:

<details>
<summary>company-diag</summary>

```
Emacs 25.3.50.1 (x86_64-apple-darwin16.7.0) of 2019-08-18 on watson.local
Company 0.9.13

company-backends: (company-lsp company-bbdb company-nxml company-css company-semantic company-cmake company-clang company-files
             (company-dabbrev-code company-gtags company-etags company-keywords)
             company-oddmuse company-dabbrev)

Used backend: company-lsp
Major mode: go-mode
Prefix: (#("E" 0 1
   (fontified t))
 . t)

Completions: none
```
</details>

hlissner:
Fixes three cases:

1. When FACE is remapped to a single face in face-remapping-alist (which fixes #1022).
2. When FACE remapped to a list of faces (and potentially contains a self-reference, which face-remap allows so faces can refer to their pre-remapped selves).
3. When FACE is a face spec, a missing argument would've caused a wrong-number-of-args error.

Note: I've submitted my FSF CA (thanks for that by the way). Still waiting to hear back.

EDIT: Corrected typo in comment.

### DIFF
--- a/company.el
+++ b/company.el
@@ -2750,13 +2750,17 @@ If SHOW-VERSION is non-nil, show the version in the echo area."
   ;; Like `face-attribute', but accounts for faces that have been remapped to
   ;; another face, a list of faces, or a face spec.
   (cond ((symbolp face)
-         (let ((remap (cadr (assq face face-remapping-alist))))
+         (let ((remap (cdr (assq face face-remapping-alist))))
            (if remap
-               (company--face-attribute remap attr)
+               (company--face-attribute
+                ;; Faces can be remapped to their unremapped selves, but that
+                ;; would cause us infinite recursion.
+                (if (listp remap) (remq face remap) remap)
+                attr)
              (face-attribute face attr nil t))))
         ((keywordp (car-safe face))
          (or (plist-get face attr)
-             (company--face-attribute (plist-get face :inherit))))
+             (company--face-attribute (plist-get face :inherit) attr)))
         ((listp face)
          (cl-find-if #'stringp
                      (mapcar (lambda (f) (company--face-attribute f attr))


dgutov:
Thanks! That was very fast ;-)

> Note: I've submitted my FSF CA (thanks for that by the way). Still waiting to hear back.

That might take some time, but the important thing is that you have submitted it. And anyway, +/- 2 lines shouldn't be a problem.

PR #1024, opened 2020-10-14
Fix company--face-attribute for text-scaled faces

hlissner:
It's me again!

This fixes [another edge case](https://github.com/hlissner/doom-emacs/issues/4092): `company--face-attribute` throws an `invalid face` error when `default` is remapped to a face-spec without an `:inherit` property (e.g. `(:height 1.44)`). To reproduce this: `M-x text-scale-increase`, then invoke company.

### DIFF
--- a/company.el
+++ b/company.el
@@ -2803,7 +2803,8 @@ If SHOW-VERSION is non-nil, show the version in the echo area."
     (let* ((nl-face (list
                      :extend t
                      :inverse-video nil
-                     :background (company--face-attribute 'default :background)))
+                     :background (or (company--face-attribute 'default :background)
+                                     (face-attribute 'default :background nil t))))
            (str (apply #'concat
                        (when nl " \n")
                        (cl-mapcan
--- a/company.el
+++ b/company.el
@@ -2749,7 +2749,8 @@ If SHOW-VERSION is non-nil, show the version in the echo area."
 (defun company--face-attribute (face attr)
   ;; Like `face-attribute', but accounts for faces that have been remapped to
   ;; another face, a list of faces, or a face spec.
-  (cond ((symbolp face)
+  (cond ((null face) nil)
+        ((symbolp face)
          (let ((remap (cdr (assq face face-remapping-alist))))
            (if remap
                (company--face-attribute


dgutov:
Hi again!

Faces are tricky, huh?

The change looks simple, but what effect does it have? Whenever text-scale is present, `company--face-attribute` will now return nil. That negates the original purpose of that code (to fetch the explicit background color of the default face) and brings back https://debbugs.gnu.org/38563 (better try it in Emacs 27, not 28; the problem is easier to reproduce there).

So it looks like simply returning nil is not the way. Instead, when `(plist-get face :inherit)` returns nil, we could fall back to the `default` face, without any remappings. Does this work for you?

```diff
diff --git a/company.el b/company.el
index fb39fa8..10a4fb4 100644
--- a/company.el
+++ b/company.el
@@ -2760,7 +2760,10 @@ If SHOW-VERSION is non-nil, show the version in the echo area."
              (face-attribute face attr nil t))))
         ((keywordp (car-safe face))
          (or (plist-get face attr)
-             (company--face-attribute (plist-get face :inherit) attr)))
+             (let ((parent-face (plist-get face :inherit)))
+               (if parent-face
+                   (company--face-attribute parent-face attr)
+                 (face-attribute 'default attr)))))
         ((listp face)
          (cl-find-if #'stringp
                      (mapcar (lambda (f) (company--face-attribute f attr))
```

hlissner:
> Faces are tricky, huh?

They are, unfortunately!

> Whenever text-scale is present, company--face-attribute will now return nil. That negates the original purpose of that code (to fetch the explicit background color of the default face) and brings back https://debbugs.gnu.org/38563 (better try it in Emacs 27, not 28; the problem is easier to reproduce there).

Ah, I see what you mean. Unfortunately, your solution would introduce another problem: if `default` is remapped to a list of faces, say `((:height 1.44) go-mode-default)`, `company--face-attribute` is mapped to each element. It is vital invalid entries are substituted with `nil`, so we can carry on and test the next face. _However_, what if _all_ the entries are invalid? e.g. you have no remappings and only text-scale active, then `company--face-attribute` returns `nil`, as you've described.

I think I have simple solution for that (in `company--replacement-string`):

```diff
     (let* ((nl-face (list
                      :extend t
                      :inverse-video nil
-                     :background (company--face-attribute 'default :background)))
+                     :background (or (company--face-attribute 'default :background)
+                                     (face-attribute 'default :background nil t))))
            (str (apply #'concat
                        (when nl " \n")
                        (cl-mapcan
                        ;; https://debbugs.gnu.org/cgi/bugreport.cgi?bug=42552#23
                        (lambda (line) (list line (propertize "\n" 'face nl-face)))
```

If all else fails, we fall back to `'default`'s unremapped :background.

In my limited tests, this seems to work fine. Do you see any problem with it?

dgutov:
>  if default is remapped to a list of faces, say ((:height 1.44) go-mode-default), company--face-attribute would stop at the first invalid element and return the fallback value instead of moving on to test the next face in the list

Makes sense.

All right, let's go with your counter-proposal, and see if anything else comes up.

Thanks!

PR #1026, opened 2020-10-16
Add cdlatex backend

benneti:
A small backend to show possible completions from [cdlatex](https://github.com/cdominik/cdlatex).

### DIFF
--- a/company-cdlatex.el
+++ b/company-cdlatex.el
@@ -1,0 +1,64 @@
+;;; company-cdlatex.el --- company-mode completion backend for cdlatex
+
+;; Copyright (C) 2020  Free Software Foundation, Inc.
+
+;; Author: Benedikt Tissot
+
+;; This file is part of GNU Emacs.
+
+;; GNU Emacs is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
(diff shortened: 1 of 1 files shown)


dgutov:
Hi! 

I don't use Latex, so it would be non-trivial for me to maintain.

And this also depends on a third-party package, right?

Have you considered proposing a `completion-at-point-functions` element for `cdlatex-mode` instead? If said mode installs an appropriate completion function, it will automatically work with `company-mode`.

benneti:
Hi,
Yes, it depends on cdlatex.
I did not consider this (cdlatex does not provide a completion-at-point-function), but I will have a look at it.
Do you think it is feasable/worth it to translate the company backend into a completion-at-point-function and propose a pr to cdlatex?

dgutov:
> Do you think it is feasable/worth it to translate the company backend into a completion-at-point-function and propose a pr to cdlatex?

Yes, that's what I suggest you do.

> EDIT: after a quick look at the docs it seems like running a post-completion hook for completion-at-point-function is not as simple as it is in company or did I miss something?

It's a bit trickier (the whole `c-a-p-f` API requires more boilerplate), but as the result it should work with different completion systems, not just company-mode.

The counterpart to `post-completion` is `:exit-function`. See, for example, how it's used in `shell--command-completion-data`.

dgutov:
Let me know if you have any further questions.

PR #1032, opened 2020-10-27
company-set-selection: adding offset internally instead of via caller

### LINKED ISSUE #1030: company-tng breaks company-search-candidates
Steps to reproduce:

1. Install MELPA and then install company through MELPA
2. Start Emacs, go into `*scratch*` buffer
3. Turn on company-mode: <kbd>M-x</kbd> `company-mode` <kbd>RET</kbd>
4. Turn on company-tng-mode: <kbd>M-x</kbd> `company-tng-mode` <kbd>RET</kbd>
5. Type something with lots of completions, such as `find-` and wait for candidates to appear
6. Type <kbd>C-s</kbd> to start searching
7. Type something to search for, such as `lit` as in `find-file-literally`

Expected results: Candidate `find-file-literally` is selected

Actual results: The candidate above `find-file-literally` is selected (if you try to match the first candidate by searching, the first candidate is never selected)

Tested in Emacs 27.1 and a native-comp build from the prior week, same results in both.  macOS 10.14.6, NS port.  I tested using an empty init.el with company-mode from 5f5949b6ae8ea9df94c6cb3e01d9cae43623b794.  No other packages were installed other than company and its deps.

kiennq:
Fix #1030 
The selection is being offset for `company-tng` to work. Before we just offset it in `company-select-next`, that logic should be mved to `company-set-selection`.

### DIFF
--- a/company.el
+++ b/company.el
@@ -1203,6 +1203,7 @@ update if FORCE-UPDATE."
     (let* ((offset (if company-selection-default 0 1))
            (company-candidates-length
             (+ company-candidates-length offset)))
+      (setq selection (+ selection offset))
       (setq selection
             (if company-selection-wrap-around
                 (mod selection company-candidates-length)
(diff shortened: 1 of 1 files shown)


dgutov:
Thanks!

PR #1033, opened 2020-11-15
Implement `company-select-first' and `company-select-last'

edwargix:
I think it's worth having these two functions so users can bind `M-<` (which is usually `beginning-of-buffer`) to `company-select-first` and `M->` (which is usually `end-of-buffer`) to `company-select-last`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,7 @@
 
 ## Next
 
+* New commands `company-select-first` and `company-select-last`.
 * `company-tng-mode` has been added to replace both
   `company-tng-configure-default` and the manual method of enabling
   `company-tng-frontend` (see also `company-tng-auto-configure`). Also,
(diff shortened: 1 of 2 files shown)


dgutov:
Sure, if you like.

Could you also add a NEWS entry? At the top of that file.

edwargix:
> Sure, if you like.
> 
> Could you also add a NEWS entry? At the top of that file.

Done.  Let me know if I should re-word

NEWS.md
```diff
@@ -2,6 +2,8 @@
 
 ## Next
 
+* `company-select-first` selects the first completion candidate and
```
dgutov, line 5: Please move the descriptions of the commands in the docstrings.

The news entry can simply say that these new commands were added (see some older entries for examples).

edwargix: Done

dgutov:
Thanks!

PR #1045, opened 2020-12-30
`company-select-mouse': per-frontend interface

### LINKED ISSUE #1044: Per-frontend `company-select-mouse`
Currently, `company-select-mouse` and consequently `company-complete-mouse` do not work with frontends such as `company-box`, because the former requires a `company-pseudo-tooltip-overlay` to be present. That design precludes frontends similar to `company-pseudo-tooltip` from supporting mouse-based auto-completion.

I propose (and have written a POC for that which I will publish soon) that a new frontend command be added: `select-mouse`. Frontends could access the selection mouse event trough the `let`-bound global `company-mouse-event` variable and should update `company-selection` along with whatever else that needs to be updated. A frontend shall set `company-mouse-move-success` to t if the mouse event it received would select a valid candidate, in which case `company-pseudo-tooltip-overlay` succeeds as well.

What do you think?

nbfalcon:
`company-select-mouse' now delegates to `company-call-frontends` with
`select-mouse'. If at least one frontend reports success (returns t), the
candidate is considered selected. This allows other,
`company-pseudo-tooltip-overlay`-like frontends (e.g. `company-box`) to support
mouse-selection.

To implement this, refactor `company-call-frontends` to return t if at least one
frontend did the same. The mouse event is passed trough the
`company-mouse-event` global, which is `let'-bound in `company-select-mouse'.

Fixes #1044.

Also, I decided to go with a slightly different interface for `company-call-frontends`: I believe it is in general more useful to have a success/failure result from multiple frontends than a list of results, so I decided to return t if at least one frontend did the same. I can change that if you think that an list of frontend results is better.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -27,6 +27,7 @@
 
 ## 2020-07-26 (0.9.13)
 
+* `company-select-mouse` is now a frontend action.
 * `company-clang`: error handling is more permissive.
 * `company-tng` stops disabling `post-completion` in backends
   ([#946](https://github.com/company-mode/company-mode/pull/946)). Instead,
--- a/company.el
+++ b/company.el
@@ -1188,11 +1188,14 @@ can retrieve meta-data for them."
                  (string-match-p "\\`company-" (symbol-name this-command)))))))
 
 (defun company-call-frontends (command)
-  (dolist (frontend company-frontends)
-    (condition-case-unless-debug err
-        (funcall frontend command)
-      (error (error "Company: frontend %s error \"%s\" on command %s"
-                    frontend (error-message-string err) command)))))
+  (let (success)
+    (dolist (frontend company-frontends)
+      (condition-case-unless-debug err
+          (when (funcall frontend command)
+            (setq success t))
+        (error (error "Company: frontend %s error \"%s\" on command %s"
+                      frontend (error-message-string err) command))))
+    success))
 
 (defun company-set-selection (selection &optional force-update)
   "Set SELECTION for company candidates.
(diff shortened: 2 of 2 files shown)


nbfalcon:
Here is a Video demo with both `company-box-mode` and normal `company-pseudo-tooltip-overlay-frontend`:

https://user-images.githubusercontent.com/69264154/103365583-b2ddf900-4ac0-11eb-809c-f3af0e8f90ea.mp4

You'll have to excuse the bad quality - I use a tiling window manager.

dgutov:
> I believe it is in general more useful to have a success/failure result from multiple frontends than a list of results, so I decided to return t if at least one frontend did the same.

I don't think so: you can implement this behavior in terms of the other one, but not vice versa, so the other one is more powerful. So we'll probably switch to that one if at some point we'll need some more complex handing of return values.

We don't have to do that now, though. `company-call-frontends` isn't really supposed to be used by third-party code, so there's no requirement to keep its behavior stable.

dgutov:
Please add a NEWS entry at the top of that file (you can be brief), and we're good to go.

nbfalcon:
I have fixed both points - there is now a NEWS entry and `company-call-frontends` returns a list.

dgutov:
Thanks!

NEWS.md
```diff
@@ -27,6 +27,7 @@
 
 ## 2020-07-26 (0.9.13)
 
+* `company-select-mouse` is now a frontend action.
```
dgutov, line 30: That line was a bit misplaced, though.

nbfalcon: Where should it go instead? I can make a PR to correct this.

dgutov: Already fixed, see the latest `master`.

PR #1047, opened 2021-01-03
company--continue: also check for company-abort-on-unique-match

kiennq:
We should also check for `company-abort-on-unique-match` in `company--continue`


dgutov:
Thanks!

Issue #1053, opened 2021-01-15
company-gtags backend misbehaving with c preprocessor macros


aagon:
GNU Global can tag C preprocessor macros, either in the form of macro-variables ```#define PAGE_SIZE 4096```, or in the form of macro-functions ```#define PAGE_BELOW(a) ((a / 4096) * 4096)```.

In the second case, we would like the completion to offer ```PAGE_BELOW(a)``` on a prefix like ```PAGE_B```, and not ```PAGE_BELOW(a) ((a / 4096) * 4096)```.

But company-gtags offers the latter, which is not what we want.

This is due to a specific regexp in the function ```company-gtags--annotation``` being present in its greedy version :
```elisp
(defun company-gtags--annotation (arg)
  (let ((meta (get-text-property 0 'meta arg)))
    (when (string-match (concat (regexp-quote arg) "\\((.*)\\).*") meta)
      (match-string 1 meta))))
```
Replacing it with its lazy version solves the problem :
```elisp
(defun company-gtags--annotation (arg)
  (let ((meta (get-text-property 0 'meta arg)))
    (when (string-match (concat (regexp-quote arg) "\\((.*?)\\).*") meta)
      (match-string 1 meta))))
```
This solution keeps intact the completions for regular functions (provided they only have one set of parentheses in their prototype, which is always the case in C/C++). I'll let you decide whether this change is not too violent (it may break other languages supported by GNU Global I am not familiar with). In the meantime, I'll prepare the trivial PR. The changes are below the 15 lines, I should not have to sign anything.

Best,

Aymeric Agon-Rambosson

aagon:
PR #1054.

aagon:
Solved by #1054, closing.

PR #1054, opened 2021-01-15
Refine company-gtags--annotation regexp

aagon:
Replacing company-gtags--annotation greedy regexp with the lazy version to make company-gtags behave better with C preprocessor macro-functions.

### DIFF
--- a/test/gtags-tests.el
+++ b/test/gtags-tests.el
@@ -1,0 +1,55 @@
+;;; gtags-tests.el --- company-mode tests  -*- lexical-binding: t -*-
+
+;; Copyright (C) 2020  Free Software Foundation, Inc.
+
+;; Author: Aymeric Agon-Rambosson
+
+;; This file is part of GNU Emacs.
+
+;; GNU Emacs is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; GNU Emacs is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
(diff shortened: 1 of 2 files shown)


aagon:
Issue #1053.

dgutov:
Thank you.

If you could also add a couple of tests, it would be great. This will be a new file, `test/gtags-tests.el`, but it should be easy to follow the example of `test/clang-tests.el`. That's optional, though.

Also, could you try this regexp instead: `"\\(([^\)]*)\\).*"`? It should be marginally faster.

aagon:
The regexp you proposed seems to work the same than mine, I have used it instead. I don't know which is faster though.

I added a set of tests like you asked. I could only test my use cases, so only C :
- C preprocessor macro variables
- C preprocessor macro functions
- C functions
- C types
- C structs

According to the documentation, GNU Global is compatible with some other languages (Yacc, PHP, Assembly (which ??) and Java). I cannot guarantee that this change doesn't break anything for these other languages.

Now that the commit has gotten bigger, I think I will have to sign stuff over to the FSF. I'll look at the website on how to do that, but if you can point me in the right direction, that will be most appreciated. I am based in France, and I am doing this on my own time.

test/gtags-tests.el
```diff
@@ -0,0 +1,55 @@
+;;; gtags-tests.el --- company-mode tests  -*- lexical-binding: t -*-
+
+;; Copyright (C) 2020  Free Software Foundation, Inc.
```
dgutov, line 3: It's 2021 now ;-)

dgutov:
This is excellent, thank you.

I'll merge the changes right away, but the copyright assignment instructions are here: https://code.orgmode.org/bzg/org-mode/raw/master/request-assign-future.txt (keep the program name as "Emacs").

aagon:
Thank you for your help, your time, and useful reviews.

I have sent the mail to request the form.

dgutov:
Thanks! Let me know when the process completes, please.

aagon:
The copyright assignment was received and sent back signed. I'll let you know when the FSF sends an acknowledgment.

PR #1063, opened 2021-02-11
Cache ispell results

iwahbe:
Add a generic LRU cache for company mode, and apply it to the `company-ispell` back end. This allows me to run the ispell backend at `company-idle-delay 0.0` and `company-minimum-prefix-length 1` without any noticeable lag. 

Without the cache I get stutters when I start a word, especially with a common letter like `a` or `t`. After some use, the cache holds `ispell-lookup-words` results for the most commonly used letter combinations. 

This feature is turned off by default. This allows people to opt in to the increased memory usage of the cache.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -28,6 +28,12 @@
   users these days, with
   ([emacs-eclim](https://github.com/emacs-eclim/emacs-eclim/)) declared obsolete
   in favor of `lsp-java`. Though it used its own backend anyway.
+* `company-ispell` now has a built-in cache mechanism. It can be turned on by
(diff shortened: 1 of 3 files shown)


dgutov:
Hi! Sorry about the very late response. I didn't have the mental bandwidth at the time.

PR #1411 implements a simple caching mechanism, but with more explicit approach to invalidation (for `company-ispell`, it currently checks the absolute file name of the dictionary used; we could also add mtime later, perhaps).

Could you try out that branch and see whether there is a significant difference between it and your proposal? According to my measurements, the latency should be below the general human threshold, but it also depends on the dictionary used and its size, of course.

PR #1070, opened 2021-03-06
Initial support for icons in company-mode core

yyoncho:
- added new defcustom `company-format-margin-function` which can be used to plug
icon provider function

- added icons from vscode dark theme and corresponding function

- extended company's protocol with `candidate-kind` backend call.


yyoncho:
and the gif: 
icons

dgutov:
Hi Ivan,

Thank you for this PR.

I'll have to get back to reviewing it properly later, but off the top of my head:

* We probably want to name the backend action just `kind`. No other action name has `candidate-` in it.
* Needs a NEWS entry.
* Is there a "vscode light" icon set? Is it different?

yyoncho:
I added a light theme and I changed candidate-kind to kind. There is one limitation that I saw while testing - if you have format function and the backend does not support icons then company-tooltip-margin should be set to 2 otherwise first letters are cut when you are completing at the beginning of the line. But generally, it will be better for the end-user to set company-format-margin-function  to nil for backends that do not provide `kind`.

dgutov:
> I added a light theme and I changed candidate-kind to kind.

Thank you.

Should the default behavior make an effort to auto-dispatch to the appropriate set depending on the value of `frame-background-mode`? Is it generally expected that the "light" set fits the light themes and vice versa?

> There is one limitation that I saw while testing - if you have format function and the backend does not support icons then company-tooltip-margin should be set to 2 otherwise first letters are cut when you are completing at the beginning of the line. But generally, it will be better for the end-user to set company-format-margin-function to nil for backends that do not provide kind.

Hmm. Could the format function return some appropriate value in this case? The use cases I expect are both backends that only provide icons for some of their completions, as well as backend groupings where one backend has icons are another doesn't. The latter case could be relevant to `lsp-mode` too.

I was also wondering about the "general" nature of the choice of the `company-format-margin-function` name. It implies that it can be used to format the margin with something else than icons, perhaps something that isn't even based on `kind`. Was that the idea?

Relatedly, perhaps `company-format-margin-function` should be passed the value of completion itself, so that is calls `(company-call-backend 'kind value)` on its own? Then other format-margin-function-s (whatever they might be) could be based on something other than the value of `kind`. WDYT?

yyoncho:
> Should the default behavior make an effort to auto-dispatch to the appropriate set depending on the value of `frame-background-mode`? Is it generally expected that the "light" set fits the light themes and vice versa?

Sounds like a good improvement but it seems to me that this can be implemented after the initial submission. 

> Hmm. Could the format function return some appropriate value in this case? The use cases I expect are both backends that only provide icons for some of their completions, as well as backend groupings where one backend has icons are another doesn't. The latter case could be relevant to lsp-mode too.

I think it does, but I was unable to find the place that causes this behavior. I will debug it a bit more. 

> I was also wondering about the "general" nature of the choice of the company-format-margin-function name. It implies that it can be used to format the margin with something else than icons, perhaps something that isn't even based on kind. Was that the idea?

Technically, you can use ASCII symbols as well, not only icons. If you prefer a different name I am fine to change it. 

> Relatedly, perhaps company-format-margin-function should be passed the value of completion itself, so that is calls (company-call-backend 'kind value) on its own? Then other format-margin-function-s (whatever they might be) could be based on something other than the value of kind. WDYT?

Sounds good, I will update the PR.

yyoncho:
> I think it does, but I was unable to find the place that causes this behavior. I will debug it a bit more.

Found the issue, I should call the format function no matter if the kind is nil.

yyoncho:
Pushed fixes related to the location get `kind` call and also fixed the issue about cutting the popup for backends that do not return kind.

dgutov:
> Sounds like a good improvement but it seems to me that this can be implemented after the initial submission.

It's not a blocker, but it could use your input, now or after the merge. One argument to do this sooner rather than later is that it can result in the change of `company-format-margin-function`'s default (adding an `-auto-` variant).

> Technically, you can use ASCII symbols as well, not only icons. If you prefer a different name I am fine to change it.

My only concern is it might be *too* general. But the implementation matches it now, so it's not a real problem.

> Found the issue, I should call the format function no matter if the kind is nil.

Thank you.

Now, I think you can add that NEWS entry.

Also, could you add at least one test for the popup rendering with icons? So that that code path is exercised on the CI. Somewhere around here maybe: https://github.com/company-mode/company-mode/blob/master/test/frontends-tests.el#L257

Also see the latest inline comments.

company.el
```diff
@@ -64,6 +64,7 @@
 (require 'cl-lib)
 (require 'newcomment)
 (require 'pcase)
+(require 'find-func)
```
dgutov, line 67: Why `find-func`?

Should it be `subr-x` instead? Which your addition of the use of `if-let*` seems to require.

This will also require a dependency bump to at least 25.1. There is still no `if-let*` in `subr-x` in 25.1, but IIUC `if-let` (no asterisk) behaves identically.

dgutov:
And we also need a description for `kind` in the `company-backends` docstring.

company.el
yyoncho: It is for `find-library-name`.

I will change if-let* to if-let

dgutov:
And some licensing information for the icons. I think dropping https://github.com/microsoft/vscode-icons/blob/main/LICENSE into the `/icons` directory is a must, at least.

dgutov:
Have you tried using the SVG icons, btw? The png ones look rather blurry on my 4K screen.

If we're going to disable this feature by default, we could as well mandate SVG support in Emacs for using it. As long as rendering quality and speed don't suffer.

yyoncho:
> Have you tried using the SVG icons, btw? The png ones look rather blurry on my 4K screen.

I had trouble with svg support on my system the last time I tried. 

> If we're going to disable this feature by default, we could as well mandate SVG support in Emacs for using it.

IMHO it should be off by default due to TUI users.

yyoncho:
> Also, could you add at least one test for the popup rendering with icons? So that that code path is exercised on the CI. Somewhere around here maybe: https://github.com/company-mode/company-mode/blob/master/test/frontends-tests.el#L257

can you suggest which test to use for reference?

yyoncho:
I pushed all(?) of the requested changes except the test.

dgutov:
> I had trouble with svg support on my system the last time I tried

OK, then let's keep it as-is for now. Seems to work fine here, after I copied the necessary files, changed `company-vscode-icons-mapping` and the `:type` value to `svg` inside `company--icons-margin-function`.

On a related note, I wonder if you tried to use the Font Awesome icons instead. No problems with svg, though I guess they are still not going to work for terminal users.

yyoncho:
> On a related note, I wonder if you tried to use the Font Awesome icons instead. No problems with svg, though I guess they are still not going to work for terminal users.

I think that I haven't. I have created eclipse/idea/netbeans themes for company-box though - https://github.com/sebastiencs/company-box/pull/98

company.el
dgutov: > It is for find-library-name

Ah, ok.

You can do the same thing a bit faster without it. See: https://github.com/company-mode/company-mode/blob/master/company-tests.el#L24-L25

dgutov:
> can you suggest which test to use for reference?

Taking examples from tests above and below, you can follow the same structure as `company-create-lines-with-multiple-width-and-keep-prefix` does, but in the end call `ert-equal-including-properties`.

Writing the resulting strings out together with their properties is admittedly verbose, but I don't have any better ideas currently.

dgutov:
> I have created eclipse/idea/netbeans themes for company-box though - sebastiencs/company-box#98

That looks neat. Wish we could standardize on a scalable format, though.

Do I see you include the `@2x` scaled versions there, though? That could work for my system as well.

NEWS.md
```diff
@@ -2,6 +2,10 @@
 
 ## Next
 
+* Added support for icons
+  ([#1070](https://github.com/company-mode/company-mode/pull/1070)).
+  `company-format-margin-function` has been added
+  Added new backend command `kind`
```
dgutov, line 8: Please end each sentence with a period.

yyoncho: done.

dgutov:
To clarify: the icon scaling issue is not too urgent, but I'd like to see a test or two before merging.

yyoncho:
> To clarify: the icon scaling issue is not too urgent, but I'd like to see a test or two before merging.

I haven't forgotten about that - it is in my queue.

yyoncho:
added test. The tests that are including the real icons related formatting functions are failing with "Window system frame should be used"

dgutov:
Do you mean they are failing when run in batch mode?

That's the purpose of the `interactive` tag: to mark tests that only succeed when run in a graphical session. Either with `M-x ert` manually, or with `make test`.

yyoncho:
> Do you mean they are failing when run in batch mode?

They fail with `make test`.

dgutov:
Ah, ok. Because it runs non-graphical Emacs. But they should still work when run manually with `M-x ert`.

If you are not sure about committing it yourself, and how to do it best, could you post these tests here?

yyoncho:
```
(ert-deftest company-create-lines-with-format-function ()
  :tags '(interactive)
  (let* (company-show-numbers
         (company-candidates '("ArrayList"))
         (company-candidates-length 1)
         (company-package-root default-directory)
         (company-format-margin-function 'company-vscode-light-icons-margin-function)
         (company-tooltip-maximum-width 60)
         (company-tooltip-minimum-width 60)
         (company-backend (lambda (c &rest _) (pcase c (`kind 'class)))))
    (should (ert-equal-including-properties
             (car (company--create-lines 0 999))
             (let ((text #("  ArrayList                                                   "
                           0 1
                           (face
                            (company-tooltip-selection company-tooltip)
                            mouse-face
                            (company-tooltip-mouse))
                           1 2
                           (face
                            (company-tooltip-selection company-tooltip)
                            mouse-face
                            (company-tooltip-mouse)
                            display
                            (space :width 0.125))
                           2 62
                           (face
                            (company-tooltip-selection company-tooltip)
                            mouse-face
                            (company-tooltip-mouse)))))
               (add-text-properties
                0 1
                `(display
                  (image :file ,(expand-file-name "vscode-light/symbol-class.png")
                         :type png :width 15 :height 15 :ascent center :background "#c8c6dd"))
                text)
               text)))))
```

dgutov:
`ert-equal-including-properties` didn't exactly work out because the `:width` value of the spaces depends on font metrics or whatever. But I think you'll like the end result more.

dgutov:
Here's a patch to play with if someone wants to test how this works with Elisp completion:

```diff
diff --git a/lisp/progmodes/elisp-mode.el b/lisp/progmodes/elisp-mode.el
index 312153052d..c791f52e29 100644
--- a/lisp/progmodes/elisp-mode.el
+++ b/lisp/progmodes/elisp-mode.el
@@ -557,6 +557,7 @@ elisp-completion-at-point
                     ((elisp--expect-function-p beg)
                      (list nil obarray
                            :predicate #'fboundp
+                           :company-kind (lambda (_) 'function)
                            :company-doc-buffer #'elisp--company-doc-buffer
                            :company-docsig #'elisp--company-doc-string
                            :company-location #'elisp--company-location))
@@ -570,6 +571,7 @@ elisp-completion-at-point
                                             (symbol-plist sym)))
                            :annotation-function
                            (lambda (str) (if (fboundp (intern-soft str)) " <f>"))
+                           :company-kind #'elisp--company-kind
                            :company-doc-buffer #'elisp--company-doc-buffer
                            :company-docsig #'elisp--company-doc-string
                            :company-location #'elisp--company-location))
@@ -580,6 +582,7 @@ elisp-completion-at-point
                                                  obarray
                                                  #'boundp
                                                  'strict))
+                           :company-kind (lambda (_) 'variable)
                            :company-doc-buffer #'elisp--company-doc-buffer
                            :company-docsig #'elisp--company-doc-string
                            :company-location #'elisp--company-location)))
@@ -626,11 +629,13 @@ elisp-completion-at-point
                                       (looking-at "\\_<let\\*?\\_>"))))
                         (list t obarray
                               :predicate #'boundp
+                              :company-kind (lambda (_) 'variable)
                               :company-doc-buffer #'elisp--company-doc-buffer
                               :company-docsig #'elisp--company-doc-string
                               :company-location #'elisp--company-location))
                        (_ (list nil obarray
                                 :predicate #'fboundp
+                                :company-kind (lambda (_) 'function)
                                 :company-doc-buffer #'elisp--company-doc-buffer
                                 :company-docsig #'elisp--company-doc-string
                                 :company-location #'elisp--company-location
@@ -646,6 +651,14 @@ elisp-completion-at-point
                                        " " (cadr table-etc)))
                     (cddr table-etc)))))))))
 
+(defun elisp--company-kind (str)
+  (let ((sym (intern-soft str)))
+    (cond
+     ((boundp sym) 'variable)
+     ((fboundp sym) 'function)
+     ((featurep sym) 'module)
+     (t 'misc))))
+
 (defun lisp-completion-at-point (&optional _predicate)
   (declare (obsolete elisp-completion-at-point "25.1"))
   (elisp-completion-at-point))
```

dgutov:
Thank you for the patience, I'm merging this now.

Further improvements welcome. Some current issues:

* Completion when prefix starts at column 1 makes the popup and the whole line contents behind it move 1 column to the left. The hacky patch below seems to fix it, but a cleaner solution is needed. Note that it might require a change in the newly added function-var's signature.

```diff
diff --git a/company.el b/company.el
index 149e15b..cd90166 100644
--- a/company.el
+++ b/company.el
@@ -1448,7 +1448,7 @@ end of the match."
                                        bkg))))
         (concat
          (propertize " " 'display spec)
-         (propertize " " 'display `(space . (:width ,(- 2 (car (image-size spec))))))))
+         (propertize " " 'display `(space . (:align-to ,(- (current-column) (length company-prefix)))))))
     "  "))
 
 (defun company-vscode-dark-icons-margin-function (candidate selected)
```

* When a backend returns no kinds, we show double-char margin. I wonder if we can show single-char margin if all the currently displayed popup lines have no kind. I wonder how will the scrolling look with this approach when some items with non-empty kind come into view, then disappear from it.

* Icon scaling and SVG. You said you have been having problems with this format. Have you talked with other lsp-mode users/contributors about it? Any idea how common such problems are? There have been talks on emacs-devel about using it for more features, so if there are popular configurations where it's unusable, the earlier we'll know, the better.

yyoncho:
Are you going to do PRs to fix melpa/elpa(?) recipes to include the icons dir?

> Icon scaling and SVG. You said you have been having problems with this format. Have you talked with other lsp-mode users/contributors about it? Any idea how common such problems are? There have been talks on emacs-devel about using it for more features, so if there are popular configurations where it's unusable, the earlier we'll know, the better.

My issues are build-related. I guess the simplest solution is to include svg images, check if Emacs supports svg, and use svg.

dgutov:
ELPA is OK: https://elpa.gnu.org/devel/company.html, and thanks for taking care of MELPA.

> My issues are build-related. I guess the simplest solution is to include svg images, check if Emacs supports svg, and use svg.

Are these difficult to fix?

If we're able to standardize on SVG-only, it would be easy to add automatic detection of the scaling factor, for instance (in `company-icon-size`'s default value form), without additional guards there. And there should be beneficial secondary effects to the ecosystem as well.

yyoncho:
> Are these difficult to fix?

I rebuilt emacs 27.2 and now I am fine. 

> If we're able to standardize on SVG-only,

I don't have a strong preference here. IMHO we should consider whether the common emacs package distributions are built with svg support.

dgutov:
> I rebuilt emacs 27.2 and now I am fine.

Excellent.

> I don't have a strong preference here. IMHO we should consider whether the common emacs package distributions are built with svg support.

Seems like the Debian package does have SVG support: https://packages.debian.org/source/sid/emacs

https://ftp.gnu.org/gnu/emacs/windows/emacs-27/emacs-27.1-x86_64.zip also includes librsvg.

Let's go with SVG-only for now (we can do that for an optional feature) until we get more evidence that some major platform/distribution does not support it. Or of some other unexpected downsides.

dgutov:
@yyoncho One more question: have you ever seen the `key` kind used for a purpose that is somehow related to its name? IIUC, its icon is the same as what VS Code shows for "words in open files" completions, and https://code.visualstudio.com/docs/editor/intellisense#_types-of-completions lists its as `text`. Should we use that name instead?

I'm thinking of backends like dabbrev[-code] or ispell.

yyoncho:
Yes, it should be `text`, not string. For reference, here are defined the completion types https://microsoft.github.io/language-server-protocol/specification (look for CompletionItemKind). I have used lsp-treemacs lookups as a base, seems like we have it wrong in that place.

dgutov:
> Yes, it should be text, not string

"Text", not "key", right?

> For reference, here are defined the completion types https://microsoft.github.io/language-server-protocol/specification (look for CompletionItemKind)

Thanks, I was also looking in the wrong place there (SymbolKind).

dgutov:
@yyoncho BTW, what is a "unit"? Is it used in practice?

yyoncho:
I haven't seen it. But I am not using a lot of language servers.

dgutov:
OK, something to get back to later, maybe.

Could you help me with picking right kinds for Elisp, too?

We have variables, functions, modules, those are simple. Faces can be tagged as `color`. What about the rest of symbols (from obarray) that are offered in completions? For now I have tagged them as `misc`. But maybe `constant`, `value` or `reference` would be a better choice? WDYT?

PR #1076, opened 2021-03-27
Add svg support

yyoncho:


company.el
```diff
@@ -1394,32 +1394,78 @@ end of the match."
           (const :tag "Prefer closest in any direction"
                  company-occurrence-prefer-any-closest)))
 
+(defvar company-vscode-icons-type
+  (if (image-type-available-p 'svg)
+      'svg
+    'png))
+
+(defvar company-vscode-icons-mapping
+  (mapcar (lambda (mapping)
```
dgutov, line 1403: If we end up supporting both formats, might as well store only names without extensions in this list, and then append the extension inside the `expand-file-name` form.

yyoncho: That can work too. I wanted to remove the runtime cost.

dgutov: `(benchmark 20 '(concat "symbol-whatever." "svg"))` shows the difference would be blip on the radar.

`expand-file-name` is much slower in comparison, BTW. Though for 20 items its perf hit is also tiny.

PR #1077, opened 2021-03-27
Get duplicates from the source

yyoncho:
- This PR is changing the current behaviour, I can change the code to test with
plist-member to preserve the current hardcoded value.

### DIFF
--- a/company-capf.el
+++ b/company-capf.el
@@ -132,7 +132,9 @@ so we can't just use the preceding variable instead.")
                     (push (cons match-start pos) chunks)
                     (setq match-start nil))))
            (nreverse chunks)))))
-    (`duplicates t)
+    (`duplicates
+     (plist-get (nthcdr 4 company-capf--current-completion-data)
+                :company-duplicates))
     (`no-cache t)   ;Not much can be done here, as long as we handle
                     ;non-prefix matches.
     (`meta


dgutov:
Please clarify: you want this why? Improved performance?

> I can change the code to test with plist-member to preserve the current hardcoded value

That would be better, yes.

yyoncho:
> Please clarify: you want this why? Improved performance?

in lsp-mode you could have 2 items with the same name and with a different kind(e. g. `class` the keyword and `class` the snippet). So one of them is removed when you turn off annotations(lsp-mode has a setting for that). Also, the items can be distinguished using the quickhelp even with icons turned off.

dgutov:
I see.

In that case, we could also add disambiguation based on `kind` to `company--strip-duplicates` (in addition to `annotation` comparison).

dgutov:
...which is now done in f06c4e2.

@yyoncho Please let me know if this change is still needed, for performance or some other reasons.

yyoncho:
I think the change is good enough.

yyoncho:
When doing some performance testing, this popped up in my perf report. 

```
     - company--strip-duplicates                                   74  16%
      - company-call-backend                                       67  15%
       - company--force-sync                                       62  14%
        - apply                                                    59  13%
         - company-call-backend-raw                                52  11%
          - apply                                                  52  11%
           - company-capf                                          44   9%
            + company-capf--annotation                             29   6%
            + lsp-completion--candidate-kind                       11   2%
```

dgutov:
Hm, okay. Does this change really help?

Here's a form you can evaluate to measure the whole cycle, from request to display:

```el
(benchmark 1 '(company-begin-backend 'company-capf))
```

PR #1083, opened 2021-03-31
(margin): Add tty compatible margin function

mohkale:
See [#756](https://github.com/company-mode/company-mode/issues/756#issuecomment-810306170).

I wasn't sure which icons should be the default... so I just picked some random greek unicode glyphs. You can probably pick more appropriate ones.

### DIFF
--- a/company.el
+++ b/company.el
@@ -1448,6 +1448,45 @@ end of the match."
                                 candidate
                                 selected))
 
+(defvar company-unicode-icons-mapping
+  '((array . "Α")
+    (boolean . "β")
+    (class . "γ")
+    (color . "Δ")
+    (constant . "ε")
+    (enum-member . "ζ")
+    (enum . "Ζ")
+    (event . "η")
+    (field . "θ")
+    (file . "Ɩ")
+    (folder . "⍳")
+    (interface . "ϰ")
+    (key . "μ")
+    (keyword . "ν")
+    (method . "λ")
+    (function . "ƒ")
+    (misc . "ξ")
+    (module . "Ο")
+    (numeric . "π")
+    (operator . "⊙")
+    (parameter . "ρ")
+    (property . "σ")
+    (ruler . "τ")
+    (snippet . "υ")
+    (string . "φ")
+    (struct . "Χ")
+    (variable . "ѱ")))
+
+(defun company-unicode-icons-margin (candidate selected)
+  "Margin function which returns icons from vscode's light theme."
+  (when-let ((candidate candidate)
+             (kind (company-call-backend 'kind candidate))
+             (icon (alist-get kind company-unicode-icons-mapping)))
+    (propertize icon
+                'face (if selected
+                          'company-tooltip-selection
+                        'company-tooltip))))
+
 (defcustom company-format-margin-function nil
   "Function to format the margin.
 It accepts 2 params `candidate' and `selected' and can be used for
--- a/company.el
+++ b/company.el
@@ -1448,7 +1448,7 @@ end of the match."
                                 candidate
                                 selected))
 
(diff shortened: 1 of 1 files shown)


mohkale:
I didn't include any padding or spacing in the margins. If an icon is chosen it's prepended directly before the margin function.

Screenshot_20210331_044611

I'd suggest if someone wants spacing they should modify `company-unicode-icons-mapping`, such as:

```lisp
(setq company-unicode-icons-mapping
      (cl-loop for (sym . icon) in company-unicode-icons-mapping
               collect (cons sym (concat icon " "))))
```

mohkale:
@dgutov 

You also mentioned adding a margin function that can switch between unicode and image depending on `(display-graphic-p)`, not sure whether we should use vsxode-dark or vscode-light?

company.el
```diff
@@ -1448,6 +1448,55 @@ end of the match."
...
+    (propertize icon
+                'face (if selected
+                          'company-tooltip-selection
+                        'company-tooltip))))
+
+(defun company-unicode-or-vscode-icons-margin (candidate selected)
+  "Margin function which picks from vscodes icons or unicode icons
+based on `display-graphic-p'."
+  (if (display-graphic-p)
+      ;; Default to dark because who in their right mind uses light 😜
```
dgutov, line 1494: Just the users who care about ergonomics :stuck_out_tongue:

dgutov: Yeah, I think there are only 2 possible values, and it doesn't seem to ever return `nil`.

The dark icons might be the more reasonable fallback in the unlikely case it ever does because they're readable on while background. Not sure if the other set is as readable on black background.

company.el
```diff
@@ -1448,6 +1448,55 @@ end of the match."
                                 candidate
                                 selected))
 
+(defvar company-unicode-icons-mapping
```
dgutov, line 1451: Maybe call it `company-text-icons-mapping`? Greek chars are just one choice.

If someone installs FontAwesome, I think they'll be able to use it here by customizing this mapping (right?), even though that font's "characters" don't exactly correspond to Unicode tables.

company.el
mohkale: Done.

company.el
```diff
@@ -1448,6 +1448,55 @@ end of the match."
...
+  "Margin function which returns unicode icons."
+  (when-let ((candidate candidate)
+             (kind (company-call-backend 'kind candidate))
+             (icon (alist-get kind company-unicode-icons-mapping)))
+    (propertize icon
+                'face (if selected
+                          'company-tooltip-selection
+                        'company-tooltip))))
+
+(defun company-unicode-or-vscode-icons-margin (candidate selected)
```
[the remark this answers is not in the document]
mohkale: Done.

company.el
mohkale: Done.

company.el
```diff
@@ -1448,6 +1448,55 @@ end of the match."
...
+    (string . "φ")
+    (struct . "Χ")
+    (variable . "ѱ")))
+
+(defun company-text-icons-margin (candidate selected)
+  "Margin function which returns unicode icons."
+  (when-let ((candidate candidate)
+             (kind (company-call-backend 'kind candidate))
+             (icon (alist-get kind company-text-icons-mapping)))
+    (propertize icon
```
dgutov, line 1485: What happens if you remove this `propertize` call and just return `icon`?

IIUC, this face should be applied later on by popup rendering code.

mohkale: Doesn't seem to change anything. I've removed it now.

dgutov:
Thanks!

Time to play with greek characters, I guess. :-)

PR #1085, opened 2021-04-01
Add choices, tty, detect

jcs090218:
This patch does the following.

- Turn variable `company-text-icons-mapping` to custom
- Add `text` and `detect` choice to custom variable `company-format-margin-function`

### DIFF
--- a/company.el
+++ b/company.el
@@ -1451,7 +1451,7 @@ end of the match."
                                 candidate
                                 selected))
 
-(defvar company-text-icons-mapping
+(defcustom company-text-icons-mapping
   '((array . "Α")
     (boolean . "β")
     (class . "γ")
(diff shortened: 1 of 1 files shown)


company.el
```diff
@@ -1478,7 +1478,10 @@ end of the match."
     (snippet . "υ")
     (string . "φ")
     (struct . "Χ")
-    (variable . "ѱ")))
+    (variable . "ѱ"))
+  "List of text icons."
```
dgutov, line 1482: Alist, maybe? Or "mapping", or somesuch.

company.el
```diff
@@ -1478,7 +1478,10 @@ end of the match."
...
+  :type 'list
+  :group 'company)
```
dgutov, line 1484: We don't add the `:group` parameter when it duplicates the automatically detected value, please just remove it.

jcs090218: Removed.

company.el
jcs090218: How about mapping?

dgutov: Sure. Thanks!

PR #1092, opened 2021-04-15
Make company-complete-company works fine the same as v0.9.13 version.

XiangQingW:
In the latest version, `(define-key company-active-map (kbd "<return>") #'company-complete-selection)` doesn't work fine.

### DIFF
--- a/company.el
+++ b/company.el
@@ -2370,6 +2370,8 @@ For use in the `select-mouse' frontend action.  `let'-bound.")
 (defun company-complete-selection ()
   "Insert the selected candidate."
   (interactive)
+  (unless company-selection
+    (setq company-selection 0))
   (when (and (company-manual-begin) company-selection)
     (let ((result (nth company-selection company-candidates)))
       (company-finish result))))


dgutov:
...because you set `company-selection-default` to some non-default value?

Please explain the problem fully.

XiangQingW:
@dgutov Yes, `company-selection-default` is nil. Howerver, I have tried setting `company-selection-default` to 0, but after I did that, the automatic completion at the point is annoying. Are there other configs to stop automatic completion at the point? Thanks.

The screen recording as follows:
https://user-images.githubusercontent.com/7948384/114883910-e8bca880-9e37-11eb-91d3-2b98d0cd8ea0.mov

dgutov:
Do you have `company-tng-mode` enabled as well?

XiangQingW:
Yeah, `company-tng-mode` is enabled by default. When I disable `company-tng-mode` and config as follows:
```
(company-tng-mode -1)
(setq lsp-enable-snippet t)
```
everything works fine for me. Thanks for the hint.

dgutov:
> Yeah, company-tng-mode is enabled by default.

It isn't. But suppose your distribution does that. Doom?

> everything works fine for me. Thanks for the hint.

No problem. But I take it you were saying that `company-complete-company` worked differently before, even with `company-tng-mode` enabled?

XiangQingW:
After I inspected the backtrace of invoking `company-tng-mode`, `evil-collection-company-setup` startups this mode :(

image

When `company-tng-mode` is enabled and  `M-x company-complete-selection` is invoked to complete a typing, nothing happened. I have added some logs to trace, it's caused by the nil value of `company-selection`. But I don't know how to fix it correctly.

On the contrary, When `company-tng-mode` is disabled, `company-complete-selection` work fine the same as before.

dgutov:
> evil-collection-company-setup startups this mode :(

How positively Evil of them. :-)

> when company-tng-mode is enabled and M-x company-complete-selection is invoked to complete a typing, nothing happened

What I'm asking is, whether is worked differently in some previous version of `company-mode`, with TNG still enabled.

The idea is that the selection is indeed nil (absent) by default, so unless you have pressed TAB once or more, no completion is selected, and RET should do nothing (and insert newline instead). That's the TNG workflow.

XiangQingW:
I'll read the document thoroughly about TNG mode, then decide whether or not to enable it. Thank you.

PR #1095, opened 2021-04-21
Avoid positional arguments to define-minor-mode

tarsius:
Back in Emacs-21.1, `define-minor-mode` grew keyword arguments to
replace its old positional arguments.  Starting with Emacs-28.1
warning will be omitted if positional arguments are still used.

`:global nil` and `:keymap MODE-map` are not necessary because
those are the defaults.


dgutov:
Thanks!

Issue #1086, opened 2021-04-01
Completion popup randomly reset


rrudakov:
Hello. 

After recent updates I've started facing the following issue:
Sometimes when completion candidates popup is visible and I already selected some item it just disappears (like if I press `C-g`, but I didn't press anything). I couldn't find 100% way to reproduce this issue, `toggle-debug-on-error` won't show anything when it happens.

How does it looks like:

https://user-images.githubusercontent.com/6093590/113273717-37692f00-92dd-11eb-9459-29dc98a05670.mp4

### My configuration:
```emacs-lisp
(use-package company
  :config
  (setq company-global-modes '(not org-mode markdown-mode shell-mode))
  (setq company-tooltip-align-annotations t
        company-idle-delay .2
        company-minimum-prefix-length 3)

  (global-company-mode)
  (company-tng-mode))
```
This problem appears at least in `CIDER`, `lsp-mode` and `emacs-lisp-mode`.

### Emacs version:
`GNU Emacs 27.2.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.20, cairo version 1.16.0) of 2021-03-28`

I think I need some instructions how to debug this problem to provide some more details. Thank you!

dgutov:
Hi! This sounds suspiciously similar to #1084.

If you can, create a step-by-step reproduction scenario starting with `emacs -Q`. If that's too difficult, please try the bisecting instructions I just wrote there. This way we can find the offending commit and, from there, the exact line.

dgutov:
Please also try to keep debugging to `emacs-lisp-mode`: that would exclude any problems with external packages and servers.

rrudakov:
After a few days of debugging I figured out that my problem was related to `company-quickhelp` package. When I disable it everything works fine.

dgutov:
Interesting.

What pointed you that way? Did you see any related errors in the echo area or `*Messages*`? Or did the problem simply occur with the same delay that the doc popup shows after?

rrudakov:
Initially I've tried to reproduce it with `emacs -Q -L ~/.emacs.d/elpa/company-xyz -l company`. Then I disabled all company related packages in my config and enable them one by one.

No errors in echo area or `*Messages*` buffer. 

> Or did the problem simply occur with the same delay that the doc popup shows after?

yes, looks like the same delay.

JekoTronik:
Same here ! 

Anytime I press the Meta key to navigate through candidate list, the list disappears.

[...]

PR #1098, opened 2021-04-24
Change default bindings to use C-n/C-p

dgutov:
This seems to be a more popular scheme, among starter kits as well as end users
(if one judges by configuration snippets posted in various forums).

Should also work around WM misbehavior like described in
https://stackoverflow.com/questions/51585036/check-bindings-in-company-active-map?noredirect=1&lq=1
and
https://github.com/company-mode/company-mode/issues/1086#issuecomment-818698244.

And here's a recent Reddit poll: https://www.reddit.com/r/emacs/comments/my6rzb/companymode_to_change_default_bindings_from_mnmp/


yugaego:
Have it also been considered to enable cycling by default? (I didn't find such an issue.)
F.i., as a fresh company user I expected to have this feature and then stumbled upon [the example in Wiki](https://github.com/company-mode/company-mode/wiki/Tips-&-tricks).

dgutov:
> Have it also been considered to enable cycling by default? (I didn't find such an issue.)

Not thus far, no. I feel it's an advanced feature, sort of, if we want to keep the "expand common" feature on TAB as well.

Happy to be proved wrong, though.

> F.i., as a fresh company user I expected to have this feature and then stumbled upon [the example in Wiki](https://github.com/company-mode/company-mode/wiki/Tips-&-tricks).

What formed this expectation? Some other editors you used previously?

dgutov:
Thanks for the wiki link either way (it will need to be updated).

yugaego:
> What formed this expectation? Some other editors you used previously?

Sure, just a subjective feel. The point was I was wondering if the feel was shared by the other users. (Similar to how the majority seemed to be up to change M-n/p defaults to C-n/p.) If it's not, then we may simply abandon this question.

> Thanks for the wiki link either way (it will need to be updated).

👍

yugaego:
Hi Dmitry,
I've encountered an issue with the remapped `C-n/p` getting in the way when I really want to move to the next/previous line of the text itself. The theoretical solution ["just nav away with C-p"](https://www.reddit.com/r/emacs/comments/my6rzb/companymode_to_change_default_bindings_from_mnmp/gvu2xb6?utm_source=share&utm_medium=web2x&context=3) didn't actually work for me.

Not sure if I can/willing to do anything else with it on the project level at the moment, so putting it's more FYI and future reference here.
Let me know if to file this somewhere else.

For my own needs, I just remapped the tooltip navigation - ironically - [back to `M-n/p`](https://github.com/company-mode/company-mode/wiki/Tips-&-tricks#remap-companys-select-nextprevious-candidate).

dgutov:
@yugaego I'm sorry I missed this message.

I'll describe how it worked for me. First of all, as a maintainer I try to stay with the default settings of this package (with minimal tweaks, e.g. for numbers), because dogfooding is important, to make sure to catch the biggest usability problems, at least.

And I also use some minibuffer completion packages. In particular, Ivy. Ivy (like Helm, Vertico, `icomplete-vertical-mode` and etc) uses `C-n` and `C-p` for switching between completions. They seem pretty popular, too.

So a lot of times, when switching contexts between minibuffer completion and in-buffer completion I ended up using the same bindings to try to switch completions (they are positioned vertically as well, after all). But I persevered with `M-n` and `M-p`, and remembered to use them most of the time. And logically, they indeed seems like a good idea, because they don't override any important existing bindings.

After the switchover, there was a period of the problem with reverse: `M-n/p` were in the muscle memory, so I stumbled on them for a while (and I guess the "deprecated" bindings helped). But as I used `C-n/p` more, the conflicts with movement commands became less and less frequent (the brain remembers that having a rectangular popup visible changes things). So I guess I just `C-g` out when necessary.

Perhaps it's not an ideal solution still: many other completion systems use tabbing and cycling. We have a couple of different supported setups for this; maybe some particular one deserves more attention. But switching the default again would be a pretty big change.

yugaego:
Thanks for elaborating on this. It sounds reasonable.

Issue #1088, opened 2021-04-02
Making `company-text-icons-margin` more attractive


dgutov:
Now that we have a basic margin-format-function for use in the terminal, we can discuss how to make it better/more attractive/usable. Looking for opinions here, especially from terminal users.

To clarify, this is about `(setq company-format-margin-function #'company-text-icons-margin)`

* Is the lack of space between the "icon" char and the completion text a problem for most users? We've tried to choose characters that are more unusual for most users but still hold some meaning (Greek chars), but I wonder if this is enough. The simplest thing we can do is to separate it with a space. That makes the popup wider, though. Or add a `bold` face, or `italic`, or color.
* Another approach would be to, after all, choose some characters from https://en.wikipedia.org/wiki/Miscellaneous_Symbols. We'll need to choose those that are present in monospaced fonts on most systems, though.
* Yet another possibility is to add an option to display the icons on the right side of the popup. Then one could even use plain ASCII chars and longer format. The result can look like this CIDER completion example: https://docs.cider.mx/cider/_images/completion-annotations.png This is the most effort-intensive option, though.

Thoughts?

If you are a terminal user, and you like the current look well enough, please also chime in. I might be worrying over nothing.

Cc @mohkale @innerout @yyoncho

mohkale:
> Is the lack of space between the "icon" char and the completion text a problem for most users?

In retrospect I think it was really dumb of me not to include some sort of spacing there. While debugging I kept wondering why the icon wasn't showing up and turns out it was but I wasn't registering it because it's joined right onto the completion candidate so I'd recommend adding spacing. 

However it might be nicer to join this with your third suggestion and include a format option that let's you configure where everything should be. Something like:

````lisp
(defvar company-display-format `(:margin " " :candidate " " :space :annotation))`
```

Where strings are inserted literally, `:space` just right aligns everything after it, and the keywords work like you'd expect.

Now if users want the margin after the candidate they can set it to: ``(:candidate " " :space :margin :annotation)`.

Aside from either of those concerns I'd like to add a seperate face for the margin (in my setup I had company-tooltip set to bold but bold icons sometimes look weird), and maybe add support for propertizing the text icons themselves. I wasn't sure how the face inheritance would work so I ended up using a pretty ugly [hack](https://github.com/company-mode/company-mode/wiki/all-the-icons-margin-function).

dgutov:
> (defvar company-display-format `(:margin " " :candidate " " :space :annotation))`

[...]

PR #1099, opened 2021-05-05
company-text-icons-mapping: Merge faces/colors into its format

mohkale:
See #1088

### DIFF
--- a/company.el
+++ b/company.el
@@ -1515,7 +1515,7 @@ The format should be an alist of (KIND . CONF) where CONF is a list of the
 form (ICON FG BG) which is used to propertize the icon to be shown for a
 candidate of kind KIND. FG can either be color string or a face from which
 we can get a color string (using the :foreground face-property). BG must be
-of the same form as FG or a cons cell of (BG-WHEN-SELECTED . BG) which each
+of the same form as FG or a cons cell of (BG . BG-WHEN-SELECTED) which each
 should be of the same form as FG.
 
 The only mandatory element in CONF is ICON, you can omit both the FG and BG
@@ -1559,8 +1559,8 @@ See `company-text-icons-mapping'."
 (defun company-text-icons--face (fg bg selected)
   ;; Narrow to specific bg used for current candidate when a CONS cell.
   (when (consp bg)
-    (setq bg (if selected (car bg) (cdr bg))))
-  (let ((bg-color (if (facep bg) (face-attribute bg :foreground) bg))
+    (setq bg (if selected (cdr bg) (car bg))))
+  (let ((bg-color (if (facep bg) (face-attribute bg :background) bg))
         (fg-color (if (facep fg) (face-attribute fg :foreground) fg)))
     `(,@company-text-face-extra-attributes
       ,@(cond
--- a/company.el
+++ b/company.el
@@ -1483,69 +1483,40 @@ end of the match."
                                 selected))
 
 (defcustom company-text-icons-mapping
-  '((array . "a")
-    (boolean . "b")
-    (class . "c")
-    (color . "#")
-    (constant . "c")
-    (enum-member . "e")
-    (enum . "e")
-    (field . "f")
-    (file . "f")
-    (folder . "d")
-    (interface . "i")
-    (keyword . "k")
-    (method . "m")
-    (function . "f")
-    (module . "{")
-    (numeric . "n")
-    (operator . "o")
-    (parameter . "p")
-    (property . "p")
-    (ruler . "r")
-    (snippet . "S")
-    (string . "s")
-    (struct . "s")
-    (text . "t")
-    (value . "v")
-    (variable . "v")
-    (t . "."))
-  "Mapping of the text icons."
+  '((array "a" font-lock-type-face)
+    (boolean "b" font-lock-builtin-face)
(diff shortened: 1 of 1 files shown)


company.el
```diff
@@ -1558,54 +1537,66 @@ Only affects `company-text-icons-margin'."
...
-                   (if selected
-                       'company-tooltip-selection
-                     'company-tooltip)
-                   :background))
-                 (color-name-to-rgb
-                  (face-attribute face :foreground))
-                 10))))
+(defun company-text-icons--face (fg bg selected)
+  ;; Narrow to specific bg used for current candidate when a CONS cell.
+  (when (consp bg)
```
dgutov, line 1561: Very nice.

There is one case that doesn't seem to be handled, though: what happens if `bg` is not a cons and `selected` is t?

We could still try to auto-generate the color for it based on the colors of `bg` and `company-tooltip-selection`, using a similar interpolation, though perhaps with a different number of steps. The exact number needs some fiddling and would probably never be ideal, but that still seems better than not extending the selection rectangle to the left border for those entries. WDYT?

company.el
```diff
@@ -1558,54 +1537,66 @@ Only affects `company-text-icons-margin'."
...
+    (setq bg (if selected (car bg) (cdr bg))))
```
dgutov, line 1562: Perhaps swap `car` and `cdr` here? The non-selected case feels like the more important one, so it should go first.

company.el
mohkale: > But there is one case that doesn't seem to be handled: what happens if bg is not a cons and selected is t?

Then bg would (ideally) be a string, face symbol or nil. If it's a string then `:background` is set to it. If it's a face then we use that faces :background property or if that doesn't exist then we use no background. If it's nil then we generate a background using the fg-color and the tooltip bg. 

I'm not sure what condition you mean in this comment.

EDIT: I got it by the time I reach the end of second paragraph below I think.

> but that still seems better than not extending the selection rectangle

Do you mean not having a separate face for the icon in the margin when we're in `company-tooltip-selection`? Ideally I'd suggest the user supply a separate bg color (using the cons cell) if they want it. If they don't set any background thten we generate separate ones for when the candidates selected and when it isn't. If they supply only one symbol for BG my guess was that they didn't want separate colors for selected and !selected backgrounds.

Are you suggesting when `selected` is `t` and BG isn't a cons cell that we should generate another background to indicate that this is the current selection? Honestly seems kinda like an overreach to me. I'd expect if someone went to the effort of specifying their own BG colors in the config, they could go the extra mile and also specify BG for when the current candidate is selected.

company.el
mohkale: Done.

company.el
dgutov: > If they supply only one symbol for BG my guess was that they didn't want separate colors for selected and !selected backgrounds.

It might be a plausible guess, but I'm not sure there is any significant number of users who would prefer such behavior. We could guard this auto-generation behind `(when company-text-icons-add-background ...)`  as well.

> Are you suggesting when selected is t and BG isn't a cons cell that we should generate another background to indicate that this is the current selection? Honestly seems kinda like an overreach to me. I'd expect if someone went to the effort of specifying their own BG colors in the config, they could go the extra mile and also specify BG for when the current candidate is selected.

I was thinking some users might want to emulate what Atom does (I showed the screenshot in the other thread): pick ~5 simple, pleasant background colors, associate each of the kinds with one of them, and let our code do the remaining blending. That would be a low-effort configuration that could, however, look nice if we do this extra step.

Further, the `company-tooltip-selection` colors come from the theme. I'm guessing not everybody will want to change their `company-text-icons-mapping` customization as they switch between themes. Whereas, if they only need to specify the base color, the result should look good enough across a whole number of dark themes (for example). They'll need to choose different base colors after switching to a light theme, though.

mohkale: @dgutov 

Okay, I understand the motivations. So just referring to the implementation how do we approach this. If there *is* a background (eg: string or face, not cons) do we blend from the background instead of the foreground, or would we prefer blending from the foreground? Just curious what precedence we're going for here.

dgutov: From the background.

Using some different number of "steps" rather than 10, probably. Maybe 2-4? That needs some testing.

mohkale: > Using some different number of "steps" rather than 10, probably. Maybe 2-4? That needs some testing.

I was hoping we could add a user option to configure the steps. `company-text-icons-bg-gradient-step`, I think I might've changed the steps from version you had before by accident cause I'm finding the contrast a little weaker.

mohkale: Alright, done. It's kind of ugly IMO but it works.

dgutov: Thank you, looks good. I'll try it out a bit later and merge.

> I was hoping we could add a user option to configure the steps. company-text-icons-bg-gradient-step

No objection from me, though now we have two values of STEPS to possibly configure. I guess we could allow `company-text-icons-add-background` to take a list value as well (with numbers). The first number would be the override for 10, the second one will override 2.

mohkale: Well for now using 10 and 2 looks well enough. If there comes a demand down the line for customization we can add the options then but for now it all looks fine to me.

dgutov:
Thanks!

PR #859, opened 2019-01-15
make numbers after candidates customizable

yqrashawn:

Issue #188, opened 2014-09-19
Type a letter to select a candidate


fniessen:
Hello,

To select one of the first 10 candidates, we now can press 'M-1' to 'M-0'. That's not the easiest key binding on a French keyboard as the figures must be "shifted". Hence, 'M-1' is for us pressing 'M-S-&'.

A better idea IMHO would be to use letters instead of figures... It could be the first 10 letters which aren't possible completions of the proposed words. For example, after typing "po":
- popup       a
- poids       b
- poignée     c
- point       f
- podio       g
- poetsen     h
- pointeur    j
- pointeurs   k
- poissons    l
- pointu      m

('d', 'e' and 'i' were not eligible)

Using letters has the added advantage that a lot of us can type them without looking at all at the keyboard. It'd be quicker, and never would imply using 'M-' nor 'S-'.

WDYT?

yugaego:
I've thought over all the related information I could find, and it seems to me there's no need at all to implement anything new fundamentally.

Firstly,
For *a-la* `ace-jump-mode` behavior (the project itself is dead btw), simply use `isearch` with `hi-lock-mode` enabled.

Or even better, use `Company` provided by default filtering `M-C-s` or searching `C-s` (press a key binding and start typing letters of a desired term --> quite a "jump").


Secondly,
Related (old-ish) PR #859 made quick-access representation layer - aka `numbers` - customizable. PR author had additionally configured custom bindings (letters) to call `Company`'s  defaults of `M-[digit]`.

So I've quick-tested that the following *rough* example seems to be a working solution.

```
(defun my/company-quick-access-key (numbered)
  "Replace default behavior of `company--show-numbers' function.
This example lists Dvorak layout home row keys."
  (let ((value (mod numbered 10)))
    (format " %s"
            (cond ((eql value 1) "a") ; a a
                  ((eql value 2) "o") ; b s
                  ((eql value 3) "e") ; c d
                  ((eql value 4) "u") ; d f
                  ((eql value 5) "i") ; k g
                  ((eql value 6) "d") ; l h
                  ((eql value 7) "h") ; m j
                  ((eql value 8) "t") ; x k
                  ((eql value 9) "n") ; y l
                  (t "s")))))         ; z ;

(setq company-show-numbers-function #'my/company-quick-access-key)

(define-key company-active-map (kbd "M-a") (kbd "M-1"))
(define-key company-active-map (kbd "M-o") (kbd "M-2"))
[...]

PR #1103, opened 2021-05-17
Add faces for quick-access numbers

### LINKED ISSUE #303: Add a face for quick-access numbers
Can you add a face for the quick-access numbers?  Right now, they're fontified in the default tooltip face, which is not as nice (see http://screencast.com/t/K9YwiWJa) when we want to have the "different part" of the candidates bolded.

Thanks in advance!

yugaego:
Closes #303.

@dgutov Hi Dmitry, please review the suggested changes.

Built-in **leuven** theme, having customized Company faces, is a representative example:
leuven

I've got two more - aside - concerns:

1. The term `numbers` stands for the representation, so `quick-access` (action) or `candidate-id` (entity) could potentially be a better choice. `(defface company-tooltip-quck-access)` or similar perhaps could provide better context to what the face - and functionality overall - does. **But**, the term is used throughout the code, so had to stick with the `numbers`.
The term may also lead to confusion when first met, such as "Which numbers? Numbers of occurrences maybe? Or numbers in the candidates? Or the numbers I type? ... etc".

2. I was confused to find `company--show-numbers` function under `;; replace` section caption. Should `company--show-numbers` and the suggested by this PR `company--show-numbers-propertize` be moved to another location?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,9 @@
 
 ## Next
 
+* New faces `company-tooltip-quick-access` and
+  `company-tooltip-quick-access-selection`
+  ([#303](https://github.com/company-mode/company-mode/issues/303)).
 * Default colors for dark themes have been changed
   ([#949](https://github.com/company-mode/company-mode/issues/949)).
 * Default key bindings have been changed, moving `company-select-next` and
--- a/company.el
+++ b/company.el
@@ -126,6 +126,16 @@
   '((default :inherit company-tooltip-annotation))
   "Face used for the selected completion annotation in the tooltip.")
 
+(defface company-tooltip-quick-access
+  '((default :inherit company-tooltip-annotation))
+  "Face used for the quick-access keys shown in the tooltip."
+  :package-version '(company . "0.9.14"))
+
+(defface company-tooltip-quick-access-selection
+  '((default :inherit company-tooltip-annotation-selection))
+  "Face used for the selected quick-access keys shown in the tooltip."
+  :package-version '(company . "0.9.14"))
+
 (defface company-scrollbar-fg
   '((((background light))
      :background "darkred")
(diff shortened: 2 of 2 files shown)


yugaego:
(I just sent a request to assign@gnu.org)

dgutov:
Hi!

I think the face should be called either `company-tooltip-number` (singular; and, correspondingly, `company-tooltip-selection-number`), or indeed `company-quick-access`.

The latter might work better if we ever get around to implementing https://github.com/company-mode/company-mode/issues/188. I'm not sure how it should work, though.

But `company--show-numbers` can/should be renamed to something without a double-dash anyway.

Regarding `;; replace` - sure, please go ahead. Looks like I've never noticed these particular comments.

yugaego:
> The latter might work better if we ever get around to implementing #188. I'm not sure how it should work, though.

Thank you for pointing me to #188. It's a tempting one to implement. I'll **try** to find a possible solution - which would take me a while in any case. And will get back to this PR after that.

dgutov:
There can be a possible middle-ground: we don't add that feature OOTB but document how one would go about implementing it, e.g. using the `company-show-numbers-function` variable. That's why it was added, actually:

https://github.com/company-mode/company-mode/pull/859
https://yqrashawn.com/2019/01/15/use-emacs-with-home-row/

It's certainly up to you how to proceed. I suppose it depends on whether you want to use it yourself.

yugaego:
Yes, that's basically what I've come to either. I'll post an extended response at #188, as it would better suit there.

yugaego:
Hi Dmitry,
It's now clear more or less what the terms would look like after closing the rest of the changes in work, so let's try to move on with this one, also resolving (for me) the trouble of having too many unfinished changes.
Let me know what you think!

dgutov:
Looking good, thanks!

Issue #273, opened 2015-01-07
Error when searching a candidate that contains a digit


Sarcasm:
Hello Dmitry,

I have a list of candidates like so:

* vec3
* vec4
* mat3
* mat4
* ivec3
* ivec4
* ...

I want to search in the long candidate list everything that contains `vec3` (should match `vec3` and `ivec3`).

I can search `vec` but as soon as I type the `3`, company triggers an error:

     Error in post-command-hook (company-post-command): (wrong-type-argument sequencep stop)

dgutov:
Hi Guillaume,

I'm having trouble reproducing this. I think I saw it once, but after that, in every session and with every backend I've tried, searching for a number works fine. For instance, with this Lisp code:

```
(let (var1
      var2 var3
      mat3 mat4)
  v|)
```

initiating completion and typing `ec3` highlights the relevant portion of the third candidate.

Sarcasm:
So I tried at home your example and while the first few times I had the feeling something wasn't right it looks like it worked properly the other times.

But today, re-testing at work (Emacs 24.3.1) it is still present.

With your example I get a different error when searching a digit, i.e: `ar1` or `1`, I get:

    Matching input is required

My example with company-irony is still producing `Wrong type argument: sequencep, stop`. I think something looks shady, do you have any tips to debug the company-mode post-command? To have something easier to debug?

dgutov:
> the first few times I had the feeling something wasn't right

The might be because the interface is clunky and different from isearch.

> But today, re-testing at work (Emacs 24.3.1) it is still present.

Even starting with `emacs -Q`? I've tried that version, and haven't seen any difference.

> I think something looks shady, do you have any tips to debug the company-mode post-command?

`edebug-defun` works for me, but it's probably supported only starting with Emacs 24.4.

But you should probably look at `company-search-map` and see why `company-search-printing-char` isn't called. I suspect it might be related to input methods. You aren't using the keypad, are you?

Sarcasm:
> Even starting with `emacs -Q`? I've tried that version, and haven't seen any difference.

Just tested with `emacs -Q`, and company version: ~/.emacs.d/elpa/company-20150102.1526/

> `edebug-defun` works for me, but it's probably supported only starting with Emacs 24.4.

Looks like it work in Emacs 24.3.1, just not sure how to debug this yet.

> You aren't using the keypad, are you?

Woopsy-doopsy. I am, non-keypad digit seems to work.

[...]

Issue #177, opened 2014-09-05
does not work with evil-repeat


redguardtoo:
enable evil-mode
type some prefix which can be completed by company-mode, for example, suppose there is already a function definition myServiceCall and I type `myser`,

the candidate myServiceCall is displayed at the dropdown. I press `M-1` to finish the completion of that candidate.

move cursor any where and press `.` (the dot character which corresponds to the evil command evil-repeat) in evil-normal state. _Only_ myser is inserted this time.

company-version 0.8.0
emacs 24.3.1
evil-1.0-dev

PR #1104, opened 2021-05-22
Replace term number(s) by term quick-access

yugaego:
#1103, #188

@dgutov Dmitry, your input is very welcome!

This is more of a WIP than an actual PR at this stage: I'm still working on possibly providing `M-(letter)` as one of the easily chosen package options. If I find a way to provide it *nicely*, then highly possible `company-quick-access-key-producer` and `company-quick-access-digit` (maybe others) would have to be changed again.

So please let me know if you see any issues with the suggested changes. (I don't like to introduce many breaking changes, but obsoletes declarations seem to be a quite comfortable approach.)
Thanks.

### DIFF
--- a/company.el
+++ b/company.el
@@ -645,16 +645,35 @@ commands in the `company-' namespace, abort completion."
                         (repeat :tag "Commands" function))
                  (repeat :tag "Commands" function)))
 
-(defcustom company-show-numbers nil
-  "If enabled, show quick-access numbers for the first ten candidates."
+(define-obsolete-variable-alias
+ 'company-show-numbers
+ 'company-show-quick-access
+ "0.10.0")
+
+(defcustom company-show-quick-access nil
+  "Show a quick-access column for the first ten candidates.
+
+Non-nil value enables a quick-access column on the right side of the tooltip,
+ unless the configured value is \\='left.
+If the value is nil, do not show a quick-access column.
+NOTE: A value of nil does not disable `M-(digit)' bindings.
+
+To customize shown quick-access keys, use `company-quick-access-key-producer'."
   :type '(choice (const :tag "off" nil)
                  (const :tag "left" 'left)
                  (const :tag "on" 't)))
 
-(defcustom company-show-numbers-function #'company--show-numbers
-  "Function called to get quick-access numbers for the first ten candidates.
+(define-obsolete-variable-alias
+ 'company-show-numbers-function
+ 'company-quick-access-key-producer
+ "0.10.0")
 
-The function receives the candidate number (starting from 1) and should
+(defcustom company-quick-access-key-producer #'company-quick-access-digit
+  "Function called to produce a quick-access key.
+
+The function called only if `company-show-quick-access'
+ is set to non-nil value.
+It receives the candidate number (starting from 1) and should
 return a string prefixed with one space."
   :type 'function)
 
@@ -710,7 +729,7 @@ asynchronous call into synchronous.")
     (define-key keymap "\C-s" 'company-search-candidates)
     (define-key keymap "\C-\M-s" 'company-filter-candidates)
     (dotimes (i 10)
-      (define-key keymap (read-kbd-macro (format "M-%d" i)) 'company-complete-number))
+      (define-key keymap (read-kbd-macro (format "M-%d" i)) 'company-complete-quick-access))
      keymap)
   "Keymap that is enabled during an active completion.")
 
@@ -2056,7 +2075,7 @@ prefix match (same case) will be prioritized."
                                           company-complete
                                           company-complete-common
                                           company-complete-selection
-                                          company-complete-number)
+                                          company-complete-quick-access)
   "List of commands after which idle completion is (still) disabled when
 `company-begin-commands' is t.")
 
@@ -2273,7 +2292,7 @@ each one wraps a part of the input string."
     (define-key keymap "\C-r" 'company-search-repeat-backward)
     (define-key keymap "\C-o" 'company-search-toggle-filtering)
     (dotimes (i 10)
-      (define-key keymap (read-kbd-macro (format "M-%d" i)) 'company-complete-number))
+      (define-key keymap (read-kbd-macro (format "M-%d" i)) 'company-complete-quick-access))
     keymap)
   "Keymap used for incrementally searching the completion candidates.")
 
@@ -2521,11 +2540,18 @@ inserted."
       (when company-candidates
         (setq this-command 'company-complete-common)))))
 
-(defun company-complete-number (n)
+(define-obsolete-function-alias
+ 'company-complete-number
+ 'company-complete-quick-access
+ "0.10.0")
+
+(defun company-complete-quick-access (n)
   "Insert the Nth candidate visible in the tooltip.
-To show the number next to the candidates in some backends, enable
-`company-show-numbers'.  When called interactively, uses the last typed
-character, stripping the modifiers.  That character must be a digit."
+
+When called interactively, uses the last typed
+character, stripping the modifiers.  That character must be a digit.
+To enable a quick-access column for some backends, use
+`company-show-quick-access'."
   (interactive
    (list (let* ((type (event-basic-type last-command-event))
                 (char (if (characterp type)
@@ -2542,6 +2568,16 @@ character, stripping the modifiers.  That character must be a digit."
     (cl-decf n)
     (company-finish (nth (+ n company-tooltip-offset)
                          company-candidates))))
+
+(define-obsolete-function-alias
+ 'company--show-numbers
+ 'company-quick-access-digit
+ "0.10.0")
+
+(defun company-quick-access-digit (numbered)
+  "Produce a digit to show in a quick-access column.
+This function used as a default value of `company-quick-access-key-producer'."
+  (format " %d" (mod numbered 10)))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
@@ -2985,9 +3021,6 @@ If SHOW-VERSION is non-nil, show the version in the echo area."
           new
           (company-safe-substring old (+ offset (length new)))))
 
-(defun company--show-numbers (numbered)
-  (format " %d" (mod numbered 10)))
-
 (defsubst company--window-height ()
   (if (fboundp 'window-screen-lines)
       (floor (window-screen-lines))
@@ -3155,7 +3188,7 @@ If SHOW-VERSION is non-nil, show the version in the echo area."
     (setq width (min window-width
                      company-tooltip-maximum-width
                      (max company-tooltip-minimum-width
-                          (if company-show-numbers
+                          (if company-show-quick-access
                               (+ 2 width)
                             width))))
 
@@ -3164,7 +3197,7 @@ If SHOW-VERSION is non-nil, show the version in the echo area."
       (setq company--tooltip-current-width width))
 
     (let ((items (nreverse items))
-          (numbered (if company-show-numbers 0 99999))
+          (numbered (if company-show-quick-access 0 99999))
           new)
       (when previous
         (push (company--scrollpos-line previous width) new))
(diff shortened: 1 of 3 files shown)


company.el
```diff
@@ -645,16 +645,35 @@ commands in the `company-' namespace, abort completion."
                         (repeat :tag "Commands" function))
                  (repeat :tag "Commands" function)))
 
-(defcustom company-show-numbers nil
-  "If enabled, show quick-access numbers for the first ten candidates."
+(define-obsolete-variable-alias
+ 'company-show-numbers
+ 'company-show-quick-access
+ "0.10.0")
```
dgutov, line 651: The next version will hopefully be 1.0 :slightly_smiling_face: I can change this line later either way.

Not sure if mentioning an unreleased version in an obsoletion message is idiomatic, but it's probably okay.

company.el
```diff
@@ -645,16 +645,35 @@ commands in the `company-' namespace, abort completion."
...
                  (const :tag "on" 't)))
 
-(defcustom company-show-numbers-function #'company--show-numbers
-  "Function called to get quick-access numbers for the first ten candidates.
+(define-obsolete-variable-alias
+ 'company-show-numbers-function
+ 'company-quick-access-key-producer
+ "0.10.0")
+
+(defcustom company-quick-access-key-producer #'company-quick-access-digit
```
dgutov, line 671: Call this one `company-quick-access-key-formatter`, maybe?

company.el
```diff
@@ -645,16 +645,35 @@ commands in the `company-' namespace, abort completion."
...
+
+(defcustom company-show-quick-access nil
+  "Show a quick-access column for the first ten candidates.
+
+Non-nil value enables a quick-access column on the right side of the tooltip,
+ unless the configured value is \\='left.
+If the value is nil, do not show a quick-access column.
+NOTE: A value of nil does not disable `M-(digit)' bindings.
+
+To customize shown quick-access keys, use `company-quick-access-key-producer'."
```
dgutov, line 661: Perhaps something like:

> To change the quick access keys shown, customize `company-quick-access-key-formatter'

?

company.el
```diff
@@ -645,16 +645,35 @@ commands in the `company-' namespace, abort completion."
...
+(define-obsolete-variable-alias
+ 'company-show-numbers-function
+ 'company-quick-access-key-producer
+ "0.10.0")
+
+(defcustom company-quick-access-key-producer #'company-quick-access-digit
+  "Function called to produce a quick-access key.
 
-The function receives the candidate number (starting from 1) and should
+The function called only if `company-show-quick-access'
```
dgutov, line 674: Suggestion:

> It's only used if `company-show-quick-access` is non-nil.

company.el
```diff
@@ -2521,11 +2540,18 @@ inserted."
       (when company-candidates
         (setq this-command 'company-complete-common)))))
 
-(defun company-complete-number (n)
+(define-obsolete-function-alias
+ 'company-complete-number
+ 'company-complete-quick-access
+ "0.10.0")
+
+(defun company-complete-quick-access (n)
```
dgutov, line 2548: Should we perhaps keep this command's name?

It only works with numbers: if the user wants to handle different inputs, they will need to either change these bindings, or add key translations as you described. But this command will still only work with numbers itself.

company.el
```diff
@@ -2521,11 +2540,18 @@ inserted."
...
   "Insert the Nth candidate visible in the tooltip.
-To show the number next to the candidates in some backends, enable
-`company-show-numbers'.  When called interactively, uses the last typed
-character, stripping the modifiers.  That character must be a digit."
+
+When called interactively, uses the last typed
+character, stripping the modifiers.  That character must be a digit.
+To enable a quick-access column for some backends, use
```
dgutov, line 2553: Why "for some backends"?

dgutov:
Hi Y.E.!

That's pretty comprehensive. :-)

I like the general direction. But see some of the comments on individual lines.

> If I find a way to provide it nicely, then highly possible company-quick-access-key-producer and company-quick-access-digit (maybe others) would have to be changed again.

That makes sense, too.

company.el
yugaego: Great! I'm all for it (with a preference of a semantic versioning).

> unreleased version

It seemed quite idiomatic to me. **What would be an alternative?** "0.9.13" seems vague, because the release doesn't contain the obsoletion definitions. (And that that the argument *WHEN* isn't optional since Emacs 28.1. BTW, the day before I re-compiled Emacs from master: the app shows up as 28.0.50 and contains the features marked for unreleased 28.1.)

company.el
yugaego: Good idea!

company.el
yugaego: I like your suggestion.

company.el
yugaego: Dmitry, I only now accidentally (consciously?) noticed there's a couple more of "0.9.14" version appearances in the code. They'd require a fix up too.

company.el
yugaego: https://github.com/company-mode/company-mode/blob/d488122d18645a20a596490e3551194aecdea08a/company.el#L2527

"in some backends" is in the master (and been there for many years), so I hesitated about it and decided not to remove. Can it be safely removed?

company.el
yugaego: Yes, I understand the concerns.
I gave it a lot of thought and  concluded that the 'quick-access' part:
  - better represents what the function does (with a high chance of `company-complete-number` name giving an impression that's a function to complete numbers a user types).
  - provides better uniformity for quick-access functionality throughout the code.

Also, I tried and failed to use it interactively, so might be that's where an important detail could be for keeping the 'number' (or 'digit') part?

Anyway, **I suggest postponing with the final decision and get back to this question on the next review.**

yugaego: By "to complete numbers a user types" I mean typing in the editor before tooltip shows-up (kind of an alternate function to complete only numbers.)

company.el
dgutov: Good eye. :-)

I suppose you could also use the "0.9.14" version string, which will then have to be changed wholesale before the release.

company.el
dgutov: > By "to complete numbers a user types" I mean typing in the editor before tooltip shows-up (kind of an alternate function to complete only numbers.)

Yeah, I get what you mean. But perhaps there could be another name which is also as specific (e.g. probably has "number" in it), but doesn't give that impression.

> I suggest postponing with the final decision and get back to this question on the next review

Sounds good. My objection wasn't strong anyway.

company.el
dgutov: Guess I missed that during the review when this was added. :disappointed: 

Yes, remove it please.

company.el
yugaego: Sounds good.

company.el
yugaego: Yeah, I've started considering a rather longish `company-complete-quick-access-(digit|number)`. Will give it some more thought.

yugaego:
Dmitry, I've been making sure my changes won't break any other appearances of a *magic number* 10 in `company.el`, and got a couple of questions:

1. What `company-search-keypad` is expected to do? I read the code and re-bound the function call to the keys I have, but still couldn't get it (maybe should try again later, dunno). My main concern is if anything related to quick-access keys can be broken here or should be tested?

2. I noticed that `company-echo-frontend` outputs zero-based numbers to mark the candidates. But that doesn't seem to make sense having implicit 1-based quick-access keys. Am I missing something here and there's some other - proper - way to use `company-echo-frontend`?

And, just as a side note, I also noticed that [the manual quick-access letters setup](https://github.com/company-mode/company-mode/issues/188#issuecomment-843803989) would break quick-access bindings on filtering (naturally, as they're not re-bounded). Hence, hopefully the changes I'm working on could be successfully implemented without breaking the filtering.

dgutov:
> I've been making sure my changes won't break any other appearances of a magic number 10 in company.el

Nice :-) You found a couple of dusty corners which possibly haven't been used by anybody in the recent few years. But I can't be sure.

> What company-search-keypad is expected to do?

Looking at the code, it's trying to translate keypad keypresses (such as `<kp-1>`, etc), into normal searches for the corresponding digits. See the list here: https://www.gnu.org/software/emacs/manual/html_node/emacs/Function-Keys.html

I don't have a keypad on either of the keyboards lying before me ATM, so I can't really test this right now.

> But that doesn't seem to make sense having implicit 1-based quick-access keys. Am I missing something here and there's some other - proper - way to use company-echo-frontend?

Looks like a desynchronization that happened ~12 years ago, after the original author of this package added this feature to two frontends, but then everybody only used one of them.

It's worth fixing, and I don't think anyone would miss the current behavior. `company-echo-frontend` feels more like an "example" frontend anyway to me.

dgutov:
> And, just as a side note, I also noticed that the manual quick-access letters setup would break quick-access bindings on filtering (naturally, as they're not re-bounded)

Would it help if the same rebinding were applied to `company-search-map` as well?

yugaego:
> Would it help if the same rebinding were applied to company-search-map as well?

Yes, this approach works (though I didn't fully understand **why** yet). Thanks.

As for `company-search-keypad`, it's great that we've got so much history stored: this function was added by https://github.com/company-mode/company-mode/commit/0a172c53217dd20fbc4c331780cd050ee3c9fdc9 with [this reasoning](https://github.com/company-mode/company-mode/issues/273#issuecomment-69241456). If I got it correctly, it fixed an issue on typing numbers of the keypad (though might be only for a backend which is absent now?). I don't have a keypad either, so cannot test with more detail unfortunately. Anyway, it closes my question if I should cover this part of the code while working on the quick-access changes - **I shouldn't**.

> It's worth fixing, and I don't think anyone would miss the current behavior. company-echo-frontend feels more like an "example" frontend anyway to me.

Great, then I'd probably do it.

dgutov:
> though I didn't fully understand why yet

We switch to a different keymap when the search mode is entered.

> If I got it correctly, it fixed an issue on typing numbers of the keypad (though might be only for a backend which is absent now?)

Bindings are backend-agnostic. I'm not sure why we needed to do this (one would imagine that keymap numbers should have automatic translation to "normal" ones), but there's no urgency in investigating this again.

> Great, then I'd probably do it.

:+1:

yugaego:
Hi Dmitry,
What if introduce [`company-complete-nth`](https://github.com/yugaego/company-mode/blob/3d4a5831a5fb28be92da26909919b6a19276a4d3/company.el#L2571) strictly for package-set quick-access key bindings, configurable by [`company-quick-access-keys`](https://github.com/yugaego/company-mode/blob/3d4a5831a5fb28be92da26909919b6a19276a4d3/company.el#L676)?

Then specific `company-complete-number` concerns could be handled separately (now or later).
Those are, according to my current understanding:
 1. BC: https://github.com/company-mode/company-mode/issues/177
    - I'm not sure switching to `company-complete-nth` would be not-breaking solution for such cases.

 2. BC keypad numbers implicit support: commit 43756857
    - Might be move keypad numbers bindings to keymaps?
 
 3. Maybe just refactor `company-complete-number` to call (company-complete-nth (1- n)) and keep it for backward compatibility mentioned in points 1. and 2.?
    - Could you provide some context on what are possible company-toolip-offset values/implications? (This value also isn't handled in `company-complete-nth` yet.)

4. I didn't/couldn't test properly `company-complete-number` BC points, so this is an additional concern.

dgutov:
Hi Y.E., sorry for the late reply.

> What if introduce company-complete-nth strictly for package-set quick-access key bindings, configurable by company-quick-access-keys?

Sounds good to me.

> I'm not sure switching to company-complete-nth would be not-breaking solution for such cases.

Can't say for sure, without having a branch to test and actually asking Evil users. Folks from that issue can still be around.

I have very little experience with Evil myself. But if anything comes up, some kind of fix will probably be arranged.

> Might be move keypad numbers bindings to keymaps?

Sure. If you can test yourself how it's working, that would be great. Otherwise, I could ping the reporter of that old bug.

> Maybe just refactor company-complete-number to call (company-complete-nth (1- n)) and keep it for backward compatibility mentioned in points 1. and 2.?

And mark it as obsolete as well, right?

> Could you provide some context on what are possible company-toolip-offset values/implications? (This value also isn't handled in company-complete-nth yet.)

`company-tooltip-offset` is the number of the first completion that is visible in the popup. It's the current scrolling position, basically. When printing numbers, we subtract its from the completion's number, so that all quick access numbers can still be used even in a scrolled popup. But, conversely, when going from the typed number to the candidate's number, we need to add that offset.

P.S. What's a BC? :-)

yugaego:
No problem, got additional time to read Elisp docs ;-)

Your feedback answered the "design" questions I had - *thanks*, so yes, I'll push something more comprehensible for the further review.

`BC` was meant to stand for Backward Compatibility.

yugaego:
Covered by #1118 and #1115 .

PR #1108, opened 2021-05-27
Change echo-common face for light themes

yugaego:
Improves visibility of common part(s) of completions with echo-frontends for all built-in light themes.
**before**
before
**after**
after

(And transposes the light/dark theme lines to match the order used by the rest of the deffaces.)

### DIFF
--- a/company.el
+++ b/company.el
@@ -156,8 +156,8 @@
   "Face used for completions in the echo area.")
 
 (defface company-echo-common
-  '((((background dark)) (:foreground "firebrick1"))
-    (((background light)) (:background "firebrick4")))
+  '((((background light)) (:foreground "firebrick4"))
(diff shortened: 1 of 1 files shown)


dgutov:
Excellent, thank you.

yugaego:
BTW, any need to put a line to the NEWS in such cases: a minor change that's 'user-visible' only in rare cases?

dgutov:
Yeah, not really.

I usually reserve NEWS to prominent new features and really important bug fixes (for bugs that have been present for longer than the current version). Basically, info that helps the user to decide they need to upgrade, or for backend/frontend authors to specify minimum required company version.

yugaego:
Thank you for the clarification.

PR #1111, opened 2021-06-02
Fix quick-access numbers shown in echo frontends

yugaego:
Replaces zero-based digits by quick-access bindings digits.
Discussed in #1104.

### DIFF
--- a/company.el
+++ b/company.el
@@ -3564,19 +3564,18 @@ Delay is determined by `company-tooltip-idle-delay'."
           (len -1)
           ;; Roll to selection.
           (candidates (nthcdr selection company-candidates))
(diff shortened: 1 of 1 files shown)


dgutov:
Thanks!

One bit I'm not sure of it renaming the var to `row`: since these frontends usually display the completions in horizontal fashion, they all reside on the same now.

So maybe rename it back to `i`. Or call it `num` (`number`)? Or `index`.

dgutov:
Or `numbered`, for simple consistency with the other code.

yugaego:
> these frontends usually display the completions in horizontal fashion

Good catch.

yugaego:
Done, renamed to `numbered` for consistency.

PR #1112, opened 2021-06-02
Fix quick-access column formatting

yugaego:
Do not break quick-access column when `company-tooltip-limit` > 10.

**before**
left-before
**after**
left-after

**before**
right-before
**after**
right-after


dgutov:
Thanks!

PR #1115, opened 2021-06-08
Refine quick-access keys customization

yugaego:
WIP for #188 

Hi Dmitry, this is a draft for possible addition of quick-access keys customization. It's planned to be based on PR #1104, which can be influenced by the changes introduced here. I prefer at the moment to keep them separated, hence this draft PRs noise. (Or maybe I'm not aware of a convenient way to have different branches discussed in one PR thread.) If they become tightly coupled, I'll simply close #1104.

Could you help with clarifying the following questions and concerns:

0. Is the direction overall fine with you?  (It doesn't reflect yet *everything* we discussed regarding `company-complete-number` and more work is planned to be put into it overall.)

1. With the current implementation, if to `(setq company-quick-access-keys '("a" "b" "c"))` after `Company` is loaded, key bindings are set to *default* value of `company-quick-access-keys`. What direction should I look at to fix this issue? (I'm a bit cheating here not trying to figure it out by myself, but the cognitive load had temporarily overheated me.)  

2. Probably, one of the options is to allow setting *key sequences* and parse non-modifier part for output. What's your opinion on it?
   (I suppose, `defcustom` would require something like:
     - `:type '(repeat key-sequence)`
     - `:set 'a-function-to-read-key-sequences`
   And since parsing of a non-modifier part cannot - I think - be reliable, might be getting back to alist idea which could be too wordy IMO: `'(("M-a" . "a") ("M-b" . "b") ... ("M-z" . "z")`)

3. Considering current implementation approach,  should the values of `company-quick-access-keys` be restricted somehow (f.i. to be in a range of characters 0 to 127)? This concern is based on this **info** chapter: *22.7.15 Putting Keyboard Events in Strings*.

4. Current implementation suggests `company-quick-access-keys` to fully replace `company-show-numbers-function`, which - supposedly - isn't widely used. Still, I'm wondering how high the chances are it could be an undesirable change?

Thank you.

### DIFF
--- a/company.el
+++ b/company.el
@@ -128,12 +128,12 @@
 
 (defface company-tooltip-quick-access
   '((default :inherit company-tooltip-annotation))
-  "Face used for the quick-access keys shown in the tooltip."
+  "Face used for the quick-access hints shown in the tooltip."
   :package-version '(company . "0.9.14"))
 
 (defface company-tooltip-quick-access-selection
   '((default :inherit company-tooltip-annotation-selection))
-  "Face used for the selected quick-access keys shown in the tooltip."
+  "Face used for the selected quick-access hints shown in the tooltip."
   :package-version '(company . "0.9.14"))
 
 (defface company-scrollbar-fg
@@ -655,18 +655,126 @@ commands in the `company-' namespace, abort completion."
                         (repeat :tag "Commands" function))
                  (repeat :tag "Commands" function)))
 
-(defcustom company-show-numbers nil
-  "If enabled, show quick-access numbers for the first ten candidates."
-  :type '(choice (const :tag "off" nil)
-                 (const :tag "left" 'left)
-                 (const :tag "on" 't)))
+(defun company-custom--set-quick-access (option value)
+  "Re-bind quick-access key sequences on OPTION VALUE change."
+  (when (boundp 'company-active-map)
+    (company-keymap--unbind-quick-access company-active-map))
+  (when (boundp 'company-search-map)
+    (company-keymap--unbind-quick-access company-search-map))
+  (custom-set-default option value)
+  (when (boundp 'company-active-map)
+    (company-keymap--bind-quick-access company-active-map))
+  (when (boundp 'company-search-map)
+    (company-keymap--bind-quick-access company-search-map)))
 
-(defcustom company-show-numbers-function #'company--show-numbers
+(defcustom company-quick-access-keys '("1" "2" "3" "4" "5" "6" "7" "8" "9" "0")
+  "Character strings used as a part of quick-access key sequences.
+To change this value without Customize interface, use `customize-set-variable'.
+
+To change the quick-access key sequences modifier, customize
+`company-quick-access-modifier'.
+
+If `company-show-quick-access' is non-nil, show quick-access hints
+beside the candidates."
+  :set #'company-custom--set-quick-access
+  :type '(choice
+          (const :tag "Digits" ("1" "2" "3" "4" "5" "6" "7" "8" "9" "0"))
+          (const :tag "QWERTY home row" ("a" "s" "d" "f" "g" "h" "j" "k" "l" ";"))
+          ;; TODO un-comment on removal of `M-n' `company--select-next-and-warn'.
+          ;; (const :tag "Dvorak home row" ("a" "o" "e" "u" "i" "d" "h" "t" "n" "s"))
+          (repeat :tag "User defined" string))
+  :package-version '(company . "0.9.14"))
+
+(defcustom company-quick-access-modifier 'meta
+  "Modifier key used for quick-access keys sequences.
+To change this value without Customize interface, use `customize-set-variable'.
+See `company-quick-access-keys' for more details."
+  :set #'company-custom--set-quick-access
+  :type '(choice (const :tag "Meta key" meta)
+                 (const :tag "Super key" super)
+                 (const :tag "Hyper key" hyper)
+                 (const :tag "Alt key" alt))
+  :package-version '(company . "0.9.14"))
+
+(defun company-keymap--quick-access-modifier ()
+  "Return string representation of the `company-quick-access-modifier'."
+  (if-let ((modifier (assoc-default company-quick-access-modifier
+                                    '((meta . "M")
+                                      (super . "s")
+                                      (hyper . "H")
+                                      (alt . "A")))))
+      modifier
+    (warn "company-quick-access-modifier value unknown: %S"
+          company-quick-access-modifier)
+    "M"))
+
+(defun company-keymap--unbind-quick-access (keymap)
+  (let ((modifier (company-keymap--quick-access-modifier)))
+    (dolist (key company-quick-access-keys)
+      (let ((key-seq (company-keymap--kbd-quick-access modifier key)))
+        (when (equal (lookup-key keymap key-seq) 'company-complete-quick-access)
+          (define-key keymap key-seq nil))))))
+
+(defun company-keymap--bind-quick-access (keymap)
+  (let ((modifier (company-keymap--quick-access-modifier)))
+    (dolist (key company-quick-access-keys)
+      (let ((key-seq (company-keymap--kbd-quick-access modifier key)))
+        (if (lookup-key keymap key-seq)
+            (warn "Key sequence %s already bound" (key-description key-seq))
+          (define-key keymap key-seq #'company-complete-quick-access))))))
+
+(defun company-keymap--kbd-quick-access (modifier key)
+  (kbd (format "%s-%s" modifier key)))
+
+(define-obsolete-variable-alias
+  'company-show-numbers
+  'company-show-quick-access
+  "0.9.14")
+
+(defcustom company-show-quick-access nil
+  "If non-nil, show quick-access hints beside the candidates.
+
+For a tooltip frontend, non-nil value enables a column with the hints
+on the right side of the tooltip, unless the configured value is `left'.
+
+To change the quick-access key bindings, customize `company-quick-access-keys'
+and `company-quick-access-modifier'.
+
+To change the shown quick-access hints, customize
+`company-quick-access-hint-function'."
+  :type '(choice (const :tag "off" nil)
+                 (const :tag "left" left)
+                 (const :tag "on" t)))
+
+(defcustom company-show-numbers-function nil
   "Function called to get quick-access numbers for the first ten candidates.
 
 The function receives the candidate number (starting from 1) and should
 return a string prefixed with one space."
   :type 'function)
+(make-obsolete-variable
+ 'company-show-numbers-function
+ "use `company-quick-access-hint-function' instead,
+but adjust the expected values appropriately."
+ "0.9.14")
+
+(defcustom company-quick-access-hint-function #'company-quick-access-hint-key
+  "Function called to get quick-access hints for the candidates.
+
+The function receives a candidate's 0-based number
+and should return a string.
+See `company-show-quick-access' for more details."
+  :type 'function)
+
+(defun company-quick-access-hint-key (candidate)
+  "Return a quick-access key for the CANDIDATE number.
+This is a default value of `company-quick-access-hint-function'."
+  (if company-show-numbers-function
+      (funcall company-show-numbers-function (1+ candidate))
+    (format "%s"
+            (if (< candidate (length company-quick-access-keys))
+                (nth candidate company-quick-access-keys)
+              ""))))
 
 (defcustom company-selection-wrap-around nil
   "If enabled, selecting item before first or after last wraps around."
@@ -719,8 +827,7 @@ asynchronous call into synchronous.")
     (define-key keymap "\C-w" 'company-show-location)
     (define-key keymap "\C-s" 'company-search-candidates)
     (define-key keymap "\C-\M-s" 'company-filter-candidates)
-    (dotimes (i 10)
-      (define-key keymap (read-kbd-macro (format "M-%d" i)) 'company-complete-tooltip-row))
+    (company-keymap--bind-quick-access keymap)
      keymap)
   "Keymap that is enabled during an active completion.")
 
@@ -2285,8 +2392,7 @@ each one wraps a part of the input string."
     (define-key keymap "\C-s" 'company-search-repeat-forward)
     (define-key keymap "\C-r" 'company-search-repeat-backward)
     (define-key keymap "\C-o" 'company-search-toggle-filtering)
-    (dotimes (i 10)
-      (define-key keymap (read-kbd-macro (format "M-%d" i)) 'company-complete-tooltip-row))
+    (company-keymap--bind-quick-access keymap)
     keymap)
   "Keymap used for incrementally searching the completion candidates.")
 
@@ -2549,7 +2655,7 @@ When called interactively, uses the last typed digit, stripping the
 modifiers and translating 0 into 10, so `M-1' inserts the first visible
 candidate, and `M-0' insert to 10th one.
 
-To show hint numbers beside the candidates, enable `company-show-numbers'."
+To show hint numbers beside the candidates, enable `company-show-quick-access'."
   (interactive
    (list (let* ((type (event-basic-type last-command-event))
                 (char (if (characterp type)
@@ -2560,6 +2666,18 @@ To show hint numbers beside the candidates, enable `company-show-numbers'."
                 (number (- char ?0)))
            (if (zerop number) 10 number))))
   (company--complete-nth (1- number)))
+
+(defun company-complete-quick-access (row)
+  "Insert a candidate visible on a ROW matched by a quick-access key binding.
+See `company-quick-access-keys' for more details."
+  (interactive
+   (list (let* ((event-type (event-basic-type last-command-event))
+                (event-string (if (characterp event-type)
+                                  (string event-type)
+                                (error "Unexpected input"))))
+           (cl-position event-string company-quick-access-keys :test 'equal))))
+  (when row
+    (company--complete-nth row)))
 
 (defun company--complete-nth (row)
   "Insert a candidate visible on the tooltip's zero-based ROW."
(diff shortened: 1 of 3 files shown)


yugaego:
Regarding setting up custom keybindings, it seems [something close to this commit](https://github.com/emacs-mirror/emacs/commit/b5416d971a720e1dfb9aedd8e7d23f24b1ea5325#diff-d3fadd7341a7280eb54af943d0287340c3a70c066af22463d70bb4b30e0602e7) may be referenced for some inspiration. We wouldn't probably need that sophisticated arguments, but at least I'll try to fix the above mentioned issue (**No 1.**) with a similar approach.

yugaego:
Just added the first attempt to actually pass user-wanted quick-access keys customization to the keymaps.
Drawbacks:
- search map still isn't re-initialized.
- a user would have to specifically use `(custom-set-variables '(company-quick-access-keys '("a" "o" "e" "u" "i" "d" "h" "t" "n" "s")))`. It was [discussed on Stackoverflow](https://emacs.stackexchange.com/questions/55021/how-to-run-a-function-when-setting-a-defcustom-option).

dgutov:
Hi Y.E.!

> Is the direction overall fine with you?

It's looking fine. :+1: But see some inline comments later.

> What direction should I look at to fix this issue?

Having a setter on this option is step one. Step two is to make it explicit to the user that they should change the variable either using the Custom interface, or by calling `custom-set-variables` in their init script, like you mentioned in the last comment.

It's generally known that if an option has a `:set` attribute, changing it via `setq` might not have the desired effect. Some docstrings make that explicit, though. See for example `frame-background-mode`.

> (I'm a bit cheating here not trying to figure it out by myself, but the cognitive load had temporarily overheated me.)

It's totally okay to ask for guidance. Especially when an issue asks for a broader set of knowledge, rather than a focused solution (like some well-defined problem that needs one to sit down, think hard and write down the algo).

> Probably, one of the options is to allow setting key sequences and parse non-modifier part for output. What's your opinion on it?

That doesn't sound too bad.

But also consider simply having a separate option to specify the modifier. I'm pretty sure everyone is going to use just one modifier for the whole set. The possible values could be `meta`, `ctrl`, `hyper` and `super`.

> Considering current implementation approach, should the values of company-quick-access-keys be restricted somehow (f.i. to be in a range of characters 0 to 127)

Why not. Adding validation to the `defcustom` element is straightforward enough. We could also provide two sets of predefined values: 0 to 9 and asdfghjkl;. Those two should satisfy 99% of people, and the rest could deal with failures with a little experimenting, even if validation doesn't catch some problem in the configuration.

> Current implementation suggests company-quick-access-keys to fully replace company-show-numbers-function, which - supposedly - isn't widely used. Still, I'm wondering how high the chances are it could be an undesirable change?

If their signatures are the same, it's better to keep the old function as obsolete alias for a while. This way we don't break the existing configurations unnecessarily, letting the users upgrade on their own pace.

company.el
```diff
@@ -645,18 +645,48 @@ commands in the `company-' namespace, abort completion."
...
+  '("1" "2" "3" "4" "5" "6" "7" "8" "9" "0")
+  "A list of quick-access keys.
+Quick access is bound to each of `M-VALUE' key sequence.
+If `company-show-quick-access' is non-nil, a quick-access column
+shows the configured quick-access keys beside candidates."
+  :type '(repeat string)
+  :initialize #'custom-initialize-default
+  :set #'(lambda (sym key)
+           (custom-set-default sym key)
+           (company-active-map-builder)
```
dgutov, line 685: It would be better for this setter not to recompute the whole keymaps, and instead only modify the existing value of the keymap. Because otherwise we can have a user who has changed the keys in, say, `company-active-map`, then they change `company-quick-access-keys` through Customize - and boom, all their previous changes are gone.

This way you won't need to call `company-enable-overriding-keymap` either.

How to do that:

- Have a function like `company--add-quick-access-keys (keymap)`, and call it from both keymap definitions.
- This setter would call it on both keymaps, if those variables are `boundp` (which they won't be the first time this defcustom is loading).
- Also have a function `company--remove-quick-access-keys (keymap)`, which would remove the previous bindings. You would call it before `custom-set-default`, when `company-quick-access-keys` is still set to the previous value (or to no value at all). A `boundp` check is probably necessary there as well.

dgutov: Alternatively, we could avoid this whole setter business and say that changing this variable won't have any effect if the package is already loaded (there are examples of such variables in Emacs as well, IIRC). And suggest, perhaps, to instead call some new function which would perform the necessary modification.

The function could take the values of the new options as argument: the modifier, and the list of keys. It would look at the current value of the var to see which keys to remove, and then change the values in the end.

This might be easier to implement, though the other approach is probably a little more user-friendly.

yugaego:
I appreciate your feedback and the valuable suggestions! It's interesting to try all these options, so it will take me awhile. (Unless I find an atomic change to suggest, which supposedly may happen in relation to the numbers/quick-access renaming.)

yugaego:
Hi Dmitry,

I'm still working on the changes. There were many subtle issues in the previously suggested changes, so I had to experiment more to find an approach that felt more consistent to me. It is better now, though it's still a WIP. Let me know if you have any strong objections.
The changes not covered in the code but mentioned here are still on my list.

A couple of questions:
- Could it be beneficial and still convenient for end-users using chars instead of strings declaring `company-quick-access-keys`?  Such as `(?1, ?2, ?3, ... ?9, ?0)`  or `(?a, ?b, ?c)`.
- Do I get it right `company-quick-access-modifier` should desirably use symbols 'meta, 'control etc?   (I initially followed this way by failed to bind `(vector meta key)` on many attempts). So it would rather look like:
```
(defcustom company-quick-access-modifier 'meta
  "A modifier to use as a part of quick-access key sequences.
See also `company-quick-access-keys'."
  :type '(choice (const :tag "Meta" meta)
                 (const :tag "Control" control)
                 (const :tag "Shift" shift)
                 (const :tag "Hyper" hyper)
                 (const :tag "Super" super)
                 (const :tag "Alt" alt)))
```

Thanks!

dgutov:
Hi Y.E.!

> A couple of questions:

Both sound fine to me.

I'm not sure if going the `(vector meta key)` way is the best approach. E.g., `(kbd "M-b")` evaluates to `[134217826]` over here.

But the exact way to compute the bindings is a secondary concern to how the preferences are set.

yugaego:
> But the exact way to compute the bindings is a secondary concern to how the preferences are set.

👍 Back to the initialization then. Thanks for the feedback!

yugaego:
Everything discussed is covered (or considered) now, so the code is ready for a thorough review. The commit includes quite many updates, but they all seem to fit into one category and probably wouldn't be easily split into smaller chunks.

There's one issue I still couldn't resolve cleanly: utility functions `company-custom--set-quick-access` and `company-keymap--bind-quick-access` had to be specifically put above their invocations, to get rid of *Symbol's function definition is void* error. Couldn't do anything better at this stage with it, so leaving it up for your consideration.

Looking forward to working on the improvements!

yugaego:
And one more thing I just recalled should be taken a note of: with the current implementation, a user can potentially overwrite a `Company`'s keybindings, especially on configuring use of `Control` modifier with letters keys.

company.el
```diff
@@ -655,19 +655,109 @@ commands in the `company-' namespace, abort completion."
...
+    (company-keymap--bind-quick-access company-search-map)))
+
+(defcustom company-quick-access-keys '("1" "2" "3" "4" "5" "6" "7" "8" "9" "0")
+  "Character strings used as a part of quick-access key sequences.
+To change this value without Customize interface, use `customize-set-variable'.
+
+To change a quick-access key sequences modifier, customize
+`company-quick-access-modifier'.
+
+If `company-show-quick-access' is enabled, show quick-access hints
```
dgutov, line 677: Rather than "enabled" (which would be appropriate when talking about a minor mode), when referring to a variable we usually say "non-nil".

company.el
```diff
@@ -655,19 +655,109 @@ commands in the `company-' namespace, abort completion."
...
+  (when (boundp 'company-active-map)
+    (company-keymap--bind-quick-access company-active-map))
+  (when (boundp 'company-search-map)
+    (company-keymap--bind-quick-access company-search-map)))
+
+(defcustom company-quick-access-keys '("1" "2" "3" "4" "5" "6" "7" "8" "9" "0")
+  "Character strings used as a part of quick-access key sequences.
+To change this value without Customize interface, use `customize-set-variable'.
+
+To change a quick-access key sequences modifier, customize
```
dgutov, line 674: a -> the

company.el
```diff
@@ -655,19 +655,109 @@ commands in the `company-' namespace, abort completion."
...
+beside the candidates."
+  :set #'company-custom--set-quick-access
+  :type '(choice
+          (const :tag "Digits" ("1" "2" "3" "4" "5" "6" "7" "8" "9" "0"))
+          (const :tag "QWERTY home row" ("a" "s" "d" "f" "g" "h" "j" "k" "l" ";"))
+          (const :tag "Dvorak home row" ("a" "o" "e" "u" "i" "d" "h" "t" "n" "s"))
+          (repeat :tag "User defined" string))
+  :package-version '(company . "0.9.14"))
+
+(defcustom company-quick-access-modifier "M"
```
dgutov, line 687: Not what we discussed, but works for me. :ok_hand:

company.el
```diff
@@ -3256,6 +3345,15 @@ If SHOW-VERSION is non-nil, show the version in the echo area."
                       (company-space-string fancy-margin-width))
               'face 'company-tooltip))
 
+(defun company-tooltip--format-quick-access-hint (row selected)
+  "Format a quick-access hint for outputting on a tooltip's ROW.
+Value of SELECTED determines an added face."
```
dgutov, line 3350: an -> the

(last occurrence only)

dgutov:
Hi again!

Everything looks good, with just a few small concerns.

> utility functions company-custom--set-quick-access and company-keymap--bind-quick-access had to be specifically put above their invocations

This is fine, I don't think it's possible to avoid this.

> a user can potentially overwrite a Company's keybindings, especially on configuring use of Control modifier with letters keys

We could add a check in `company-keymap--unbind-quick-access` that the keys are bound `company-complete-quick-access` before we remove the bindings. It's not very urgent or crucial, though. Might as well leave it for a later.

dgutov:
Here's a sort-of blocker:

The obsoletion message of `company-show-numbers-function` says that the possible arguments have changed. That would mean that anybody who customized this variable in the past to a function whey wrote themselves, would have to change it, wouldn't they? That makes it not a real rename.

Can we adjust the argument values so that they match what `company--show-numbers` received? UUIC this shouldn't be difficult.

company.el
yugaego: Do you mean [symbols usage](https://github.com/company-mode/company-mode/pull/1115#issuecomment-868698762)? Unfortunately, I couldn't make them work, so just used the approach that worked. (yeah, my ELisp-fu isn't strong enough yet.)

As for the usage of characters for the keys setup (instead of strings), after thinking more about it, I concluded that they seemed to be potentially more confusing (especially to newcomers). 

And as for the discussed earlier quick-access keys validation, it worked fine by default when assigning keys from Customize: the bindings are re-added and if something goes wrong the error is shown in the minibuffer. That seemed to be enough to me.

That's all I could think of regarding "Not what we discussed". Let me know if that's something else to avoid any miscommunication.

yugaego:
Hi Dmitry,

> We could add a check in company-keymap--unbind-quick-access that the keys are bound company-complete-quick-access before we remove the bindings. It's not very urgent or crucial, though. Might as well leave it for a later.

Appears I dislike this drawback strong enough to work on it right away, so will do it.

> The obsoletion message of company-show-numbers-function says that the possible arguments have changed.
> Can we adjust the argument values so that they match what company--show-numbers received?

Yes, that's an issue. I had hoped it might be a neglectable one and would allow being consistently 0-based with this and recent changes (and programming field habits).

So, relatively to this PR code, for `company-quick-access-hint-function` to still accept functions that handle 1-based numbers we'd:
   - Increment a passed argument (or similar) for the three invocations of `company-quick-access-hint-function`.
   - Change `company-quick-access-hint-key` (a.k.a. `company--show-numbers`) to accept 1-based candidates.


I've been thinking on another workaround, but it would require to distinguish somehow whether a customized function accepts 0- or 1- based candidates. The *raw* idea:
  - Declare `company--show-numbers` obsolete without aliasing. Remove it much later.
  - Check on `company-quick-access-hint-function` invocation whether  `company-quick-access-hint-key` or `company--show-numbers`  is called.
  (Naturally, it wouldn't work for functions with other names, as already mentioned.)

Might be you could think of something in the spirit, but workable? (With an adapter-function maybe.)

_[Sounds like too much hassle for probably very rarely customized `company-show-numbers-function`
but those 0-based indexes have enchanted me.]_

Best.

company.el
```diff
@@ -3609,27 +3707,32 @@ Delay is determined by `company-tooltip-idle-delay'."
...
-                               '(face company-echo-common) comp))
+        (let ((beg 0)
+              (end (string-width (or company-common ""))))
+          (when (< numbered qa-keys-len)
+            (let ((qa-hint
+                   (format "%s: " (funcall
+                                   company-quick-access-hint-function
+                                   numbered))))
+              (setq beg (string-width qa-hint)
+                    end (+ beg (string-width (or company-common ""))))
```
yugaego, line 3729: I noticed a duplicated evaluation here (already done on L3722), it'll be fixed with the next push.

company.el
dgutov: > Do you mean symbols usage? Unfortunately, I couldn't make them work, so just used the approach that worked. (yeah, my ELisp-fu isn't strong enough yet.)

That's the one, yes.

It's not a big problem, but in general I like to follow existing examples (e.g. `cua-rectangle-modifier-key`) in such situations, if only to reduce arbitrary differences between packages.

The code examples out there construct binding vectors with it, which I agree is non-trivial and requires knowledge of the respective data structures (which the use of `kbd` helpfully avoids). But we can use a simple alist to translate from one set of values to another (100% untested code):

```el
(defun company-keymap--bind-quick-access (keymap)
  (let ((modifier-char (company-keymap--quick-access-modifier)))
    (dolist (key company-quick-access-keys)
      (define-key keymap
        (company-keymap--kbd-quick-access modifier-char key)
        #'company-complete-quick-access))))

(defun company-keymap--quick-access-modifier ()
  (let ((modifier-char (assoc-default company-quick-access-modifier
                                      '((meta . "M")
                                        (control . "C")
                                        (shift . "S")
                                        (hyper . "h")
                                        (super . "s")))))
    (warn "company-quick-access-modifier value unknown: %S" company-quick-access-modifier)
    modifier-char))

(defun company-keymap--unbind-quick-access (keymap)
  (let ((modifier-char (company-keymap--quick-access-modifier)))
    (dolist (key company-quick-access-keys)
      (define-key keymap (company-keymap--kbd-quick-access modifier-char key) nil))))

(defun company-keymap--kbd-quick-access (key)
  (kbd (format "%s-%s" company-quick-access-modifier key)))
```

dgutov:
> I had hoped it might be a neglectable one and would allow being consistently 0-based with this and recent changes (and programming field habits).

It can be a problem if the user switches between Emacs versions. E.g. between the last release and the development snapshot. And they'll have to discover the necessary change first. Perhaps the two users who actually take advantage of this possibility won't mind too terribly, but it's better to make the transition smoother if we can.

> Increment a passed argument (or similar) for the three invocations of company-quick-access-hint-function.

I was thinking of something along these lines, yes.

> (Naturally, it wouldn't work for functions with other names, as already mentioned.)

And those are the main problematic cases which I'm thinking of.

> Might be you could think of something in the spirit, but workable? (With an adapter-function maybe.)

Well, we could do the transition in two steps. Do the rename in this version, and the indexing change in the next version.

Is that really worth it, though? Whether to start numbering from 0 or from 1, is largely an arbitrary choice. Even if most programming languages these days have converged on one option.

dgutov:
> Might be you could think of something in the spirit, but workable? (With an adapter-function maybe.)

Another option might go like:

* Keep `company-show-numbers-function` variable. Set it to nil by default.
* Add the new `company-quick-access-hint-function` variable and use it. The default value of this new variable would check whether `company-show-numbers-function` is non-nil, and if so, call that value (with appropriate argument). Otherwise, do what it is expected to do by default.
* Add this below their definitions: `(make-obsolete-variable 'company-show-numbers-function "use company-quick-access-hint-function, but adjust the expected values appropriately.")`.

yugaego:
Cool, thanks a lot! I'll wrap it all up and update the branch.

yugaego:
Dmitry, the code is ready for the next round of review/updates.

Changes:

#### Added quick-access bindings validation.
While working on already bound key sequences validation, I noticed it seemed strange - saying it mildly - to provide default quick-access bindings that would break default `Company` keymaps.

`cua-rectangle-modifier-key` example without `Control` key support suggests a solution to this issue, so in this code version `Control` is dropped from the list of configurable quick-access modifiers.

Now - if I'm not missing anything - the only default failing combination would be `M-n`.
So I've commented out Dvorak home row keys with the note to un-comment then on the currently deprecated `M-n` binding removal.
(For my own setup I now use 8 candidates limit, up to the `M-n` Dvorak home row quick-access binding.)

#### Quick-access modifier customized by setting the symbol.
Fallback to "M" if an unknown value is attempted to be configured.
(Fixes an unpleasant behavior with an 'unknown value' more warnings raised on re-setting.)

#### Deprecate (without aliasing) `company-show-numbers-function` and `company--show-numbers`.
The solution you suggested seemed to be an elegant one, so now `company-quick-access-hint-key` "funcalls" `company-show-numbers-function` value, if it is non-nil.

##### NEWS rewritten.
May be not perfect though and has too many same PR links.

##### Minor doc strings fixes.

##### Removed `echo-frontend` duplicated line.

--
So, let's see if it works fine this time, or needs more adjustments.

dgutov:
Looks good, thank you for being thorough. :+1: 

I'll add some final adjustments later myself. In short:

* We can shorten the NEWS entry (yes, too many links).
* Let's not support the `alt` value in `company-quick-access-modifier`. Because it's a different Alt, not the same key we know as Meta, and that can be confusing for 99.9% of our users. We can add that later if somebody actually finds a use for that.

dgutov:
> So I've commented out Dvorak home row keys with the note to un-comment then on the currently deprecated M-n binding removal.
(For my own setup I now use 8 candidates limit, up to the M-n Dvorak home row quick-access binding.)

I wouldn't say it's a problem: the `M-n` binding is scheduled to be removed anyway. It's only there still to work as a gentle reminder to migrate, rather than have users encounter broken behavior in the middle of their work if they upgraded without reading NEWS.

yugaego:
🥳 🥳 🥳

PR #1117, opened 2021-06-12
Clarify company-search-keypad function role

yugaego:
Related issue: #273.

- Adds a doc string: to clarify the function's use-case, and to eliminate the need in looking it up in the history.
- Uses `kbd`: I assume since the Company's minimum required version of Emacs is now 25.1 it's fine to do so for better uniformity.

The suggestions are, as always, welcome.

### DIFF
--- a/company.el
+++ b/company.el
@@ -2139,6 +2139,14 @@ each one wraps a part of the input string."
       (cl-incf i))))
 
 (defun company-search-keypad ()
+  "Prevent disabling of `company-search-mode' on a keypad number press.
+
+When in search mode, keypad numbers presses cause invocation
+of `company-search-other-char', which in turn disables search mode.
+To prevent this behavior, `company-search-map' binds keypad numbers
+to `company-search-keypad', which performs the explicit
+continuation of the search. As a result of the latter action,
+keypad input is treated as numbers to print."
   (interactive)
   (let* ((name (symbol-name last-command-event))
          (last-command-event (aref name (1- (length name)))))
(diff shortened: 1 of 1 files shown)


company.el
```diff
@@ -2254,7 +2262,7 @@ each one wraps a part of the input string."
       (define-key keymap (vector i) 'company-search-printing-char)
       (cl-incf i))
     (dotimes (i 10)
-      (define-key keymap (read (format "[kp-%s]" i)) 'company-search-keypad))
+      (define-key keymap (kbd (format "<kp-%d>" i)) 'company-search-keypad))
```
dgutov, line 2265: :+1:

company.el
```diff
@@ -2139,6 +2139,14 @@ each one wraps a part of the input string."
       (cl-incf i))))
 
 (defun company-search-keypad ()
+  "Prevent disabling of `company-search-mode' on a keypad number press.
```
dgutov, line 2142: The first line of a docstring should be a (short) complete sentence which describes what the command does.

The rest can explain further details, including, like in this example, the reasons for the command's existence.

As far as I'm concerned, this explanation can go in a comment somewhere, for example inside the `company-search-map`'s definition above the line where `'company-search-keypad` is bound.

But if you want to add it in the docstring, let's make the first line something like:

> Handle keypress on the keypad and append the typed key to the search string.

`company-search-printing-char` could probably use a docstring as well.

yugaego:
Oops, branch renaming didn't work quite well in the UI. The follow-up is in the commit 60b01bf. 
And seems creating a new PR would be the cleanest option now.

PR #1118, opened 2021-06-12
Rename and document company-complete-number

yugaego:
[Initially discussed in #1104.](https://github.com/company-mode/company-mode/pull/1104#issuecomment-851591903)

Main points and questions:

- I could eventually test keypad behavior (hurray!) and got these results on Company's master:
  - `M-<kp-%d>` completes candidates both with *active-map* and *search-map*.
  - `<kp-%d>` does not. Since *search-map* explicitly binds `<kp-%d>` to print numbers, I suppose it'd be quite natural to intent `<kp-%d>` to work the same way with *active-map*.
  - Related commit: 43756857.

- #177 evil mode compatibility is still a question. (But `company-complete-number` is still in the code.)

- In the code, I had to put `company--add-quick-access-keys` function definition before its invocation.
With the definition located after invocation, I've spent hours trying to make it work with funcall/quoting, but got either *a la* "Symbol's function definition is void" errors or no bindings. I suppose the underlying concepts are not fully clear to me yet. Could you hint me how to fix this issue?

- If there's no restricting conventions or use-cases, maybe rather use `company--bind-quick-access-keys`? (With probably `company--unbind-quick-access-keys` counterpart according to what you'd suggested before.)

- `company--add-quick-access-keys` doc string is planned to be extended after keys customization is fully implemented. Actually, this commit could be stabilized but postponed for merge, used as a (at least mostly) certain foundation for the next layers of quick-access related changes. Please let me know what would be a preferred approach.

- Moved `company-active-map` to `;;; keymaps` section to have related definitions closer to each other. (Could be a separate commit for better clarity of the changes history.)

Thanks!

### DIFF
--- a/company.el
+++ b/company.el
@@ -720,7 +720,7 @@ asynchronous call into synchronous.")
     (define-key keymap "\C-s" 'company-search-candidates)
     (define-key keymap "\C-\M-s" 'company-filter-candidates)
     (dotimes (i 10)
-      (define-key keymap (read-kbd-macro (format "M-%d" i)) 'company-complete-number))
+      (define-key keymap (read-kbd-macro (format "M-%d" i)) 'company-complete-tooltip-row))
      keymap)
   "Keymap that is enabled during an active completion.")
 
@@ -2067,7 +2067,7 @@ prefix match (same case) will be prioritized."
                                           company-complete
                                           company-complete-common
                                           company-complete-selection
-                                          company-complete-number)
+                                          company-complete-tooltip-row)
   "List of commands after which idle completion is (still) disabled when
 `company-begin-commands' is t.")
 
@@ -2286,7 +2286,7 @@ each one wraps a part of the input string."
     (define-key keymap "\C-r" 'company-search-repeat-backward)
     (define-key keymap "\C-o" 'company-search-toggle-filtering)
     (dotimes (i 10)
-      (define-key keymap (read-kbd-macro (format "M-%d" i)) 'company-complete-number))
+      (define-key keymap (read-kbd-macro (format "M-%d" i)) 'company-complete-tooltip-row))
     keymap)
   "Keymap used for incrementally searching the completion candidates.")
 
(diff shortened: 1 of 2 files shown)


company.el
```diff
@@ -2536,12 +2545,16 @@ character, stripping the modifiers.  That character must be a digit."
                         (car (last (string-to-list (symbol-name type))))))
                 (n (- char ?0)))
            (if (zerop n) 10 n))))
+  (company-complete-nth (cl-decf n)))
```
dgutov, line 2548: You probably want `(1- n)` rather than `(cl-decf n)`. The latter is side-effecting, and there's no point in that here.

dgutov:
Hi Y.E., sorry for the delay.

There is a particular thing that sticks out here: the use of anonymous functions for key bindings set up by `company--add-quick-access-keys`. Those are a bit problematic, for example because they make the resulting keymap harder to examine.

What do you think about after still handling the exact quick-access mappings inside the `(interactive ...)` form, like `company-complete-number` does currently, and along the latest change to `company-search-printing-char`?

I get the idea about specifying the argument together with the binding and not worrying about it later, but doing it the current way has its benefits. For example, we can keep `<kp-%d>` bindings do the quick-access action, if the user sets up those bindings manually (that feature was added in https://github.com/company-mode/company-mode/commit/4375685, though I can't find the corresponding feature request in the bug tracker; perhaps it was made over email directly).

So... we can still rename-with-an-alias `company-complete-number` to `company-complete-nth`, and even renumber it to start addressing with 0 not 1. But keep the `interactive` form which handles the current known cases.

And then we can add `company-complete-quick-access` which will only handle chars from the customized list. In its interactive form, it will do something like `(cl-position (event-basic-type last-command-event) company-quick-access-keys)` to get the index, and then pass the number to `company-complete-nth`.

WDYT?

company.el
yugaego: Gladly. Used `cl-decf` only because it was there already.

yugaego:
> because they make the resulting keymap harder to examine.

I agree, the question marks are not informative at all and rather irritating.

> What do you think about after still handling the exact quick-access mappings inside the (interactive ...) form, like company-complete-number does currently, and along the latest change to company-search-printing-char?

It all makes sense, so I'm fine with it (and the rest of the suggestions). Will work with the code and see how it works out.

yugaego:
Hi Dmitry,
Here's what I've got after working with the code.

These (all?) important points supported:
  - No backward compatibility breaking changes.
  - Names and documentation better (I think) reflect what the function does.
    - Aligns well with the renaming 'numbers' to 'quick-access' term.
  - Extracting `company--complete-nth` is not necessary, but prepares addition of `company-complete-quick-access`.

dgutov:
Looking good, thanks! :+1: 

I imagined the layout where `company--complete-nth` would be basically inlined in `company-complete-tooltip-row`, but this way works too, and it's probably more explicit when we use it the new command later.

yugaego:
Good work on your side too :-)  
(and makes me want to enable code auto-formatting.)

FYI: I just had a couple more email exchanges with the FSF clerk, so moving forward with the process.

dgutov:
> (and makes me want to enable code auto-formatting.)

There is https://github.com/Malabarba/aggressive-indent-mode, which some people swear by. Seems a bit too much for me, though.

> FYI: I just had a couple more email exchanges with the FSF clerk, so moving forward with the process.

:+1:

yugaego:
> aggressive-indent-mode

Thanks, will take a look!

BTW flymake worked out very nicely so far for me, thank you for suggesting it.

yugaego:
@dgutov , good news: the process of the copyright assignment has been complete for me.

dgutov:
Terrific, welcome to the club! :tada:

PR #1119, opened 2021-06-14
Refactor search-mode keypad key press workaround

yugaego:
Accidental follow-up of  #1117.
Relates to #273.

I've been thinking over your feedback and came to an idea that maybe removal of `company-search-keypad` could be an even nicer move.  What do you think?

(I've got no good idea for `company-search-printing-char` doc string so far.)


dgutov:
Works for me. Thanks!

PR #1120, opened 2021-06-14
Fix width of tooltip offset lines number display

yugaego:
The issue is visible with `(setq company-tooltip-offset-display 'lines)` and non-nil `company-format-margin-function`.
BTW, this `'lines` feature is a very nice one.

**before**
lines-before
**after**
lines-after

### DIFF
--- a/company.el
+++ b/company.el
@@ -3172,7 +3172,7 @@ If SHOW-VERSION is non-nil, show the version in the echo area."
           (numbered (if company-show-numbers 0 99999))
           new)
       (when previous
-        (push (company--scrollpos-line previous width) new))
+        (push (company--scrollpos-line previous width left-margin-size) new))
 
       (dotimes (i len)
         (let* ((item (pop items))
@@ -3199,7 +3199,7 @@ If SHOW-VERSION is non-nil, show the version in the echo area."
                 new)))
 
       (when remainder
-        (push (company--scrollpos-line remainder width) new))
+        (push (company--scrollpos-line remainder width left-margin-size) new))
 
       (cons
        left-margin-size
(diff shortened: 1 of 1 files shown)


company.el
```diff
@@ -3218,10 +3218,10 @@ If SHOW-VERSION is non-nil, show the version in the echo area."
                   'company-scrollbar-fg
                 'company-scrollbar-bg)))
 
-(defun company--scrollpos-line (text width)
+(defun company--scrollpos-line (text width right-margin-width)
```
dgutov, line 3221: Perhaps `left` rather than `right`?

Looks good otherwise.

yugaego: Yes, I thought so initially and the initial implementations used the following approaches, none of which worked fine in my opinion. The current solution has its drawback too (confusing change of names), but reflects correctly the use of the width-value: we pass left-margin-width to be used as a right-margin-width. I'd be happy to improve relatively to the described issues, but nothing better comes to my mind at the moment.
1. UI didn't look nice.
```
(concat (company-space-string left-margin-width)
                       (company-safe-substring text 0 width)
                       (company-space-string company-tooltip-margin)))
```
2. Lost flexibility of configuring the right margin specifically (probably in the future).
```
(concat (company-space-string company-tooltip-margin)
                       (company-safe-substring text 0 (+ width left-margin-width)))
```
3. Confusingly looking internals of `company--scrollpos-line` function: why do we add left-margin to the right side of the line?
```
(concat (company-space-string company-tooltip-margin)
                       (company-safe-substring text 0 width)
                       (company-space-string left-margin-width)))
```

dgutov: Fair enough, thank you.

We can rename it to `extra-margin-width` sometime later, I guess.

yugaego: Sure. And thank you for maintaining the packages!

PR #1122, opened 2021-06-16
Silence compiler

tarsius:
```
lib/company/company.el:3221:1: Warning: Lexical argument
    shadows the dynamic variable right-margin-width
```

### DIFF
--- a/company.el
+++ b/company.el
@@ -3218,10 +3218,10 @@ If SHOW-VERSION is non-nil, show the version in the echo area."
                   'company-scrollbar-fg
                 'company-scrollbar-bg)))
 
-(defun company--scrollpos-line (text width right-margin-width)
+(defun company--scrollpos-line (text width fancy-margin-width)
   (propertize (concat (company-space-string company-tooltip-margin)
                       (company-safe-substring text 0 width)
-                      (company-space-string right-margin-width))
+                      (company-space-string fancy-margin-width))
               'face 'company-tooltip))
 
 ;; show


company.el
```diff
@@ -3218,10 +3218,10 @@ If SHOW-VERSION is non-nil, show the version in the echo area."
                   'company-scrollbar-fg
                 'company-scrollbar-bg)))
 
-(defun company--scrollpos-line (text width right-margin-width)
```
dgutov, line 3221: Oh, right.

But could you rename it to e.g. `fancy-margin-width`, please?

Because "margin width", in general, is covered by the variable `company-tooltip-margin`.

tarsius: done

yugaego:
What would be a recommended practice to avoid similar issues in future? The ones that come to my mind require to memorize an additional step: 
- re-check each variable name with `C-h v`,
- recompile changed file(s).
Could probably compilation stage be added to the package tests or github actions? (I'm not proficient with any of the mentioned, so just guessing.) Anyway, some automation might help I suppose. It should be possible to configure something locally, too, but that would not prevent other commits from similar issues.

company.el
dgutov: Thanks!

tarsius:
I use my [`auto-compile`](https://github.com/emacscollective/auto-compile), which compiles when saving, so I am immediately informed about mistakes I have just made. It is not 100% bullet-proof as some issue only become visible if you compile in a fresh environment, but it goes pretty far.

dgutov:
`flymake-mode` can also help. Here's relevant setup from my config:

```el
(add-hook 'emacs-lisp-mode-hook
          (lambda ()
            (remove-hook 'flymake-diagnostic-functions 'elisp-flymake-checkdoc t)
            (flymake-mode)))
```

But then one has to check `M-x flymake-show-diagnostics-buffer` from time to time.

Absence of byte-compilation warnings is not a hard requirement in my projects (reaching 0 warnings can be a pain sometimes), but since we apparently have reached that state here, we could indeed add a check to the build pipeline.

yugaego:
Thank you!

PR #1143, opened 2021-07-21
Add company-tooltip-offset docstring

yugaego:
The variable `company-tooltip-offset` was confusing when I was working on one of the previous changes.
An [explanation](https://github.com/company-mode/company-mode/pull/1104#issuecomment-853479362) was provided at that time. This PR adds a docstring to clarify the variable's role.

Additionally, may I suggest the removal of the following variables?
(I understand we declare them here and them assign default values later with `defvar-local` but still, it looks like as a not-needed duplication to my eye, that forced me double think why it is so and where should I put the docstring).

```
(defvar company-pseudo-tooltip-overlay)
(defvar company-tooltip-offset)
```

It's git-logged these `defvar`s were added to fix compilation errors, but I got no compilation errors after removing them (possibly I didn't reproduce some other conditions).

Optionally, could corresponding `defvar-local` be re-located, replacing `defvar`s under the question?
Thus these variables would be declared/assigned a value at once, but earlier in the code.

Please let me know if I miss something.

Thank you.

### DIFF
--- a/company.el
+++ b/company.el
@@ -2933,7 +2933,10 @@ If SHOW-VERSION is non-nil, show the version in the echo area."
 
 (defvar-local company-pseudo-tooltip-overlay nil)
 
-(defvar-local company-tooltip-offset 0)
+(defvar-local company-tooltip-offset 0
+  "Current scrolling state of the tooltip.
+Represented by the index of the first visible completion candidate
+from the candidates list.")
 
 (defvar-local company--tooltip-current-width 0)
 


dgutov:
> Additionally, may I suggest the removal of the following variables?

If I remove them and save the buffer, I get these flymake warnings right away:

```
 2534  14 warning  reference to free variable ‘company-pseudo-tooltip-overlay’
 2684  34 warning  reference to free variable ‘company-tooltip-offset’
```

With `flymake-mode` enabled, you should be able to see those with `M-x flymake-show-diagnostics-buffer`.

> Optionally, could corresponding defvar-local be re-located, replacing defvars under the question?

Yeah, that should work. But right now they're neatly residing in the "pseudo-tooltip" section, so probably some reorganization would be needed.

Alternatively, the functions using them could be moved lower. Which could also make sense if those functions are dealing with the pseudo-tooltip.

yugaego:
> I get these flymake warnings right away

Yup, thanks, here there're. I byte-compiled the file without errors but missed those warnings (for some obscure reason).

[So now I'm thinking on how to raise \*Flymake Log\* buffer automatically if new warnings appeared, for a configurable set of files. That's not currently supported by flymake, so that's a chance to check on practice `advice-add` and probably `display-buffer` family.]

 > Alternatively, the functions using them could be moved lower. Which could also make sense if those functions are dealing with the pseudo-tooltip.

This sounds good to me, I'll take a look.

dgutov:
> [So now I'm thinking on how to raise *Flymake Log* buffer automatically if new warnings appeared, for a configurable set of files. That's not currently supported by flymake, so that's a chance to check on practice advice-add and probably display-buffer family.]

That might be annoying: one can often create such warnings temporarily while typing new code, e.g. making a typo or creating a new variable and saving the buffer before writing the part of the code that uses it.

I have a Flymake entry in the mode-line which is possible to consult. Looks like the vanilla configuration even has the error/warning/info numbers in there highlighted with different colors. So you might look at those (the yellow in particular).

A more radical approach would be to add a stage checking for compilation warnings in the CI.

yugaego:
> That might be annoying ... while typing new code

Yeah, you're probably right.

> Flymake entry in the mode-line ... error/warning/info numbers in there highlighted

Yes, I've got those. Probably just need a little bit more time to develop a habit of **always** noticing those marks.

> A more radical approach would be to add a stage checking for compilation warnings in the CI

Automation often helps (unless that's too much of a burden for a task at hand to configure/maintain that automation).
Frankly, I often forget to manually run `make test` before pushing a change (and run tests on the next day as a result)  and supposedly that's common. Unfortunately, our memory is not the most reliable thing in the world, no matter how hard we try.

dgutov:
Looking around, it seems Projectile uses eldev in the CI to make sure byte-compilation produces no warnings:

https://github.com/bbatsov/projectile/blob/master/.github/workflows/test.yml#L34

yugaego:
👍🏻 That example helps to find inspiration to delve into it :-) There're also examples with [ert](https://github.com/search?q=purcell%2Fsetup-emacs+ert+extension%3Ayml&type=Code).
I expect to have some free time again during the following days, so can take a look if you don't already (I see you tested actions earlier).

dgutov:
ert runs tests, but I don't think it has any mode for checking for byte-compilation errors, does it?

Migration to Github Actions would probably be a good thing (Travis has been trying to deprecate its open-source offering for years), but this issue is orthogonal, and so doesn't require that migration. Just like Actions is running a certain script (calling `eldev` with some arguments), it can run on Travis, as defined in `.travis.yml`. Probably with an indirection through make, just for uniformity's sake.

yugaego:
> ert runs tests, but I don't think it has any mode for checking for byte-compilation errors, does it?

Yes, I meant its potential usage for the tests part of the possible checks.

> Migration to Github Actions would probably be a good thing (Travis ...

I don't have an opinion regarding Actions/Travis since I'm well acquainted with Gitlab CI only.
Though Github Actions seem to be quite handy for the trivial stuff (with powerful options available for more specific needs).

yugaego:
I probably should mention that I've already spent some time on configuring GitHub Actions setup, so it's work in progress now. I'm going to post any additional details/considerations with - hopefully - a future PR.

dgutov:
Sounds good :+1:

PR #1147, opened 2021-07-25
Stop completion about tempo if no matching prefix

condy0919:


condy0919:
Before this patch, a single `"` can trigger company to complete about tempo.

condy0919:
I found more characters can lead to completion before this patch, e.g. `>`, `"`, `.` which is annoying when programming.

dgutov:
Do you only see this problem with `company-tempo`?

What's your value of `company-minimum-prefix-length`?

condy0919:
> Do you only see this problem with `company-tempo`?
> 
> What's your value of `company-minimum-prefix-length`?

3, the default.

It can be reproduced by the following company conf.

```
(use-package company
  :ensure t
  :hook (prog-mode . company-mode)
  :bind (:map company-mode-map
         ([remap completion-at-point] . company-complete)
         :map company-active-map
         ("C-s"     . company-filter-candidates)
         ([tab]     . company-complete-common-or-cycle)
         ([backtab] . company-select-previous-or-abort))
  :custom
  (company-idle-delay 0)
  ;; Easy navigation to candidates with M-<n>
  (company-show-quick-access t)
  (company-require-match nil)
  (company-minimum-prefix-length 3)
  (company-tooltip-width-grow-only t)
  (company-tooltip-align-annotations t)
  ;; complete `abbrev' only in current buffer and make dabbrev case-sensitive
  (company-dabbrev-other-buffers nil)
  (company-dabbrev-ignore-case nil)
  (company-dabbrev-downcase nil)
  ;; make dabbrev-code case-sensitive
  (company-dabbrev-code-ignore-case nil)
  (company-dabbrev-code-everywhere t)
  ;; call `tempo-expand-if-complete' after completion
  (company-tempo-expand t)
  ;; Stop annoying me
  (company-etags-use-main-table-list nil)
  ;; No icons
  (company-format-margin-function nil)
  (company-backends '((company-capf :with company-tempo)
                      company-files
                      (company-dabbrev-code company-keywords)
                      company-dabbrev
                      ;; HACK: prevent `lsp-mode' to add `company-capf' back.
                      company-capf)))
```

dgutov:
> 3, the default

Did you check the value at runtime, or did you just look at the config for the answer?

I don't have `use-package` installed, and I don't have any C++ language servers at hand. Can you provide a repro without `use-package` and `lsp-mode`? `company-clang` could be the replacement for `company-capf` in that scenario.

I tried reproducing this problem by copying just the necessary parts of your config and failed. The templates were working, but no trace of the problem from the description.

condy0919:
> > 3, the default
> 
> Did you check the value at runtime, or did you just look at the config for the answer?

Both are 3.

> I don't have `use-package` installed, and I don't have any C++ language servers at hand. Can you provide a repro without `use-package` and `lsp-mode`? `company-clang` could be the replacement for `company-capf` in that scenario.
> 
> I tried reproducing this problem by copying just the necessary parts of your config and failed. The templates were working, but no trace of the problem from the description.

After using `company-clang`, it can be only produced after `".`.

``` elisp
(require 'tempo)
(require 'cc-mode)

(setq company-backends '((company-clang :with company-tempo)
                         company-files
                         (company-dabbrev-code company-keywords)
                         company-dabbrev))
(setq company-idle-delay 0)
(setq company-require-match nil)
(setq company-minimum-prefix-length 3)
(setq company-tempo-expand t)



(defvar c-tempo-tags nil)
(defvar c++-tempo-tags nil)

(defun c-mode-tempo-setup ()
  (tempo-use-tag-list 'c-tempo-tags))
(defun c++-mode-tempo-setup ()
  (tempo-use-tag-list 'c-tempo-tags)
  (tempo-use-tag-list 'c++-tempo-tags))

(tempo-define-template "c-ifndef"
                       '("#ifndef " (P "ifndef-clause: " clause) > n
                         "#define " (s clause) n> p n
                         "#endif // " (s clause) n>
                         )
                       "ifndef"
                       "Insert #ifndef #define #endif directives"
                       'c-tempo-tags)
(tempo-define-template "c-main"
                       '("int main(int argc, char* argv[]) {" > n>
                         p n
                         "}" > n>
                         )
                       "main"
                       "Insert a main function"
                       'c-tempo-tags)

(tempo-define-template "c++-class-equality"
                       '((P "type: " type 'noinsert)
                         (tempo-save-named 'inline (if (y-or-n-p "Inline? ") "inline " ""))
                         (tempo-save-named 'constexpr (if (y-or-n-p "Constexpr? ") "constexpr " ""))
                         (tempo-save-named 'noexcept (if (y-or-n-p "Noexcept? ") "noexcept " ""))
                         (s inline) (s constexpr) "bool operator==(const " (s type) "& lhs, const " (s type) "& rhs) " (s noexcept) "{" > n>
                         p n
                         "}" > n>
                         (s inline) (s constexpr) "bool operator!=(const " (s type) "& lhs, const " (s type) "& rhs) " (s noexcept) "{" > n>
                         "return !(lhs == rhs);" > n>
                         "}" > n>
                         )
                       "eq"
                       "Class equality comparison"
                       'c++-tempo-tags)
(tempo-define-template "c++-class-totally-ordered"
                       '((P "type: " type 'noinsert)
                         (tempo-save-named 'inline (if (y-or-n-p "Inline? ") "inline " ""))
                         (tempo-save-named 'constexpr (if (y-or-n-p "Constexpr? ") "constexpr " ""))
                         (tempo-save-named 'noexcept (if (y-or-n-p "Noexcept? ") "noexcept " ""))
                         (s inline) (s constexpr) "bool operator<(const " (s type) "& lhs, const " (s type) "& rhs) " (s noexcept) "{" > n>
                         p n
                         "}" > n>
                         (s inline) (s constexpr) "bool operator>(const " (s type) "& lhs, const " (s type) "& rhs) " (s noexcept) "{" > n>
                         "return rhs < lhs;" > n>
                         "}" > n>
                         (s inline) (s constexpr) "bool operator<=(const " (s type) "& lhs, const " (s type) "& rhs) " (s noexcept) "{" > n>
                         "return !(lhs > rhs);" > n>
                         "}" > n>
                         (s inline) (s constexpr) "bool operator>=(const " (s type) "& lhs, const " (s type) "& rhs) " (s noexcept) "{" > n>
                         "return !(lhs < rhs);" > n>
                         "}" > n>
                         )
                       "ord"
                       "Class totally ordered comparison"
                       'c++-tempo-tags)

(add-hook 'c++-mode-hook #'c++-mode-tempo-setup)
```

image

dgutov:
> After using company-clang, it can be only reproduced after "., see screenshots below.

So basically after any `.`, right?

If so, I think I understand what's going on.

condy0919:
> > After using company-clang, it can be only reproduced after "., see screenshots below.
> 
> So basically after any `.`, right? Or `::` or `->`, I suppose.
> 
> If so, I think I understand what's going on.

Yes, you're right. All `::` and `->` have the same issue.

dgutov:
The problem:

We have a mechanism that "overrides" the actual prefix length for the returned prefix string, when we want, in certain contexts, to initiate idle completion right away. Then that backend returns a cons `("prefix-string" . t)` from the `prefix` handler.

`company-clang` does that after member access, `lsp-mode` does it in a more generic fashion -- after any character that the language server has designated as a trigger character, and there's usually more of the (that must make the effect worse).

Grouped backends work behind an adapter that combines them together to make them look like a single backend. But there is no good way to combine such hints (or at least none have been implemented so far) that would retain the difference between the backends WRT triggers depending on the context. What it does is simple: if the first backend says "go", completions are collected from all of them.

Not sure what's a good approach to improve this. I suppose the multi-backend adapter in its `candidates` implementation could track whether the current completion is done by explicitly invoking a command, poll each backend's `prefix` again, compare against `company-minimum-prefix-length`, and depending on that, skip the backend's completions. That seems a little involved, though, and might not work with third-party frontends that use company's backends. But it's something to try.

Your particular patch here seems like it will break `M-x company-tempo`, and it would only help in the "empty prefix" case, but not, say, when the prefix is 1 character long and the `company-minimum-prefix-length` has its default value of 3.

dgutov:
Turns out this could be managed inside the framework.

Check out the latest master.

PR #1154, opened 2021-07-30
Merge variables declaration and assignment

yugaego:
Discussed in #1143.

What if to do it this way? I tried to find an approach with the least number of changes. 

`company-pseudo-tooltip-overlay` seem to fit nicely `;; show` subsection of the `;;; pseudo-tooltip`.
`company-tooltip-offset` position is more arguable, though may be justified by the variable definition placed right before its first invocation.

Side question: are there any particular/real reason(s) for not extracting some of the functionality out of `company.el` to separate files? (f.i. starting with echo-frontends).
Based on my own perception, it might be easier to work with smaller files devoted to the specific parts of functionality. (And, potentially, that could make the code look more approachable for fixes/improvements to other programmers).

### DIFF
--- a/company.el
+++ b/company.el
@@ -2526,24 +2526,6 @@ With ARG, move by that many elements."
         (company-set-selection (- company-selection
                                   company-tooltip-limit))))))
 
-(defvar company-pseudo-tooltip-overlay)
-
-(defvar company-tooltip-offset)
-
-(defun company--inside-tooltip-p (event-col-row row height)
-  (let* ((ovl company-pseudo-tooltip-overlay)
-         (column (overlay-get ovl 'company-column))
-         (width (overlay-get ovl 'company-width))
-         (evt-col (car event-col-row))
-         (evt-row (cdr event-col-row)))
-    (and (>= evt-col column)
-         (< evt-col (+ column width))
-         (if (> height 0)
-             (and (> evt-row row)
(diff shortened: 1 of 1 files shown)


dgutov:
Yeah, ok. Let's do it this way, thanks.

dgutov:
> Based on my own perception, it might be easier to work with smaller files devoted to the specific parts of functionality. (And, potentially, that could make the code look more approachable for fixes/improvements to other programmers).

It probably would.

There are reasons not to change the existing structure, though. Most prominently: `vc-annotate` doesn't (yet?) support tracking changes across files, so any piece of code extracted to a new file will be harder to follow the history of. Like, its history will effectively stop where the new file was created.

I'm talking about the command `vc-annotate-revision-previous-to-line`, BTW, bound to `a`. Very handy when researching why something was written one way and not another.

Here's hoping we solve this drawback one day (https://debbugs.gnu.org/cgi/bugreport.cgi?bug=26345).

yugaego:
> There are reasons not to change the existing structure, though. Most prominently: vc-annotate 

Thank you for the feedback, that's an unexpected reason (I guessed that could be something more files-combining related).
Yes, `vc-annotate*` is an amazingly useful feature(s).

>  (https://debbugs.gnu.org/cgi/bugreport.cgi?bug=26345).

Thanks, noted.

PR #1157, opened 2021-08-02
Allow Control quick-access modifier key

### LINKED ISSUE #1151: [Feature] company-quick-access-modifier doesn't support control
I use control key modifier alongside with the numeric keys 1-9 and 0, to choose from the list of 10 choices.

I found out about `company-quick-access-modifier`, and realised this could be customized but only between the `meta`, `super`, `hyper` keys, and the implementation safeguard and forces the modifier to be fallback to `meta` when it doesn't recognize the value of `company-quick-access-modifier`. 

This is all well and good until, me, who would like control key to be part of the options as well.

Best wish,
Inc0n

yugaego:
Closes #1151

As was suggested, the `Control` was added to the `Customize` interface too.
The warnings given in case of an attempted default company keymap overwrite seem to be informative enough and worked smoothly during my checks.

Still, I put this key last in the list, so alternatives would be reviewed first.

### DIFF
--- a/company.el
+++ b/company.el
@@ -692,7 +692,8 @@ See `company-quick-access-keys' for more details."
   :set #'company-custom--set-quick-access
   :type '(choice (const :tag "Meta key" meta)
                  (const :tag "Super key" super)
-                 (const :tag "Hyper key" hyper))
+                 (const :tag "Hyper key" hyper)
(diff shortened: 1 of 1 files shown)


dgutov:
Thanks!

PR #1161, opened 2021-08-06
Add GitHub Actions checks

yugaego:
Hi Dmitry,
(Hope you're out from zeitnot and are doing fine.)
I've considered several possible setups - referenced below - and chose to keep it (stupid) simple:
`Makefile` already exists, is in use, and can potentially be used in future with fewer number of external dependencies.

Example with the [deliberate fails](https://github.com/yugaego/company-mode/actions/runs/1104317877).

Example with the [real checks](https://github.com/yugaego/company-mode/actions/runs/1104875213).

There's a particular issue: `company-gtags.el:79:47:Error: executable-find called with 2 arguments, but accepts only 1` on [Emacs < 27.1](https://github.com/emacs-mirror/emacs/commit/6f649e77b8512f73b17f03fd795beea9965c4029).
I didn't have a chance to put much attention to it yet: should a fallback be given for pre-27.1 Emacses? Or ignore compilation check for these versions?

If to start byte-compile `test/*.el` files, this error would appear:
`test/frontends-tests.el:153:60: Error: Unused lexical argument `candidate'`
(Same here, I didn't have a chance to look into fixing it yet).

- Would it be fine to add `test/*.el` files to `make clean` and `make compile-warn`?

### TODO
- Consider adding a linter (I've got zero knowledge at this stage about `elint`).
- Add/fix(travis) badges.

### Reference:
- [Nix usage without purcell/setup-emacs](https://github.com/magit/magit/blob/master/.github/workflows/test.yml)
- [Comparison](https://github.com/leotaku/elisp-check/blob/master/COMPARISON.md) of Emacs-related CI solutions.
- [Purcell's Linter](https://github.com/purcell/package-lint) itself and its workflow.

### Actions vs Travis

**GitHub Actions**:
- Something new to try (though might be not stable).
- Potentially could be used as one point for CI control.
- [Free usage limits](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions).
- ?Potential vendor lock-in (not a problem with a simple setup).

**Travis CI**:
- It's already setup and working.
- Not fully integrated into GitHub
  - Less visible checks results in comparison to the Actions.
- "has been trying to deprecate its open-source offering".
- ?Used to be not stable (doesn't mean it's unstable now).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,42 @@
+name: CI
+
+on:
+  push:
+    paths:
+      - '**.el'
+      - '**ci.yml'
+      - 'Makefile'
+    branches:
+      - master
+
+  pull_request:
+    paths:
+      - '**.el'
+      - '**ci.yml'
+      - 'Makefile'
+    branches:
+      - master
+
+jobs:
+  check:
+    runs-on: ubuntu-20.04
+
+    strategy:
+      fail-fast: false
+      matrix:
+        emacs_version: [25.1, 25.3, 26.3, 27.2, snapshot]
+
+    steps:
+      - name: Setup Emacs
+        uses: purcell/setup-emacs@v3.0
+        with:
+          version: ${{ matrix.emacs_version }}
+
+      - name: Checkout Company
+        uses: actions/checkout@v2
+
+      - name: Run tests
+        run: make test-batch
+
+      - name: Run compilation
+        run: make compile-warn
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ elpa: *.el
 	tar cvf company-$$version.tar --mode 644 "$$dir"
 
 clean:
-	@rm -rf company-*/ company-*.tar company-*.tar.bz2 *.elc ert.el
+	@rm -rf company-*/ company-*.tar company-*.tar.bz2 *.elc ert.el test/*.elc
 
 test:
 	${EMACS} -Q -nw -L . -l test/all.el \
@@ -33,3 +33,8 @@ test-batch:
 
 compile:
 	${EMACS} -Q --batch -L . -f batch-byte-compile company.el company-*.el
+
+compile-warn:
+	${EMACS} -Q --batch -L . \
+	--eval "(setq byte-compile-error-on-warn t)" \
+	-f batch-byte-compile company*.el test/*.el


dgutov:
Hi Yuzhana!

I'm definitely better, how are things on your side?

This is looking great, thanks a lot. There's little downside to migrating (except for the effort required, which you have already done): the vendor lock-in is minimal (`ci.yml` can be rewritten for any similar service), and we'll think about usage limit when/if we hit them.

I've never really looked much into linters for Emacs Lisp. I suppose they are most useful in projects with many active contributors. But experiments welcome.

> There's a particular issue: company-gtags.el:79:47:Error: executable-find called with 2 arguments, but accepts only 1 on Emacs < 27.1.
I didn't have a chance to put much attention to it yet: should a fallback be given for pre-27.1 Emacses? Or ignore compilation check for these versions?

I think wrapping it with `with-no-warnings` is the way to go: the feature is guarded by a version check, and there seems to be no easy alternative in older versions. I'll just push that change now.

> If to start byte-compile test/*.el files, this error would appear:
test/frontends-tests.el:153:60: Error: Unused lexical argument candidate'`
(Same here, I didn't have a chance to look into fixing it yet).

Should be fixed now too.

> Would it be fine to add test/*.el files to make clean and make compile-warn?

Let's enable it, sure. We can always turn it back off if it ends up annoying.

yugaego:
> how are things on your side?

So far, so good :knock on wood:-)

> Let's enable it, sure. We can always turn it back off if it ends up annoying.

Added together with the `branches: master` on-trigger filter. 

**company-gtags.el**: 
The warnings are still raised on older versions with the version check: `(version<= "27" emacs-version)`
I didn't find anything related at `debbugs.gnu.org`. Any ideas how to proceed?

**test/core-tests.el**:
`fboundp` check should be enough

yugaego:
> how are things on your side?

So far, so good :knock on wood:-)

> Let's enable it, sure. We can always turn it back off if it ends up annoying.

Added together with the `branches: master` on-trigger filter. 

**company-gtags.el**: 
```
   ((and (fboundp 'with-connection-local-variables)
         (fboundp 'connection-local-set-profile-variables)
         (file-remote-p default-directory))
```
`fboundp` in place of the version check [fixes compilation warnings](https://github.com/yugaego/company-mode/actions/runs/1112965452) on Emacs 26,
but not earlier.


**test/core-tests.el**:
The same, `fboundp` on Emacs 25 doesn't suppress the warning.

I'm looking into other approaches described in the [recent docs for suppressing](https://www.gnu.org/software/emacs/manual/html_node/elisp/Compiler-Errors.html) compiler errors,
but fixes/suggestions are welcome.

yugaego:
> company-gtags.el:

Ah, I missed/confused one more function name: hence using `fboundp` in `company-gtags` [fixes compilation warning](https://github.com/yugaego/company-mode/actions/runs/1113141934) on all the checked `Emacs` versions.

dgutov:
Thanks a lot! :+1:

yugaego:
Cool! 🥳 It looks nice.

PR #1164, opened 2021-08-09
Fix compilation warnings on Emacs 25-26

yugaego:
Fix for #1161 .

Uses `fboundp` as suggested in the [Compiler Error docs](https://www.gnu.org/software/emacs/manual/html_node/elisp/Compiler-Errors.html).

@dgutov , I didn't test `company-gtags--executable` in action. Please do so if it's simple for you. Thank you.

### DIFF
--- a/test/core-tests.el
+++ b/test/core-tests.el
@@ -586,6 +586,10 @@
       (setq header-line-format "aaaaaaa")
       (should (= (company--row) 0)))))
 
+;; Avoid compilation warnings on Emacs 25.
+(declare-function display-line-numbers-mode "ext:display-line-numbers")
+(declare-function line-number-display-width "indent.c")
+
 (ert-deftest company-column-with-line-numbers-display ()
   :tags '(interactive)
   (skip-unless (fboundp 'display-line-numbers-mode))
--- a/company-gtags.el
+++ b/company-gtags.el
@@ -67,8 +67,11 @@ completion."
   (cond
    ((not (eq company-gtags--executable 'unknown)) ;; the value is already cached
     company-gtags--executable)
-   ((and (version<= "27" emacs-version)           ;; can search remotely to set
-         (file-remote-p default-directory))
+   ((and  ;; Run remotely on supported versions of Emacs.
+     (fboundp 'with-connection-local-variables)
+     (fboundp 'connection-local-set-profile-variables)
+     (fboundp 'connection-local-set-profiles)
+     (file-remote-p default-directory))
 
     (with-connection-local-variables
      (if (boundp 'company-gtags--executable-connection)


yugaego:
Also added a probable fix for the tests warnings thus getting rid of [all of them](https://github.com/yugaego/company-mode/actions/runs/1113369744).

dgutov:
Thanks for the efforts, but this approach feels kinda brittle to me: you make a typo, and the test suddenly ends up doing nothing (becoming a false positive).

I suggest we use another approach, briefly mentioned in that document: [declare-function](https://www.gnu.org/software/emacs/manual/html_node/elisp/Declaring-Functions.html).

For example, core-tests.el seems to only need

```el
(declare-funciton display-line-numbers-mode "display-line-numbers")
(declare-function line-number-display-width "indent.c")
```

yugaego:
> this approach feels kinda brittle to me: you make a typo, and the test suddenly ends up doing nothing

I agree it's brittle, even despite the fact I didn't understand the typo example. Could you elaborate more?

Well, CI ran [successfully](https://github.com/yugaego/company-mode/actions/runs/1115368779) with `declare-function`.
To my great surprise, it worked even given an absent in Emacs 25 `display-line-numbers` file (as well as being passed `nil` as a FILE value).
However, in the latest change I added `ext:` prefix to comply with the Emacs 25 [`declare-function` docstring](https://github.com/emacs-mirror/emacs/blob/emacs-25/lisp/subr.el#L32).
Or am I misunderstanding it?

test/core-tests.el
```diff
@@ -586,6 +586,10 @@
       (setq header-line-format "aaaaaaa")
       (should (= (company--row) 0)))))
 
+;; Avoid compilation warnings on Emacs 25.
+(declare-function display-line-numbers-mode "ext:display-line-numbers")
```
yugaego, line 590: On the second thought, could it be safer for Emacs > 25 having this line wrapped into the Emacs-25 only version check? (I mean I'm not fully sure such function declaration is safe for the versions that have `display-line-numbers` file).

dgutov: They are safe enough: the declaration doesn't make the function defined, it just removes the byte-compilation warning.

dgutov: Such declarations are also used when you don't want to `require` a dependency outright, but load it on demand later when some function is called.

If a function belonging to said dependency is used that is not autoloaded, the file will get byte-compilation errors, which are handled with `declare-function`.

dgutov: BTW, I think people have been using the `ext:` prefix to mean third-party code (like externally distributed packages), and `display-line-numbers` is in the core.

Not that the file name matter much, though.

dgutov:
> I agree it's brittle, even despite the fact I didn't understand the typo example. Could you elaborate more?

Consider that if an expression looks like `(and (fboundp 'foo) (fboundp 'bar') (fboundp 'bar))`, you only need to make a typo in one of the names to make the branch unreachable (or someone else could do that later when modifying the code).

Of course, the odds to hit this problem are low, but those byte-compilation checks are mainly supposed to protect against typos (I think) anyway.

dgutov:
So I'm going to change the fix in the first file as well, but I'll go ahead and merge this first together with the other PR, to get the benefit of the new CI when doing so. ;-)

dgutov:
> However, in the latest change I added ext: prefix to comply with the Emacs 25 declare-function docstring.
Or am I misunderstanding it?

I think `ext:` only really makes sense in Emacs core.

It could help if we were running `check-declare-file` on the CI, and if we did that on Emacs versions that don't have the file in question. But we don't, and unlikely to do so. We could also wrap those declares in a version check anyway.

yugaego:
Thank you for the clarifications!

PR #1166, opened 2021-08-14
Add more badges

yugaego:
Hi Dmitry,
I had some fun today with [the badges](https://github.com/yugaego/company-mode/tree/badges).
Emacs' one is my favorite hand-crafted guy.
Let me know if you'd take *some* of them.
Best.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,4 +5,7 @@ File a bug report in [Issues](https://github.com/company-mode/company-mode/issue
 Ask a question or suggest a feature in [Discussions](https://github.com/company-mode/company-mode/discussions/).
 
 [![Build Status](https://github.com/company-mode/company-mode/actions/workflows/ci.yml/badge.svg)](https://github.com/company-mode/company-mode/actions/workflows/ci.yml)
+[![GNU Emacs](https://img.shields.io/static/v1?logo=gnuemacs&logoColor=fafafa&label=Made%20for&message=GNU%20Emacs&color=7F5AB6&style=flat)](https://www.gnu.org/software/emacs/)
 [![MELPA](https://melpa.org/packages/company-badge.svg)](https://melpa.org/#/company)
+[![GNU ELPA](https://elpa.gnu.org/packages/company.svg)](https://elpa.gnu.org/packages/company.html)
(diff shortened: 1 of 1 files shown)


dgutov:
Hi Yuzhana,

Nice!

Let's drop two of them: the license (it should be obvious: same as Emacs) and melpa stable (lately it's not recommended even by melpa maintainers).

And let's also move melpa in front of gnu elpa because it's very handy to ask users to try a new version by installing from it, so I'd rather people installed from there by default.

yugaego:
> Let's drop two of them the license (it should be obvious: same as Emacs)

Sure, though I might surprise you, that it's not **that** obvious (Emacs license including).
At least, it wasn't obvious to me and there's a chance I'm not alone in this regard.

I tried [another order](https://github.com/yugaego/tmp) of the badges, but that rather longish elpa-devel svg doesn't want to look nicely. It'd much better to have a twice as shorter elpa-devel svg.
(And I *personally* prefer to keep it since it was exciting to discover elpa-devel.)

dgutov:
> Sure, though I might surprise you, that it's not that obvious (Emacs license including).

I think the more canonical way to do this is to add a LICENSE file, then the license will be displayed in the right sidebar. See how it looks at https://github.com/mooz/js2-mode/.

> I tried another order of the badges, but that rather longish elpa-devel svg doesn't want to look nicely. It'd much better to have a twice as shorter elpa-devel svg.

The first one looks pretty nice. And the 2nd is the one I suggested, right? Looks a tad worse than the two others. The 3rd option also has a good idea, so... would you mind making the choice? My (mild) preference is on #1, though.

> (And I personally prefer to keep it since it was exciting to discover elpa-devel.)

Yeah, it's good to spread awareness. The badge is long-ish, but that's not a good enough reason to drop it.

yugaego:
> I think the more canonical way to do this is to add a LICENSE file

Yes, thought so too.

> My (mild) preference is on # 1, though.

Updated to No 1.
No 2 looked worse also because of a wider gap between the lines (had to add a newline there).
Considering colors, it would be ideal to put all the red badges (ELPA) on the second line, but that would result in a wider gap too.

> it's good to spread awareness

👍🏻

dgutov:
Thanks!

yugaego:
👍🏻

PR #1180, opened 2021-08-23
New option company-keywords-ignore-case

twlz0ne:


company-keywords.el
```diff
@@ -307,7 +315,7 @@
                  (not (company-in-string-or-comment))
                  (or (company-grab-symbol) 'stop)))
     (candidates
-     (let ((completion-ignore-case nil)
+     (let ((completion-ignore-case company-keywords-ignore-case)
```
dgutov, line 318: You also need to add an `ignore-case` clause to the `cl-case` here.

See its description in the `company-backends` docstring.

dgutov:
@twlz0ne See the comment here.

PR #1215, opened 2021-09-20
company-dabbrev/code: allow to use completion-styles for filtering

kiennq:
Allow to use `completion-styles` for filtering by switch to use `completion-all-completions` instead of `all-completions` for `company-dabbrev/code`.
This allows using `flex` completion style with `company-dabbrev/code`.
Also, extract the logic to calculating `:company-match` from string's faces so we can reuse it to highlight `company-dabbrev/code` items.

image

### DIFF
--- a/company-capf.el
+++ b/company-capf.el
@@ -111,27 +111,10 @@ so we can't just use the preceding variable instead.")
     (`match
      ;; Ask the for the `:company-match' function.  If that doesn't help,
      ;; fallback to sniffing for face changes to get a suitable value.
-     (let ((f (plist-get (nthcdr 4 company-capf--current-completion-data)
-                         :company-match)))
-       (if f (funcall f arg)
-         (let* ((match-start nil) (pos -1)
-                (prop-value nil)  (faces nil)
-                (has-face-p nil)  chunks
-                (limit (length arg)))
-           (while (< pos limit)
-             (setq pos
-                   (if (< pos 0) 0 (next-property-change pos arg limit)))
-             (setq prop-value (or
-                               (get-text-property pos 'face arg)
-                               (get-text-property pos 'font-lock-face arg))
-                   faces (if (listp prop-value) prop-value (list prop-value))
-                   has-face-p (memq 'completions-common-part faces))
-             (cond ((and (not match-start) has-face-p)
-                    (setq match-start pos))
-                   ((and match-start (not has-face-p))
-                    (push (cons match-start pos) chunks)
-                    (setq match-start nil))))
-           (nreverse chunks)))))
+     (let ((f (or (plist-get (nthcdr 4 company-capf--current-completion-data)
+                             :company-match)
+                  #'company--match-from-face)))
+       (funcall f arg)))
     (`duplicates t)
     (`no-cache t)   ;Not much can be done here, as long as we handle
                     ;non-prefix matches.
--- a/company-dabbrev-code.el
+++ b/company-dabbrev-code.el
@@ -81,25 +81,30 @@ also `company-dabbrev-code-time-limit'."
 The backend looks for all symbols in the current buffer that aren't in
 comments or strings."
   (interactive (list 'interactive))
-  (cl-case command
-    (interactive (company-begin-backend 'company-dabbrev-code))
-    (prefix (and (or (eq t company-dabbrev-code-modes)
-                     (apply #'derived-mode-p company-dabbrev-code-modes))
-                 (or company-dabbrev-code-everywhere
-                     (not (company-in-string-or-comment)))
-                 (or (company-grab-symbol) 'stop)))
-    (candidates (let ((case-fold-search company-dabbrev-code-ignore-case))
-                  (company-dabbrev--search
-                   (company-dabbrev-code--make-regexp arg)
-                   company-dabbrev-code-time-limit
-                   (pcase company-dabbrev-code-other-buffers
-                     (`t (list major-mode))
-                     (`code company-dabbrev-code-modes)
-                     (`all `all))
-                   (not company-dabbrev-code-everywhere))))
-    (kind 'text)
-    (ignore-case company-dabbrev-code-ignore-case)
-    (duplicates t)))
+  (pcase command
+    (`interactive (company-begin-backend 'company-dabbrev-code))
+    (`prefix (and (or (eq t company-dabbrev-code-modes)
+                      (apply #'derived-mode-p company-dabbrev-code-modes))
+                  (or company-dabbrev-code-everywhere
+                      (not (company-in-string-or-comment)))
+                  (or (company-grab-symbol) 'stop)))
+    (`candidates (let ((case-fold-search company-dabbrev-code-ignore-case))
+                   (company-dabbrev--filter
+                    arg
+                    (company-dabbrev--search
+                     (company-dabbrev-code--make-regexp
+                      (substring arg 0 company-minimum-prefix-length))
+                     company-dabbrev-code-time-limit
+                     (pcase company-dabbrev-code-other-buffers
+                       (`t (list major-mode))
+                       (`code company-dabbrev-code-modes)
+                       (`all `all))
+                     (not company-dabbrev-code-everywhere)))))
+    (`kind 'text)
+    (`ignore-case company-dabbrev-code-ignore-case)
+    (`match (company--match-from-face arg))
+    (`duplicates t)
+    (`no-cache t)))
 
 (provide 'company-dabbrev-code)
 ;;; company-dabbrev-code.el ends here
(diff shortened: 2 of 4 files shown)


dgutov:
Hi!

Sorry for the wait. This looks very interesting, but really not how I imagined this feature (dabbrev with fuzzy matching) would materialize.

Does it work well for you? Trying it out, I've noticed two things:

* The popup blinks when you type. This often indicates a caching problem. And indeed, when I added `('no-cache t)` to dabbrev-code's definition, the blinking went away.
* What is the expected/recommended `completion-styles` setup for this?

In a ruby-mode buffer, for example, I couldn't do fuzzy matching on class names because `completion-pcm-word-delimiters` don't support CamelCase delimiters.

And if I do `(setq company-styles '(flex))`, suddenly Elisp completion starts behaving differently (which is expected, but still a little uncomfortable) and slightly slower.

Perhaps per-mode customization of `completion-styles` is the way to go. Or getting used to Elisp completion working differently.

kiennq:
> Does it work well for you?

Yes, it works reasonably well for me. I have been working on a very large project that renders LSP useless for quite a time and I have to fall back to dabbrev. With this patch, I can do fuzzy matching now, which is much better than before.



> * The popup blinks when you type. This often indicates a caching problem. And indeed, when I added `('no-cache t)` to dabbrev-code's definition, the blinking went away.

Oh, I didn't notice the blink, partly I've got used to it and don't feel annoyed anymore. 😄 
But yes, I agree with adding `('no-cache t)` there.



> * What is the expected/recommended `completion-styles` setup for this?

I've been using
``` elisp
(setq completion-styles `(basic partial-completion initials flex)))
```
With that, if the prefix matching is failing, the fuzzy match will be kicked in and used to find candidates.
Using `orderless` instead of `flex` is also okay as I heard so far (haven't tried it).


The `flex` completion style is very slow for me so I have a little patch for it so it becomes more performant by do prefix match first to scope down the number of candidates that need to be filtered against the fuzzy-matching pattern. 
It works very well for me now and makes using `flex` much more toleratable.
``` elisp
  ;; patches for flex completion style
  (defvar x/flex-minimum-prefix-length 2)
  (advice-add #'completion-flex--make-flex-pattern :override
              (lambda (pattern)
                "Turn (prefix \"foo\" point) to (\"fo\" any \"o\" point)."
                (when (and x/flex-minimum-prefix-length (equal (car pattern) 'prefix))
                  (setq pattern (cdr pattern)))
                (let ((prefix-len 0))
                  (mapcan (lambda (elem)
                            (prog1
                                (if (stringp elem)
                                    (mapcan
                                     (lambda (char)
                                       (when prefix-len (cl-incf prefix-len))
                                       (if (and prefix-len
                                                (<= prefix-len x/flex-minimum-prefix-length))
                                           (list (string char))
                                         (list 'any (string char) 'any)))
                                     elem)
                                  (list elem))
                              (setq prefix-len nil)))
                          pattern)))
              '((name . --prefix-match)))
```

kiennq:
@dgutov Do you have more suggestions on this?

dgutov:
Yes, sorry for the wait.

> Using orderless instead of flex is also okay as I heard so far (haven't tried it).

For posterity: apparently that one requires `orderless-component-separator` to be customized (Company backends usually don't allow spaces in the prefix).

> (setq completion-styles `(basic partial-completion initials flex)))

Hmm, that's not too bad.

But it's not like we can change the default value of `completion-styles`, right? And `company-dabbrev-code` should still work, and no worse than before.

> The flex completion style is very slow for me so I have a little patch for it so it becomes more performant by do prefix match first to scope down the number of candidates that need to be filtered against the fuzzy-matching pattern.

Makes sense to me. Maybe we could propose this change as a user option, or something.

In the meantime... how about we add a new option like `company-dabbrev-completion-styles`? Default it to `nil` and have the current behavior in that case. But when it is set to a list of styles (in the same format as `completion-styles`), use it.  A value of `t` could also be used.

The disabling of built-in cache (`no-cache t`) would also be dependent on this variable being non-nil.

Here's a couple of other problems:

* `company-minimum-prefix-length` should not be used here because `company-manual-begin` binds it to 0. As a result, we have different behavior between idle completion and manual completion. I've spent more time than I'd care to admit investigating this. If you need to use a number, use `company-dabbrev-minimum-length` -- but in a different way, see below.

* Using `(substring arg 0 company-minimum-prefix-length)` as prefix still limits the possible completions, such as non-prefix `flex` (which is its default behavior, before your customization) or the `initials` style which stubbornly refuses to match obvious inputs. That's a non-obvious behavior, and if we want to support completion styles, let's do that properly. When completion styles are enabled, `company-dabbrev-code--make-regexp` should construct something like `"\\_<\\([a-zA-Z]\\|\\s_\\)\\(\\sw\\|\\s_\\)\\{3,\\}\\_>"` as the search prefix.

Admittedly, this *will* slow down the search for completions. My print-debugging says the search with an empty prefix only manages to reach like 5-6 buffers before reaching the 0.1s time limit. Maybe this is fine.

Or maybe it could use a smarter cache, since the built-in prefix-based cache is not available.

I've always figured that to have smooth fuzzy dabbrev completion we'd need to have per-buffer caches of words, invalidated only on buffer's modification. It might even be possible to make that invalidation very fast, if we store both the concatenated list of words, as well as references to the conses on both sides of the slice of the list that came from a particular buffer (in buffer-local vars).

I wouldn't mind merging this change without the last part (smart cache), but it might turn out to be both easy enough to implement and necessary for a smooth experience here.

dgutov:
We could also try using regexp like `"\\(\\sw\\|\\s_\\)*a\\(\\sw\\|\\s_\\)*b\\(\\sw\\|\\s_\\)*c\\(\\sw\\|\\s_\\)*"` (for input `abc`), to limit the number of matches, making the cache implementation less urgent. This should work for most completion styles, too. But not Orderless.

And it's bound to slow down the matching somewhat, too.

dgutov:
Speaking of the cache, caching the global (already concatenated) list is probably not a good idea, because different buffers will want to get works from different subsets of other buffers. And `(apply #'append ...)` is pretty fast anyway.

dgutov:
@kiennq Do you want to try the below patch on top of the branch in PR #1411?

It's not much different from your code, but with the help of the new caching thingy it's now more or less guaranteed to be relatively fast. So I'm curious how useful this shortcut (*) can be in "a very large project".

If your value of `company-minimum-prefix-length` was 1 already (is it?), then the behavior could be exactly the same.

(*) prefix-filtering by the first prefix char only.

```diff
diff --git a/company-dabbrev-code.el b/company-dabbrev-code.el
index 49293a6..cdd4b47 100644
--- a/company-dabbrev-code.el
+++ b/company-dabbrev-code.el
@@ -88,18 +88,31 @@ comments or strings."
                  (or company-dabbrev-code-everywhere
                      (not (company-in-string-or-comment)))
                  (or (company-grab-symbol) 'stop)))
-    (candidates (let ((case-fold-search company-dabbrev-code-ignore-case))
-                  (all-completions
-                   ""
-                   (company-dabbrev--search
-                    (company-dabbrev-code--make-regexp arg)
-                    company-dabbrev-code-time-limit
-                    (pcase company-dabbrev-code-other-buffers
-                      (`t (list major-mode))
-                      (`code company-dabbrev-code-modes)
-                      (`all `all))
-                    (not company-dabbrev-code-everywhere)))))
+    (candidates
+     (let* ((case-fold-search company-dabbrev-code-ignore-case)
+            (prefix (if (string-empty-p arg) "" (substring arg 0 1)))
+            (res
+             (completion-all-completions
+              arg
+              (company-cache-fetch
+               'dabbrev-code-candidates
+               (lambda ()
+                 (company-dabbrev--search
+                  (company-dabbrev-code--make-regexp prefix)
+                  company-dabbrev-code-time-limit
+                  (pcase company-dabbrev-code-other-buffers
+                    (`t (list major-mode))
+                    (`code company-dabbrev-code-modes)
+                    (`all `all))
+                  (not company-dabbrev-code-everywhere)))
+               :expire t
+               :check-tag prefix)
+              nil (length arg))))
+       (if (numberp (cdr (last res)))
+           (setcdr (last res) nil))
+       res))
     (kind 'text)
+    (no-cache t)
     (ignore-case company-dabbrev-code-ignore-case)
     (duplicates t)))
 

```

dgutov:
I've pushed a version of this to master, test at your convenience.

It adds a new option `company-dabbrev-code-completion-styles`, which naturally only affects one backend. Not sure if non-prefix matching would be similarly useful for dabbrev/ispell, but adding options for those as well wouldn't be hard.

The set of matches returned by `company-dabbrev-code` is still limited by the performance of the scan, which is repeated for every completion session.

PR #1228, opened 2021-09-26
Add User Manual: Texinfo for Info and HTML generation

### LINKED ISSUE #926: Add info manual
It would be very useful to have an Info manual for this package. I realize it could be a big task, so please consider this issue request to be only for a manual which includes the bare minimum: what you need to get up and running. It could be extended later.

Thanks in advance.

yugaego:
This is an early stage draft with the main goal to close #926 .

@dgutov , both `ELPA` and `MELPA` support `info` files auto-generation.

[Seemingly, `ELPA` would require adding the following options to `elpa-packages.el` for `company`:]

```lisp
  :renames (("doc/" ""))
  :doc "doc/company.texi")
```

So there's no *need* in adding `info` generation to `Company` `Makefile` for these package archives.

But, probably, it may be handy to have a target for manual `info` generation.
It could be provided as `make info` target, f.i., or/and added to `make package` and/or `make elpa` targets.
[BTW, is there any reason to keep supporting both `make package` and `make elpa`?]

Also, optionally, other packages often publish the same manual(s) for on-line access.
Do you have an opinion if we should consider this option from the start?


dgutov:
Hi @yugaego!

Looks like a good start.

> So there's no need in adding info generation to Company Makefile for these package archives.

Excellent. IIRC GNU ELPA added this feature relatively recently (I didn't really follow this topic but remember seeing related mailing list threads).

> But, probably, it may be handy to have a target for manual info generation.

I have no opinion on that. :-)

> [BTW, is there any reason to keep supporting both make package and make elpa?]

Probably not. I think at least one of them was useful when Marmalade was still a thing.

> Also, optionally, other packages often publish the same manual(s) for on-line access.

I'm all for it! I guess Texinfo has some option for HTML output? But it might be not trivial to make it look good (maybe there are some existing CSS styles that could be reused).

IIUC some of the popular "other packages" don't actually use texinfo and write docs in other formats right away, such as Cider (https://github.com/clojure-emacs/cider/tree/master/doc) or (https://github.com/emacs-lsp/lsp-mode/tree/master/docs). That allows easy publishing on some online documentation platforms. I'm not sure how easy it is to generate .info files from either of the formats, though.

yugaego:
> Excellent. IIRC GNU ELPA added this feature relatively recently

Yeah, it was added during Stefan M.'s great job with `GNU ELPA` throughout 2020-21.

> > [BTW, is there any reason to keep supporting both make package and make elpa?]

> Probably not. I think at least one of them was useful when Marmalade was still a thing.

Okay, then I'll probably suggest some related updates.
Docs related changes could be applied on top of that.

> Looks like a good start.

I'm glad you say so. (And sorry I may be slow to respond/act from time to time. Well, we all know that, but still.)

> I guess Texinfo has some option for HTML output? But it might be not trivial to make it look good

Yes.
To generate HTML it should be basically: `makeinfo --html company.info`.
The thing to decide would be where to serve from that/those `html` file(s).
Examples approaches for HTML/CSS generation/output:
- Swiper
  - https://github.com/abo-abo/swiper#manual 
  - https://github.com/abo-abo/swiper/blob/master/doc/Makefile
- Transient
  - https://magit.vc/manual/transient/
  - https://github.com/magit/transient/blob/master/docs/Makefile
- AUCTeX
  - https://www.gnu.org/software/auctex/manual/auctex.index.html
  - http://git.savannah.gnu.org/cgit/auctex.git/tree/Makefile.in

> some of the popular "other packages" don't actually use texinfo and write docs in other formats right away

I had a plan to fallback to `org` for that, if needed.
[ `ELPA` supports generation of `org` format docs and there're `M-x org-texinfo-export-to-texinfo` and `M-x org-html-export-to-html`.]
Though if `texinfo` works sufficiently good, it'd seem to be a fine choice currently for an Emacs package.

Anyway, I'm going to postpone on my side with any additional considerations until the actual texts have been written and tried to be wrapped up with `texinfo` format.

dgutov:
> The thing to decide would be where to serve from that/those html file(s).

From the website, I guess? Put them in a subdir, and there we go. You know the repository already. ;-)

> Examples approaches for HTML/CSS generation/output:

All three have different looks, though. I wonder where the styles come from.

> Anyway, I'm going to postpone on my side with any additional considerations until the actual texts have been written and tried to be wrapped up with texinfo format.

Sounds good. :+1:

yugaego:
All the commands to generate the files are in `/doc/Makefile`

My main concerns are written in the comments.

One more is: I'd like to output images in the Info manual, and they're loaded fine on my system with either: `C-u C-h i </doc/company.info>` or pre-installed in Emacs with `make install-info`.
But I don't know:
- if it would work with older Emacs.
- if it's a good idea at all to have images in the system's `info` directory.
- if/how package archives would handle it. 

So this part would need more investigation.

yugaego:
Hi Dmitry,
I noticed that I didn't specifically mention these additions are ready for review.

Please take a look whenever you can (no hurry). And let me know if you have any questions regarding setting this all up for a review.

I'm postponing with the previously mentioned images setup question or any other potential changes until getting overall feedback from you.

Thank you.

dgutov:
Hi Yuzhana,

Sorry I've been putting it off, swamped by other things.

Regarding the images, I think we could merge the current version and then see how it works in action when the info file is generated by different pipelines (e.g. MELPA and GNU ELPA). TBH I have zero experience with this, but we can easily afford to make some mistakes and then fix them.

doc/company.texi
```diff
@@ -0,0 +1,1499 @@
...
+@c company-select-first
+@c company-select-last
+@c company-next-page
+@c company-previous-page
+@c company-select-next-if-tooltip-visible-or-complete-selection
+@c company-version
+@c But that should not confuse a newbie with too much information.
+@c Or maybe describe how to list all the available commands
+@c in a simple way (would it be M-x occur <interactive> ?).
+@c Might be provide a separate reference with Company commands
```
dgutov, line 403: If these don't seem essential to you, maybe indeed avoid documenting them specifically.

And while `company-select-next` is indeed important, you probably indirectly described it already if you mentioned when `C-n` does when the completion is in progress.

doc/company.texi
```diff
@@ -0,0 +1,1499 @@
...
+description characters (@pxref{Syntax Class Table,,,elisp}), or a
+predicate function.  By default, @code{company-auto-commit-chars} is
+set to the list of the syntax characters: @w{@code{(?\ ?\) ?.)}},
+which translates to the whitespaces, close parenthesis, and
+punctuation.  The particular convenience of
+@code{company-auto-commit-chars} is that they do not act as a trigger
+when they are a part of valid completion.
+@end defopt
+
+@c TODO: Maybe also describe the following options.
```
dgutov, line 545: Up to you. But if you are going to do that, I'd say the first 3 are probably more important than the last two.

As in, more likely to be customized.

doc/company.texi
```diff
@@ -0,0 +1,1499 @@
...
+(setq company-tooltip-offset-display 'lines)
+@end lisp
+
+@img{tooltip-offset-display}
+@end defopt
+
+@c TODO: what is the correct description / behavior?
+@c Maybe improve the docstring.
+@c Does it influence minimal height of the tooltip
+@c together with company-tooltip-limit?
```
dgutov, line 703: Should be, yes.

"limit" is the max height.

"mimium" is the min height.

doc/company.texi
```diff
@@ -0,0 +1,1499 @@
...
+@c Can we say it ``Controls the position of a tooltip above
+@c or below the point''?
+@c Does this option control two different things: height and position?
+@c @defopt company-tooltip-minimum
+@c Sets the minimum height of a tooltip (also influenced by
+@c @code{company-tooltip-limit}).  If this quantity of lines is not
```
dgutov, line 709: > If this quantity of lines is not available in the window, then the tooltip is displayed above the point.

Something like that, yes.

(quantity -> number)

doc/company.texi
```diff
@@ -0,0 +1,1499 @@
...
+@c available in the window, then the tooltip is displayed above the
+@c point.
+
+@c @c This doesn't work the way I expect it to.
```
dgutov, line 713: What's the problem?

doc/company.texi
```diff
@@ -0,0 +1,1499 @@
...
+backend(s), as shown in the following concise example:
+
+@lisp
+(setq company-backends '((company-capf :with company-yasnippet)))
+@end lisp
+
+@node Package Backends
+@section Package Backends
+
+@c TODO: Maybe extend some of the backends description
```
dgutov, line 1233: Maybe.

dgutov:
In addition, I would probably avoid documenting the "toy" frontends like `company-echo-frontend` because I don't think anybody really is using them. If the manual is an intro to tweaks and customizations we actually recommend.

(`company-echo-metadata-frontend` is fine, on the other hand)

doc/company.texi
yugaego: It seems I didn't "re-init" company tooltip (remove letters fully, type again) after scrolling the page and hence was getting wrong positions of the tooltip (above/below the point).

doc/company.texi
yugaego: Okay, then I'm removing these comments for now. These descriptions can be added later, if there's such a need.

> while company-select-next is indeed important, you probably indirectly described it already if you mentioned when C-n

Yeah, almost. `C-n` is documented with `company-select-next-or-abort`. Maybe `company-select-next` can be mentioned there too, though not fully sure. I think this also can be resolved on a later stage.

doc/company.texi
yugaego: > I'd say the first 3 are probably more important than the last two.

Thanks, good to know.

I'm removing this comment too (I don't like todos or fixmes hanging around in the source).

doc/company.texi
yugaego: Would the following description be precise then?

It gets away from using the word "height", that makes think of ~"actual height" of a tooltip
rather than an ~"abstract height" of the requested space below the point.

"When the number of lines between the point and the bottom of the
window is less than @code{company-tooltip-minimum} value, then the
tooltip is displayed above the point."

yugaego:
Hi Dmitry,

> Sorry I've been putting it off, swamped by other things.

Yeah, I almost got nervous this time 😄 

> Regarding the images, I think we could merge the current version and then see how it works in action 
> ...  we can easily afford to make some mistakes and then fix them.

Fine by me.

> I would probably avoid documenting the "toy" frontends like company-echo-frontend 

I'll think about it. I personally found them to be "fun", "nice-to-discover" backends; and they don't seem to take too much space in the text; and yes, there's still `company-echo-metadata-frontend` to mention. Maybe add a comment they're considered to be "toy" frontends to lower the expectations bar for this pair?

doc/company.texi
dgutov: Sounds good.

doc/company.texi
dgutov: Yep. Or at least it should work like that.

dgutov:
> Yeah, I almost got nervous this time smile

:wink: 

> Maybe add a comment they're considered to be "toy" frontends to lower the expectations bar for this pair?

Yes, that would be fine. Something to effect that these two aren't used usually, but provide a fun/interesting alternative to the main UI to play with.

doc/company.texi
yugaego: Yes, I want to try extending/changing it a bit. At least to extend the tiniest "company-files" section, so that it wouldn't look that miserable.
The initial idea was to devote a separate subsection to a backend, but that didn't work nicely. The current version of this section was the third - and the best IMO so far - attempt to make it fit in more or less.

dgutov:
I think I should just merge this now, so it's easier to iterate and see how the build on ELPA works.

Just one thing: the manual currently says "This user manual is for Company version 0.9.14", whereas the current release is 0.9.13. I think it should either document the latest stable release, or explicitly call out that it's documenting the development version.

Or we can have two different versions of the manual at the same time.

dgutov:
Anyway, that's shouldn't be a problem either, if we only generate and publish the manual when doing a new release. But it would be nice to fix this particular version, if we're going public with the manual before the next release.

Looking forward to next revisions. :+1:

yugaego:
> the manual currently says "This user manual is for Company version 0.9.14", whereas the current release is 0.9.13.

Yes, this was one of the concerns I postponed to decide on.

[It made sense to me to write for the development version since the gap is too
big between the stable 0.9.13 and the development snapshot; and many of the new
users seem to start with the development version of Company.]

> or explicitly call out that it's documenting the development version.

Yup, this is what it should be in this case.

Do you have a preference for the version name?

I'd prefer a pre-release type of naming in this case (`0.9.14-dev`,
`0.9.14-alpha`, `0.9.14.rc`, etc; something in the spirit of
[semver](https://semver.org/#spec-item-3)).

But sticking closer to the GNU devel type of versioning might make sense too,
for the sake of consistency. Then it would look more like
`0.9.13.0.last_change_date`. (BTW, I'm experiencing troubles loading data from
the gnu.org servers today.)

> Or we can have two different versions of the manual at the same time.

In the case of Info manual, nothing would have to be done, I think. A user would
get the manual with the downloaded Company snapshot/release.

For the online version, it may be convenient to provide at least the last stable
release manual too (starting after the following release). F.i., after the
release date, the manual can be moved into the folder `/manual-0.9.14/`, and the
development version may still be kept at `/manual/`. Or `/manual/0.9.14/` and
`/manual/latest/`, adapting a quite wide-spread practice. Or even
`/docs/user-manual/latest/`, in case another documentation adds up in the
future. If so, we might be better to rename the folder now. What do you think?

Now, after writing this, I think that maybe the sentence under the question
may simply be changed to:
"This user manual is for Company development snapshot."
(or some variation of this).

doc/company.texi
yugaego: Further improvements are suggested by the https://github.com/company-mode/company-mode/pull/1279.

dgutov:
> I'd prefer a pre-release type of naming in this case (0.9.14-dev,
0.9.14-alpha, 0.9.14.rc, etc; something in the spirit of
semver).

These might look unnecessarily complex.

> Do you have a preference for the version name?

If the plan is to only publish the development version of the manual now, and then only update it with each release, "This user manual is for Company development snapshot." should indeed suffice. Then we change that to the new version during each release and voila.

But I suppose the current (development) version of the manual would show up in the shapshot versions of the packages later? E.g. in MELPA or in ELPA-devel. Then we'd need to be more careful with versioning indeed.

And indeed change the current header version to something like `0.9.14snapshot`. (`version-to-list`'s docstring mentions other acceptable suffixes.)

yugaego:
Then maybe  "This user manual is for Company version 0.9.14snapshot (23 December 2021)."?
It'd actually be closer to one of the recommendations given in `Texinfo` manual, that even suggests to use [`editions`](https://www.gnu.org/software/texinfo/manual/texinfo/texinfo.html#FOOT1) (footnote 1).
Or see [Magit approach](https://github.com/magit/magit/commit/74b84b57cc6aca261392e9189e8e6c5e9cc19ca7).

dgutov:
I think I'm fine with either.

Which of the options requires less frequent fiddling inside `.el` files? :slightly_smiling_face:

yugaego:
Dmitry, I've got an additional concern regarding *Info* images: even the small images would add around 650KB to the size of the package. I've come up with two alternate solutions:
1. Do not use images in the *Info* version at all.
2. Output URLs instead. For example (the caption can be any text):

```
 ...
     beside the candidates.  Setting the option value to ‘t’ aligns
     annotations to the right side of the tooltip.

         (setq company-tooltip-align-annotations t)

    Online demo:
    (https://company-mode.github.io/manual/images/small/tooltip-annotations.png)
```

Do you have any preference/opinion?

dgutov:
> even the small images would add around 650KB to the size of the package

I think it might be fine. Or people will tell us, and then we'll try to solve that problem.

But speaking of images and sizes:

The HTML seems to have images with text size (perhaps scaling?) varying wildly. The Frontends section has images where the size of text on images is almost never constant between images. Some very noticeable examples here:

https://company-mode.github.io/manual/Frontends.html

Is that due to how Texinfo HTML export works? It assigns the same image width everywhere, no matter the original sizes?

dgutov:
Speaking of "Output URLs instead" (a solution to the previous question): a number of die-hards will dislike this approach anyway because it would require an Internet connection to view the manual.

yugaego:
> >  images would add around 650KB to the size 

> I think it might be fine. 

> > Output URLs instead

>  a number of die-hards will dislike this approach anyway because it would require an Internet connection to view the manual.

Okay, thanks, it's good to know. I'll proceed with initial approach then.

> The Frontends section has images where the size of text on images is almost never constant between images.

The differences are the middle ground between the quality of the shown picture; weight and size(dimensions) of the images; and the space each example requires in Emacs to be clear enough but small.  As usual, it should be possible to improve, especially given the existing examples, but on my side all the pictorial resources are exhausted (at least currently, I wouldn't dare going through this work again). If you see an easy fix to improve even slightly, I'm all ears.

>  It assigns the same image width everywhere, no matter the original sizes?

The images displayed size is now restricted by the CSS width parameter.

yugaego:
>  If you see an easy fix to improve even slightly, I'm all ears.

Or maybe a better methodology. I'll think about it again and will try to come up with a better approach.

yugaego:
I tested locally the following snippet produces the expected result with MELPA:
- doc/*.texi is recognized by default (so company.info is already being generated).
- instruct to create "<root>/images/small" from the source's "doc/images/small/*.png".

```
(company :repo "company-mode/company-mode"
         :fetcher github
         :files (:defaults "icons" ("images/small" "doc/images/small/*.png")))
```

But before this PR, please help me to clarify and commit the ELPA configuration (to be sure no changes to the current directory structure would be required).

I keep getting "Invalid version syntax: ‘0.9.13.0...." error trying to build the package locally. Maybe you have a proper local ELPA setup to check this out?

I assume the following snippet should work:
- `:doc` says from which file to generate `<root>/company.info`
- `:renames` supposedly should move `doc/images/small` to the `<root>/images/small` and then get rid of the rest of the `doc/` folder. (If the latter wouldn't work, `:ignored-files` or `.elpaignore` would do it).

```
 ("company"
  :url "https://github.com/company-mode/company-mode.git"
  :doc "doc/company.texi"
  :renames (("doc/images/small/" "images/small") ("doc/" ""))
  :auto-sync t)
```

Thank you!

dgutov:
It's been a while since I last tried building GNU ELPA checkout's packages locally.

Which command are you trying that leads to this error?

yugaego:
> Which command are you trying that leads to this error?

I've been experimenting with the commands listed under `* Make targets` in `elpa/admin/README`.
I believe that was `make build/company` that resulted in that error.

> It's been a while since I last tried building GNU ELPA checkout's packages locally.

Okay, I'll then give it a try again soon.

dgutov:
OK, I checked out the latest `main`, did `make` to have the `admin` worktree set up, then I had to `apt install markdown`, or else the `markdown` command was complained at.

After that, `make build/company EMACSBIN=~/vc/emacs-master/src/emacs` finishes successfully:

```sh
$ make build/company EMACSBIN=~/vc/emacs/src/emacs
~/vc/emacs/src/emacs --no-site-lisp --batch -l /home/dgutov/vc/elpa/admin/elpa-admin.el	\
         -f elpaa-batch-make-one-package company
======== Building tarball archive-devel/company-0.9.13.0.20220102.1036.tar...
######## Built new package archive-devel/company-0.9.13.0.20220102.1036.tar!
======== Building tarball archive/company-0.9.13.tar...
######## Built new package archive/company-0.9.13.tar!
```

Perhaps you have some older generated file you could remove, which the Makefile rules don't recreate as needed.

Or your local `emacs` is too old? I'm not sure which versions should be considered too old, but in some recent-ish past there were some improvements to the version-parsing code. See the above example for how to override which Emacs executable is used.

yugaego:
Thanks!

> Or your local emacs is too old?

Not really: 28.0.90. And the build was fresh. I saw there were updates upstream to more reliably fetch a package version, and it looks like the version may be the cause of the issue. Here's the full error message from today's attempt (after `make clean` and `make clean/company`):

> Build error for archive-devel/company-0.9.13.0......tar: (error "Invalid version syntax: ‘0.9.13.0.....’")

So yes, I'll try to compile the latest emacs, maybe this will help fix the issue.

But anyway, since you have a working setup, could you [try the recipe](https://github.com/company-mode/company-mode/pull/1228#issuecomment-1002982997)?
The expected result is to have `company.info` and `images/small/*.png`, but not the `doc/` folder in the tarball.

yugaego:
I've made *some* progress after making all the steps anew:

- git clone https://git.savannah.gnu.org/git/emacs/elpa.git
- cd elpa
- make setup
- make packages/company
  - on the second attempt also 'cd packages/company; git pull'
- make build/company
  - same result with the EMACSBIN=/latest/compiled/src/emacs

Output:
```
emacs --no-site-lisp --batch -Q -l admin/elpa-admin.el \
         -f elpaa-batch-pkg-spec-make-dependencies .pkg-descs.mk
emacs --no-site-lisp --batch -l /Users/yuga/Programs/Sources/elpa/admin/elpa-admin.el	\
         -f elpaa-batch-make-one-package company
======== Building tarball archive-devel/company-0.9.13.0.20220103.35144.tar...
######## Built new package archive-devel/company-0.9.13.0.20220103.35144.tar!
======== Building tarball archive/company-0.9.13.tar...
######## Built new package archive/company-0.9.13.tar!
```

But I still don't see the symlinked tar files (am I misunderstanding something here?), so I'm back to figuring out how to handle it.

`ls archive-devel`

```
273B Jan  3 12:16 archive-contents
1.3K Jan  3 12:16 company-readme.txt
3.9K Jan  3 12:16 company.html
1.3K Jan  3 12:16 company.svg
35B Jan  3 12:16 company.tar -> company-0.9.13.0.20220103.35144.tar
128B Jan  3 12:13 doc
4.7K Jan  3 12:11 elpaa--export-inputGg8wLS.html
4.7K Jan  3 12:16 elpaa--export-inputeLoayw.html
2.0K Jan  3 12:16 index.html
```

dgutov:
> I've made some progress after making all the steps anew:

I guess it was indeed because of some existing generated files. :-(

> 35B Jan  3 12:16 company.tar -> company-0.9.13.0.20220103.35144.tar

That's a symlink, isn't it?

> But anyway, since you have a working setup, could you try the recipe?
The expected result is to have company.info and images/small/*.png, but not the doc/ folder in the tarball.

It's odd. I don't really understand what's going on.

First, I had to fix it a little to say `("doc/images/small/" "images/small/")` (note the slash after the second "small").

`make build/company` finishes with messages like:

```
======== Building tarball archive-devel/company-0.9.13.0.20220102.1036.tar...
doc/company.texi:651: warning: could not find @image file `./images/small/tooltip-annotations.txt' nor alternate text (possibly involving @img)
doc/company.texi:665: warning: could not find @image file `./images/small/tooltip-limit.txt' nor alternate text (possibly involving @img)
...

doc/company.texi:651: warning: @image file `./images/large/tooltip-annotations' (for HTML) not found, using `./images/large/tooltip-annotations..png' (possibly involving @img)
doc/company.texi:665: warning: @image file `./images/large/tooltip-limit' (for HTML) not found, using `./images/large/tooltip-limit..png' (possibly involving @img)
...
```

and then the tar file contains both `images/large` and `images/small`, but not `doc/`. `company.texi` is there (correctly), however.

I haven't tried reading the resulting manual yet.

yugaego:
Thanks for checking this out!

> That's a symlink, isn't it?

Yes, linking to a non-existing `.tar` file (both `archive/` and `archive-devel/`). Maybe I should report it as an `GNU ELPA` bug? I plan on debugging `elpaa--make-one-tarball-1` today. Hopefully there's something trivial, just a bit obscure.

yugaego:
> Hopefully there's something trivial, just a bit obscure.

Ha-ha. Trivial enough. I learned there's `debug` setting in `elpa-config`. With it enabled and all the previous investigation, the issue became obvious: `BSD`/`GNU Linux` commands options incompatibility again.

```
call-process tar ("--exclude-vcs" "-X" "/../elpa/packages/company/.elpaignore"
 "--transform"  "s|^packages/company|company-0.9.13.0.20220103.35144|" "-chf"
 "archive-devel/company-0.9.13.0.20220103.35144.tar" "packages/company")
```

`Mac`'s `bsdtar` doesn't have `--transform` option. So the issue was quick-fixed by installing/using `gnu-tar`. Hence, now I can proceed with the recipe tests (I think).
(And am going to report it as a bug.)

Thanks again for your help with it!

dgutov:
Ah yes, the macOS's outdated command line tools are a pain.

Not sure using certain options is a bug per se (we expect GNU Make, so depending on GNU Tar would make sense too), but the lack of sane error reporting (if it indeed failed silently) seems like a problem.

yugaego:
> but the lack of sane error reporting (if it indeed failed silently) seems like a problem.

I filed the bug report with a tiny patch to ensure failing on a non-existing tarball at [debbugs#53048](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=53048) .

yugaego:
@dgutov FYI, appeared there's another issue related to GNU ELPA setup (auto-generated HTML documentation images and styles). I created [a bug report with a patch](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=53205) for it. The plan is to move forward with the rest after this patch is considered.

dgutov:
Cool. Let's see how it goes.

I was not aware GNU ELPA publishes any HTML-generated documentation. Is there a package which has such docs there already?

yugaego:
> I was not aware GNU ELPA publishes any HTML-generated documentation.

Yes, I noticed it being generated doing the tests locally.

> Is there a package which has such docs there already?

Hm, good question. Looks like maybe they have this feature disabled currently on the GNU ELPA server.(?)
I checked several of the packages pages with the documentation but don't see that 'Manual' link to the generated docs appearing online.

yugaego:
Hi Dmitry 👋🏻 

Since there was no progress on the [ELPA images issue](https://github.com/company-mode/company-mode/pull/1228#issuecomment-1011125105) for more than 2 weeks,
I'd like to proceed with the archive managers recipes updates for the manual.

Could you commit the following change to the ELPA or should I send it to the emacs-devel as a patch?

Commit message: `* elpa-packages (company): Add info manual`

``` diff
diff --git a/elpa-packages b/elpa-packages
index 82a2ec9071..7d006e9af5 100644
--- a/elpa-packages
+++ b/elpa-packages
@@ -108,6 +108,9 @@
   :auto-sync t)
  ("company"
   :url "https://github.com/company-mode/company-mode.git"
+  :shell-command "cp -R doc/images ."
+  :doc "doc/company.texi"
+  :ignored-files ("doc")
   :auto-sync t)
  ("company-ebdb"	:url nil)
  ("company-math"	:url "https://github.com/vspinu/company-math.git")
```

`:shell-command "cp -R doc/images ."` fixes [not found images warnings](https://github.com/company-mode/company-mode/pull/1228#issuecomment-1005326006).

[I assume that the stable package won't be re-generated unless a new release is published;
otherwise, errors would be thrown because of an absent 'doc/' directory on the stable package build.]

And I'm going to create a MELPA PR with [the following commit](https://github.com/yugaego/melpa/commit/fa4c54f01f8dabcce04d8e316c478a7e0d01049b).

Is it all fine with you?

dgutov:
Hi YE!

> Could you commit the following change to the ELPA or should I send it to the emacs-devel as a patch?

Sure, done.

> [I assume that the stable package won't be re-generated unless a new release is published;
otherwise, errors would be thrown because of an absent 'doc/' directory on the stable package build.]

Probably not. Otherwise, we'll probably hear about it. :-)

> And I'm going to create a MELPA PR with [the following commit](https://github.com/yugaego/melpa/commit/fa4c54f01f8dabcce04d8e316c478a7e0d01049b).

Looks good, thanks.

dgutov:
@yugaego I've just updated the manual for the 1.0.0 release, including the website.

Thanks again for the setup that made syncing very easy!

yugaego:
Hi @dgutov , I am delighted the setup works fine.

Fun fact: several months after releasing this manual, I was invited to work on a proprietary documentation. This tells a story about how a FOSS hobby can sometimes turn into a paid activity.

PR #1230, opened 2021-09-30
Makefile updates and fixes

yugaego:
Hi Dmitry, I kept all the commits for now so to briefly explain each change.
Let me know if it'd be sufficient or you'd prefer to have the commits squashed.


dgutov:
Thanks!

No need to squash, I rather like separate meaningful commits.

PR #1240, opened 2021-10-20
Add support for deprecated property

yyoncho:


dgutov:
@yyoncho Looking at the gif you just posted: how about we apply the strike-through effect (and the deprecation face, I guess) only to the textual part of the completion item? Meaning, not the margins or the whitespace.

Like on VS Code's screenshot here:

https://external-content.duckduckgo.com/iu/?u=https%3A%2F%2Fcode.visualstudio.com%2Fassets%2Fupdates%2F1_38%2Fcss-deprecated-properties.png&f=1&nofb=1

PR #1261, opened 2021-11-18
Add missing dot in `company-vscode-icons-mapping`

vatai:
This missing dot caused `expand-file-name` to receive a list instead of a string, which in turn caused it to fail.

### DIFF
--- a/company.el
+++ b/company.el
@@ -1541,7 +1541,7 @@ end of the match."
     (string . "symbol-string.svg")
     (struct . "symbol-structure.svg")
     (text . "symbol-key.svg")
-    (type-parameter "symbol-parameter.svg")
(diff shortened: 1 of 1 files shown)


dgutov:
Thanks for noticing!

I was wondering just recently which LS used this kind. The Python one does?

vatai:
Dear @dgutov, 

Yeah. It crashed for example when triggering the auto completion after say `pathlib.` in python (with `pathlib` imported ofc). 

Happy to have helped :) Happy to use your software! :) Thx!

dgutov:
`pathlib` members include type parameters, huh? I guess that has something to do with typed Python, hmm.

vatai:
Dunno... `pathlib` was just an example :shrug: I think there were other modules I tried which didn't work...

PR #1270, opened 2021-11-29
Rename 'auto-commit' user options

yugaego:
Based on #1264.

I'm still considering if this would be a good idea to apply this change.
It may be convenient considering it in parallel with the suggested manual text. 
So let's keep it open for now, without merging.

My first serious attempt was to use `company-triggered-complete`/`company-triggered-complete-chars` but that proved to be a weak choice on a closer look (as with many other names).

The current choice seems to be self-explanatory (as of today) but its long-ish.

### DIFF
--- a/company.el
+++ b/company.el
@@ -566,7 +566,8 @@ doesn't match anything or finish it manually, e.g. with RET."
 This can be a function do determine if a match is required.
 
 This can be overridden by the backend, if it returns t or `never' to
-`require-match'.  `company-auto-commit' also takes precedence over this."
+`require-match'.  `company-insertion-on-trigger' also takes precedence over
+this."
   :type '(choice (const :tag "Off" nil)
                  (function :tag "Predicate function")
                  (const :tag "On, if user interaction took place"
@@ -578,11 +579,18 @@ This can be overridden by the backend, if it returns t or `never' to
   'company-auto-commit
   "0.9.14")
 
-(defcustom company-auto-commit nil
-  "Determines whether to auto-commit.
-If this is enabled, all characters from `company-auto-commit-chars'
-trigger insertion of the selected completion candidate.
-This can also be a function."
+(define-obsolete-variable-alias
+  'company-auto-commit
+  'company-insertion-on-trigger
+  "0.9.14")
+
+(defcustom company-insertion-on-trigger nil
+  "If enabled, allow triggering insertion of the selected candidate.
+This can also be a predicate function, for example,
+`company-explicit-action-p'.
+
+See `company-insertion-triggers' for more details on how to define
+triggers."
   :type '(choice (const :tag "Off" nil)
                  (function :tag "Predicate function")
                  (const :tag "On, if user interaction took place"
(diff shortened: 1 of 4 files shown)


dgutov:
Not completely convinced by the name yet, to be honest.

Maybe `company-complete-on-trigger`? And `complete-complete-trigger-chars`? That's also long-ish.

Though I think the current name does well at resolving the ambiguity where "complete" at the same time mean initiating completion (showing the popup), completing a common part (but keeping the popup), and final insertion (when the popup goes away).

If we try to retain that specificity, maybe `company-insert-on-trigger` or `company-trigger-insertion`? And `company-insertion-trigger-chars`?

yugaego:
This naming is a tricky one. And finding an alternate approach didn't work out
too so far.

I've written down some of the (shorter) names to compare them side by side, maybe this
would help to move forward or decide on dropping this idea.

1.
company-insertion-on-trigger
company-insertion-triggers

( - The names are almost the same, so they are easy to confuse.)


2.
company-complete-on-trigger
company-complete-triggers

( - Are we going to insert (complete) some triggers?. Probably not.)


3.
company-trigger-complete / company-insert-on-trigger / company-triggered-...
company-triggers

( - Is the second name informative/clear enough? It may lack "namespace" context.)


4.
company-trigger-insertion
company-insertion-triggers

( - Broken "namespacing" because of the words order change.)


5.
company-triggered-insert
company-trigger-chars

( - Why not simply "company-triggers"? "Chars" seems to be a redundant word. Not
perfect "namespacing".)


Please share if this list causes any fresh thoughts.

dgutov:
`company-triggers` can also be easily interpreted (I think) as a synonym for `company-begin-commands`. Because those trigger completion (idle completion).

I think `company-insertion-triggers` and either `company-trigger-insertion` or `company-insertion-on-trigger` seem to be the most meaningful options. And I'm not getting the "easy to confuse" vibe.

If you like them well enough yourself, we can go ahead with those.

I don't have a strong opinion on whether to proceed with renames or drop this because OT1H the current names seemed fine when I introduced them (and I even felt clever about the naming choice ;-)), but OT2H the ones above might sound a little clearer, and I would probably pick them back then, if such option was suggested.

yugaego:
> I think company-insertion-triggers and either company-trigger-insertion or company-insertion-on-trigger seem to be the most meaningful options. 

I agree. Let's try again and see if they are still a good choice when in the context.

company.el
```diff
@@ -1954,18 +1970,20 @@ prefix match (same case) will be prioritized."
...
-(defun company-auto-commit-p (input)
-  "Return non-nil if INPUT should trigger auto-commit."
-  (and (if (functionp company-auto-commit)
-           (funcall company-auto-commit)
-         company-auto-commit)
-       (if (functionp company-auto-commit-chars)
-           (funcall company-auto-commit-chars input)
-         (if (consp company-auto-commit-chars)
+(defun company-insertion-on-trigger-p (input)
+  "Return non-nil if INPUT should trigger completion.
```
dgutov, line 1974: "trigger insertion", I guess.

company.el
```diff
@@ -578,11 +579,18 @@ This can be overridden by the backend, if it returns t or `never' to
...
-If this is enabled, all characters from `company-auto-commit-chars'
-trigger insertion of the selected completion candidate.
-This can also be a function."
+(define-obsolete-variable-alias
+  'company-auto-commit
+  'company-insertion-on-trigger
+  "0.9.14")
+
+(defcustom company-insertion-on-trigger nil
+  "If enabled, allow triggered completion with the selected candidate.
```
dgutov, line 588: "allow triggering insertion of the selected candidate"?

company.el
```diff
@@ -595,16 +603,24 @@ This can also be a function."
...
 
-This can also be a function, which is called with the new input and should
-return non-nil if company should auto-commit.
+If this is a string, then each character in it can trigger insertion of the
+selected candidate.  If it is a list of syntax description characters (see
+`modify-syntax-entry'), then characters with any of those syntaxes can act
+as triggers.
+
+Note that a character that is part of a valid completion never triggers
+completion.
```
dgutov, line 620: "insertion" here too, I guess?

company.el
```diff
@@ -595,16 +603,24 @@ This can also be a function."
...
 
-A character that is part of a valid completion never triggers auto-commit."
+This can also be a function, which is called with the new input.  To
+trigger completion, the function should return a non-nil value."
```
dgutov, line 623: and here

doc/company.texi
```diff
@@ -519,16 +510,16 @@ See @uref{https://www.gnu.org/software/emacs/manual/html_node/eintr/Wrong-Type-o
...
 
-@anchor{company-auto-commit-chars}
-@defopt company-auto-commit-chars
-This option acts only when @code{company-auto-commit} is enabled.  The
-value can be one of: a string of characters, a list of syntax
-description characters (@pxref{Syntax Class Table,,,elisp}), or a
-predicate function.  By default, @code{company-auto-commit-chars} is
-set to the list of the syntax characters: @w{@code{(?\ ?\) ?.)}},
+@defopt company-insertion-triggers
+This option acts only when @code{company-insertion-on-trigger} is
```
dgutov, line 514: "This option has an effect only ..."

dgutov (commented):
I've added a few comments with the aim of reducing the completion<->insertion ambiguity.

Up to you.

yugaego:
> I've added a few comments with the aim of reducing the completion<->insertion ambiguity.

Thanks, the suggested sounds better. Updated.

dgutov:
All right, this is looking good. Thanks!

With apologies to those who have migrated to the `-auto-commit` naming, and now have to migrate again, I'm merging this.

PR #1271, opened 2021-11-29
Rename tooltip scrollbar faces

yugaego:
This PR suggests using more descriptive faces names, better fitting the rest of the `tooltip` related faces.

Also, the `-fg`/`-bg` part may make it look like `:foreground`/`:background` face attribute is (only) set there.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,9 @@
 
 ## Next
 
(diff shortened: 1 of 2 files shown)


dgutov:
Better names. Thanks!

PR #1272, opened 2021-11-29
Fix docstrings

yugaego:
These minor fixes are intended to make docstrings declared behavior to match the reality: dot-icons are also affected by those user options.
The rest are minor stylistic updates.

### DIFF
--- a/company.el
+++ b/company.el
@@ -1659,14 +1659,15 @@ The only mandatory element in CONF is ICON, you can omit both the FG and BG
 fields without issue.
 
 When BG is omitted and `company-text-icons-add-background' is non-nil, a BG
-color will be generated using a gradient between the active tooltip color and
+color is generated using a gradient between the active tooltip color and
 the FG color."
   :type 'list)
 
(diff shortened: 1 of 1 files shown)


company.el
```diff
@@ -1659,22 +1659,23 @@ The only mandatory element in CONF is ICON, you can omit both the FG and BG
...
+Affects `company-text-icons-margin' and `company-dot-icons-margin'."
   :type 'list)
 
 (defcustom company-text-icons-format " %s "
   "Format string for printing the text icons."
   :type 'string)
 
 (defcustom company-text-icons-add-background nil
-  "When non-nil, generate a background color for text icons when none is given.
+  "Generate a background color for text/dot icons when none is given.
```
dgutov, line 1678: "When non-nil" had a standard purpose here, but I guess it's easy to infer without mentioning the condition explicitly.

yugaego: Sure. I removed it only because the line exceeded 75 chars limit.

PR #1273, opened 2021-11-29
Add complete-common-or-show-delayed-tooltip

yugaego:
Based on #1214.

All in all, I found approach [2. (complete common or show tooltip)](https://github.com/company-mode/company-mode/discussions/1214#discussioncomment-1424185)  to be the most convenient during everyday use. So if that name is not too long and ugly (or other considerations), then let's add it, shall we?


dgutov:
Not sure how I missed this notification.

Sorry, and thanks!

yugaego:
> Not sure how I missed this notification.

Yeah, no problem, it happens from time to time. Can't wait to utilize upstream defun 🎉

PR #1279, opened 2021-12-21
Improve company-tooltip-minimum docs

yugaego:
The suggested texts should better reflect what `company-tooltip-minimum` option influences.


company.el
```diff
@@ -266,8 +266,9 @@ The visualized data is stored in `company-prefix', `company-candidates',
   :type 'integer)
 
 (defcustom company-tooltip-minimum 6
-  "The minimum height of the tooltip.
-If this many lines are not available, prefer to display the tooltip above."
+  "Ensure visibility of this number of candidates.
+When that many lines are not available between the point and bottom of the
```
dgutov, line 270: In Emacs lingo, we usually use the term "point" without the article.

So: "between point and ..." and "above point".

Some examples here: https://www.gnu.org/software/emacs/manual/html_node/emacs/Point.html

yugaego: Removed. Please see now.

dgutov: Could you also fix that in the manual?

yugaego: Argh! Yes, sure. 👍🏻

yugaego: And now done. Let's try again 😉

dgutov:
Thanks!

PR #1280, opened 2021-12-21
Fix company-files-exclusions docstring and application

yugaego:
Replaces destructive `cl-delete-if-not` by non-destructive `cl-remove-if-not` to still exclude files when a directory is also configured for exclusion. The added test demonstrates the fixed case. (I checked it on Emacs 28.0.90).

Also improves the docstring and `:type` of this option.


dgutov:
Thanks!

PR #1284, opened 2021-12-25
User manual updates: Extend 'Package Backends' and fix typos

yugaego:
Hi Dmitry ,
I didn't figure out why docstring of `company-dabbrev-downcase` says 

> If you set this value to nil, you may also want to set ‘company-dabbrev-ignore-case’ to any value other than ‘keep-prefix’.

Could you elaborate on this?
Your input is welcome on other aspects of the suggested changes too.

This is the last major update I had on my mind at this stage.
The next proper step I think would be to check/polish package archives releases, which I didn't have a chance to look into yet and may need your help with. 

Thanks. And have a nice day!

### DIFF
--- a/doc/company.texi
+++ b/doc/company.texi
@@ -3,7 +3,7 @@
 @setfilename company.info
 @settitle Company User Manual
 @set VERSION 0.9.14snapshot
-@set UPDATED 27 December 2021
+@set UPDATED 28 December 2021
 @documentencoding UTF-8
 @documentlanguage en
 @paragraphindent asis
@@ -1251,11 +1251,8 @@ Optionally, search for the matches with @w{@kbd{M-x isearch @key{RET}
 
 @end itemize
 
-@node Symbol Names Completion
-@subsection Symbol Names Completion
-
-A completion for symbol names typically assists programmers working
-with the source code.
+@node Code Completion
+@subsection Code Completion
 
 @defun company-capf
 In the Emacs's world, the current tendency is to have the completion
@@ -1417,8 +1414,8 @@ dictionary.
 
 @end defun
 
-@node File Paths Completion
-@subsection File Paths Completion
+@node File Name Completion
+@subsection File Name Completion
 
 @defun company-files
 This backend can be used to retrieve completion candidates for the
@@ -1452,8 +1449,8 @@ file/directory known to be located under the inserted directory.
 
 @end defun
 
-@node Templates Expansion
-@subsection Templates Expansion
+@node Template Expansion
+@subsection Template Expansion
 
 @cindex template
 @cindex expansion
--- a/doc/company.texi
+++ b/doc/company.texi
@@ -3,7 +3,7 @@
 @setfilename company.info
 @settitle Company User Manual
 @set VERSION 0.9.14snapshot
-@set UPDATED 25 December 2021
+@set UPDATED 27 December 2021
 @documentencoding UTF-8
 @documentlanguage en
 @paragraphindent asis
@@ -923,13 +923,13 @@ The look of the preview is controlled by the following faces:
 @code{company-preview}, @code{company-preview-common}, and
 @code{company-preview-search}.
 
-@itemize @w{}
-@item
 @img{preview-light}
 
-@item
+@ifnothtml
+@sp 1
+@end ifnothtml
+
 @img{preview-dark}
-@end itemize
 
 @node Echo Frontends
 @section Echo Frontends
@@ -1066,16 +1066,19 @@ If @code{company-show-quick-access} is enabled, @emph{tooltip-} and
 (setq company-show-quick-access 'left)
 @end lisp
 
-@itemize @w{}
-@item
 @img{tooltip-quick-access}
 
-@item
+@ifnothtml
+@sp 1
+@end ifnothtml
+
 @img{echo-qa}
 
-@item
+@ifnothtml
+@sp 1
+@end ifnothtml
+
 @img{echo-strip-qa}
-@end itemize
 
 @cindex custom
 @cindex configure
@@ -1233,11 +1236,19 @@ following:
 
 @itemize
 @item
-Execute command @code{M-x customize-group @key{RET} <backend-name>}.
+Execute command @w{@kbd{M-x customize-group @key{RET}
+<backend-name>}}.
 
 @item
-Open the source file of the backend and run @w{@code{M-x occur
+Open the source file of the backend and run @w{@kbd{M-x occur
 @key{RET} ^(defcustom}}.
+
+@itemize @minus
+@item
+Optionally, search for the matches with @w{@kbd{M-x isearch @key{RET}
+(defcustom}}.
+@end itemize
+
 @end itemize
 
 @node Symbol Names Completion
@@ -1248,7 +1259,7 @@ with the source code.
 
 @defun company-capf
 In the Emacs's world, the current tendency is to have the completion
-logic executed by @w{@code{completion-at-point-functions}}
+logic provided by @w{@code{completion-at-point-functions}}
 (@acronym{CAPF}) implementations.  [Among the other things, this is
 what the popular packages that support language server protocol
 (@acronym{LSP}) also rely on.]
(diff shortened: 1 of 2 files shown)


dgutov:
Hi Yuzhana!

The default config is optimized for writing prose (which we're assuming company-dabbrev is mostly used for). Same goes for `company-ispell`. Where you might start your current word with a capital letter, but most likely the rest of them will be lower case. So both "keep prefix" and "downcase" kind of work.

If you try one of the described non-recommended combinations, you can see something like below. Which seems kind of Frankenstein-ian to me.

Screenshot from 2021-12-27 01-34-38

doc/company.texi
```diff
@@ -1223,17 +1226,18 @@ The following sections give a short overview of the commonly used
...
 following:
 
 @itemize
 @item
 Execute command @code{M-x customize-group @key{RET} <backend-name>}.
 
 @item
-Open the source file of the backend and run @code{M-x occur @key{RET}
-^(defcustom}.
+Open the source file of the backend and run @w{@code{M-x occur
```
dgutov, line 1239: I would probably suggest `isearch` instead of `M-x occur`.

The latter takes more effort, including jumping across windows.

doc/company.texi
```diff
@@ -1243,11 +1247,11 @@ A completion for symbol names typically assists programmers working
 with the source code.
 
 @defun company-capf
-In Emacs' world, the current tendency is to have the completion logic
-executed by @w{@code{completion-at-point-functions}} (@acronym{CAPF})
-implementations.  [Among the other things, this is what the popular
-packages that support language server protocol (@acronym{LSP}) also
-rely on.]
+In the Emacs's world, the current tendency is to have the completion
+logic executed by @w{@code{completion-at-point-functions}}
```
dgutov, line 1251: "completion logic *provided* by ...", maybe?

Or "supplied by", or "implemented by".

doc/company.texi
```diff
@@ -1266,9 +1270,9 @@ For more details on @acronym{CAPF}, @ref{Completion in Buffers,,,elisp}.
 @end defun
 
 @defun company-dabbrev-code
-This backend works similarly to the Emacs built-in package
-@emph{dabbrev}, parsing completion candidates from the text in open
-buffer(s).  Internally, its based on the backend
+This backend works similarly to the built-in Emacs package
+@emph{dabbrev}, searching for completion candidates in the text
+available in the open buffer(s).  Internally, its based on the backend
```
dgutov, line 1275: Suggestion:

> searching for completion candidates inside the contents of all open buffers. Or a subset of them.

doc/company.texi
```diff
@@ -1305,22 +1309,136 @@ switching to @emph{company-capf} backend.
 @subsection Natural Language Completion
```
dgutov, line 1309: Perhaps we should call this "Text Completion" or "Prose Completion"?

I'm not 100% sure, but I think the phrase "natural language" comes up more often in specialized contexts, where one also deals with the (vague, complex, but still existing) rules of natural languages and the meanings behind words. And we just provide dictionary completion, of sorts.

Which (Dictionary Completion) could be another option for the name of this subsection.

doc/company.texi
```diff
@@ -1338,13 +1456,14 @@ their expansions.
...
-Emacs package for creating and inserting (expanding) templates.
+@uref{https://www.lysator.liu.se/~davidk/elisp/, Tempo}, one more
+built-in Emacs package for creating and inserting (expanding)
+templates.
 @end defun
 
 @defun company-yasnippet
-Used as a completion backend for a third-party template system
-@uref{https://github.com/joaotavora/yasnippet, YASnippet}.
+Used as a completion backend for a quite popular third-party template
```
dgutov, line 1465: We could even call it "the most popular", I think.

dgutov:
See the inline comments. Everything looks good otherwise.

Let me know if you prefer I just merge this as-is, with comments to be addressed later.

Cheers!

yugaego:
> If you try one of the described non-recommended combinations, you can see something like below. Which seems kind of Frankenstein-ian to me.

Ah, okay, I see now. Thank you for the explanation.

> Let me know if you prefer I just merge this as-is, with comments to be addressed later.

Thank you for the comments, I'd like to address them right away.

doc/company.texi
yugaego: Yep, "provided by" should be fine.

doc/company.texi
yugaego: I settled with "the popular".

doc/company.texi
yugaego: "Text Completion" seems to be a good enough title, and a simple one. Updated to use it.
Note that the titles are used in the generated URLs, so it might help avoid issues in the future to choose the "durable".  So let me know if you see a vague title.

doc/company.texi
yugaego: Updated to "inside the contents of the open buffer(s)", because I have a mild preference to give as an overall definition here as possible.

doc/company.texi
yugaego: Updated to also mention `isearch`: it's a good one to remind about, but the main idea here is to propose something to "fetch a list of options". I personally find it more comprehensible to see the options of a new for me package listed row by row someplace, rather than only scattered in the source. In other words, seems I'm a strong `occur` proponent 😄 .

dgutov: Sounds good. :-)

doc/company.texi
dgutov: Regarding other titles:

"Templates Expansion" could be renamed to "Snippet Expansion". These are more or less equivalent, but in my mind (at least) a template usually has some place-holders for values. And `company-abbrev` (or `abbrev` package in general) does not do that.

"File Paths Completion" -- according to GNU standards, a "path" is a list of values (like the value of an env var `PATH` on unixy systems). Whereas an absolute or relative file name (with directory parts, etc) is still a file name. So the section should probably be called "File Name Completion".

yugaego: Thank you for looking into this.

> "Templates Expansion" could be renamed to "Snippet Expansion".  These are more or less equivalent

I agree with you, these words seem to be more or less interchangeable and "snippet" might be a bit more suitable choice overall. 

After looking into updating the text, it appeared there's a reason to stick with the "template": both `Tempo` and `YASsnippet` define themselves with this term. Respectively: "Flexible template insertion" and ["is a template system"](https://github.com/joaotavora/yasnippet). Changing the current text feels clumsy because of this, so let's keep the current version.

[BTW, (probably you're aware of it) `abbrev` allows to configure a hook function for rudimentary additional actions (f.i. move point to a place-holder in the expansion).]

> File Name Completion

It's an interesting catch. From a targeted reader perspective, it may confuse (since the term "path" is common in this context), but the section's text should clarify what exactly is to be completed.  

Would it be fine to use the plural form "File Names Completion" then? Otherwise, maybe turn siblings into the singular form: "Symbol Name Completion", "Template Expansion"?

I was also going to suggest POSIX's "Pathname" term, but then stumbled upon this text in `(emacs) Glossary`, `File Name` section:

> Some people use the term `pathname` for file names, but we do not;
we use the word `path` only in the term `search path` (q.v.).

dgutov: > After looking into updating the text, it appeared there's a reason to stick with the "template": both Tempo and YASsnippet define themselves with this term. Respectively: "Flexible template insertion" and "is a template system". Changing the current text feels clumsy because of this, so let's keep the current version.

But it is ya*snippet*, right? And tempo is probably old enough that its description was written back when the term "snippet" didn't enter the common language.

And if we look at VS Code's docs (which most likely reflect the popular usage), the main name is "snippet", which is then clarified to be a [kind of] template: https://code.visualstudio.com/docs/editor/userdefinedsnippets

Anyway, you to you, it's not too serious an issue either way.

> [BTW, (probably you're aware of it) abbrev allows to configure a hook function for rudimentary additional actions (f.i. move point to a place-holder in the expansion).]

Not at all. I've never used this backend either. Good to know, though. :-)

> Would it be fine to use the plural form "File Names Completion" then? Otherwise, maybe turn siblings into the singular form: "Symbol Name Completion", "Template Expansion"?

I think the singular form is best. File Name Completion, Symbol Name Completion, etc.

We could also replace "Symbol Name Completion" with "Code Completion". Which is a more general term, and it's what `company-capf` does, really. Especially when used together with "smart" completion providers, such as, for example, the LSP clients.

> I was also going to suggest POSIX's "Pathname" term, but then stumbled upon this text in (emacs) Glossary, File Name section:

Yup.

yugaego: > the main name is "snippet", which is then clarified to be a [kind of] template

Funny enough, this could be looked at as one more voice for the "template" term: since it defines a superset of "snippet", and a more general title is preferred.

> it's not too serious an issue either way.

+1

> I think the singular form is best.  ... We could also replace "Symbol Name Completion" with "Code Completion". 

Thanks! The names looks much cleaner now.

dgutov: > since it defines a superset of "snippet", and a more general title is preferred

But we don't provide expansion of any other kind of templates?

yugaego: > But we don't provide expansion of any other kind of templates?

Yes, sure. I meant to state "just in case, for future possible (even if unlikely) changes".

PR #1288, opened 2021-12-30
Bind ‘non-essential’ when performing idle completion.

phst:
Idle completion isn’t a direct response to a user action, so I believe marking
it as non-essential is justified.

### DIFF
--- a/company.el
+++ b/company.el
@@ -1898,9 +1898,10 @@ prefix match (same case) will be prioritized."
        (eq win (selected-window))
        (eq tick (buffer-chars-modified-tick))
        (eq pos (point))
-       (when (company-auto-begin)
-         (let ((this-command 'company-idle-begin))
-           (company-post-command)))))
+       (let ((non-essential t))
+         (when (company-auto-begin)
+           (let ((this-command 'company-idle-begin))
+             (company-post-command))))))
 
 (defun company-auto-begin ()
   (and company-mode


dgutov:
Hi!

That is currently done in `company--fetch-candidates.

Doesn't that work to your satisfaction?

phst:
I don't think that binds `non-essential` around the backends, at least not for `completion-at-point-functions`. I tried this using

```
(add-hook 'completion-at-point-functions
          (lambda ()
            (message "completion-at-point-functions: non-essential = %s"
                     non-essential)
            nil))
```

and after attempting idle completion (just hitting space in some buffer with `company-mode` enabled) I got the message

> completion-at-point-functions: non-essential = nil

I also don't see anything in the code of company.el that would call `company--fetch-candidates` before invoking the `completion-at-point` backends.

dgutov:
That's true: `non-essential` comes into play only when the completion table is polled.

Do you need that value already when the c-a-p-f is called? Could you explain what kind of logic would use it?

phst:
Yeah, I think you're right, I added a completion function to c-a-p-f that accesses the filesystem before constructing the completion table, which the manual explains is something to not do. So feel free to close this PR; OTOH, binding `non-essential` here also shouldn't hurt.

dgutov:
We can keep this open, together with this discussion.

There might be a legitimate use of it there: like returning a different function for TABLE depending on `non-essential`. But I'd like to know if somebody's doing it.

phst:
I'm at least planning to, see https://github.com/bazelbuild/emacs-bazel-mode/pull/324 and https://github.com/bazelbuild/emacs-bazel-mode/pull/325. I haven't checked whether that actually improves things significantly in the case of bazel.el; maybe https://github.com/bazelbuild/emacs-bazel-mode/pull/329 is the better/more correct fix.

dgutov:
OK, thanks.

Without getting into detailed comparisons between the fixes (which you should be a better judge of), let's merge this change, since it indeed shouldn't hurt and probably is TRT.

PR #1289, opened 2021-12-31
User manual: Minor fixes

ssnnoo:
Minor fix in manual: company-search-flex-regexp was missing

### DIFF
--- a/doc/company.texi
+++ b/doc/company.texi
@@ -1011,7 +1011,7 @@ Searches for words separated with spaces in the given order.
 Searches for words separated with spaces in any order.
 @end defun
 
-@defun company-search-words-in-any-order-regexp
+@defun company-search-flex-regexp
 Searches for characters in the given order, with anything in between.
 @end defun
 


dgutov:
Was the intention to replace one with the other? Instead of mentioning both?

ssnnoo:
sorry was not clear: company-search-words-in-any-order-regexp is already mentioned above, hence there was a duplication instead of mentioning company-search-flex-regexp

before it looked like 

`@defun company-search-words-in-any-order-regexp
Searches for words separated with spaces in any order.
@end defun

@defun company-search-words-in-any-order-regexp
Searches for characters in the given order, with anything in between.
@end defun`

and now:

`@defun company-search-words-in-any-order-regexp
Searches for words separated with spaces in any order.
@end defun

@defun company-search-flex-regexp
Searches for characters in the given order, with anything in between.
@end defun`

dgutov:
Oh! Thank you, I didn't notice that.

PR #1291, opened 2022-01-06
company-complete: Insert current selection when no more common part

wullic:
When bind `company-complete`  to the completion key ( usually `<tab>`), If candidates have no more common part and the first candidate is just in need, press `<tab>` once to insert current selection, rather than  press `<tab>` twice that really make annoyed. Or in this situation I should use `<enter>` to insert it ,  in other case use `<tab> `?

### DIFF
--- a/company.el
+++ b/company.el
@@ -2645,7 +2645,9 @@ inserted."
   (when (company-manual-begin)
     (if (or company-selection-changed
             (and (eq real-last-command 'company-complete)
-                 (eq last-command 'company-complete-common)))
+                 (eq last-command 'company-complete-common))
+            (not company-common)
+            (equal company-prefix company-common))
         (call-interactively 'company-complete-selection)
       (call-interactively 'company-complete-common)
       (when company-candidates


dgutov:
Hi!

This will make it behave differently than currently designed. Like, as a user, it's hard to be sure whether there are further common chars, or the current input is it. I'm not sure I'd want to see completion inserted by pressing TAB too many times.

OTOH, this is a valid approach too. You can contribute it as a new command with a different name, or add it to the wiki (https://github.com/company-mode/company-mode/wiki/Tips-&-tricks). Maybe both: add a command and mention it in the wiki as a possible configuration.

dgutov:
> Yeah, could press TAB to see whether exist potential common parts that haven't been found, and then press TAB or ENTER to insert it, which works exactly nice when using ENTER as main insert key.

One doesn't necessarily keep in mind how many times they pressed TAB already, or whether they did so. At least, I'm not sure it's ideal to present a modal UI by default (one where "modes" a not always visible).

> BTW, there seems to be almost no more common chars for large number of candidates.

That really depends on the backend. But with non-prefix (fuzzy) matching, that is often the case, yes.

PR #1293, opened 2022-01-08
Improve sorting of case-insensitive completions

### LINKED ISSUE #1290: Unexpected completion on case-insensitive backends
#### Output of the command `M-x company-diag`
```
Emacs 27.2 (x86_64-w64-mingw32) of 2021-03-26 on CIRROCUMULUS
Company 0.9.13

company-backends: (company-bbdb company-semantic company-cmake company-capf company-clang company-files
              (company-dabbrev-code company-gtags company-etags company-keywords)
              company-dabbrev)

Used backend: (company-dabbrev-code company-gtags company-etags company-keywords)

Major mode: opascal-mode
Prefix: #("begin" 0 5
  (fontified t ws-butler-chg chg face font-lock-keyword-face))

Completions:
  "begin"
  "begin"
  "begin"
  "begin"
<tons of "begin" entries skipped>
  "begin"
  "begin"
  "begin"
  "BeginWrite"
  "begin"
  "begin"
  "begin"
  "begin"
  "BeginRead"
  "begin"
  "BeginRead"
  "begin"
  "begin"
  "BeginRead"
  "begin"
  "begin"
  "BeginRead"
  "begin"
  "BeginRead"
  "begin"
  "begin"
  "begin"
  "begin"
  "begin"
  "BeginWrite"
  "begin"
  "begin"
  "BeginWrite"
  "BeginRead"
  "begin"
  "begin"
  "begin"
  "begin"
  "BeginWrite"
  "begin"
  "begin"
  "begin"
<tons of "begin" entries skipped>
```

#### Describe the issue
Setting the `company-abort-on-unique-match` variable to `t` doesn't seem to work as expected: the list of completions is displayed as seen on the screenshot below. The typed text is present in the completion list, but it's not selected by default, so when user presses Enter, the typed text is replaced with whatever happens to be the first on the list. This is especially annoying if the Enter was pressed before the completion list was loaded/displayed, so the user has no chance to anticipate the trouble.

#### Steps to reproduce
Set the `company-abort-on-unique-match` variable to `t` using Customize.
Type the keyword `begin` in any Pascal/Delphi project that has functions like BeginRead or BeginWrite defined, and press Enter, because you always want that keyword to be the last word on the line.
Watch as company-mode replaces it with BeginRead or some other nonsense starting with "begin".

#### Expected behavior
The completion should be aborted, because `begin` is a full match of an item in the completion list. The typed word should not be replaced with anything else when user presses Enter or Space, etc.

#### Screenshots
2022-01-05 05-30-31 emacs
2022-01-05 05-35-17 emacs

alex-ilin:
Fixes #1290.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 ## Next
 
+* Improved sorting of case-insensitive completions
+  ([#1290](https://github.com/company-mode/company-mode/issues/1290)).
 * New command `company-complete-common-or-show-delayed-tooltip`
   ([#1214](https://github.com/company-mode/company-mode/discussions/1214)).
 * Faces `company-scrollbar-fg` and `company-scrollbar-bg` have been renamed to
--- a/company.el
+++ b/company.el
@@ -1468,12 +1468,22 @@ update if FORCE-UPDATE."
   (and (bound-and-true-p flyspell-mode)
        (version< emacs-version "27")))
 
+(defun string-less-ignore-case-p (string1 string2)
+  "Return non-nil if string1 < string2 case-insensitively."
+  (let ((res (compare-strings string1 0 nil string2 0 nil t)))
+    (and (numberp res) (< res 0))))
+
 (defun company--preprocess-candidates (candidates)
   (cl-assert (cl-every #'stringp candidates))
   (unless (company-call-backend 'sorted)
     (setq candidates (sort candidates 'string<)))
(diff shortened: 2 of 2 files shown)


dgutov:
Sorry, I don't think we're going to do this as proposed.

But you can use a `company-transformers` element for this purpose. You're welcome to also propose such a function for inclusion.

PR #1295, opened 2022-01-09
Update copyright information

yugaego:
I inferred the copyright year should be added only to the files updated during that year (approx. > 15 lines change).
Is it correct?

This patch adds copyright information to the makefiles and /doc/images according to the approach
suggested in the `<emacs-source>/admin/notes/copyright` file:
- Add copyright to the files larger than 15 lines.
- Add README file to each folder with the images.

### DIFF
--- a/company-abbrev.el
+++ b/company-abbrev.el
@@ -1,6 +1,6 @@
 ;;; company-abbrev.el --- company-mode completion backend for abbrev
 
-;; Copyright (C) 2009-2011, 2015, 2021  Free Software Foundation, Inc.
+;; Copyright (C) 2009-2011, 2013-2015, 2021  Free Software Foundation, Inc.
 
 ;; Author: Nikolaj Schumacher
 
--- a/company-bbdb.el
+++ b/company-bbdb.el
@@ -1,6 +1,6 @@
 ;;; company-bbdb.el --- company-mode completion backend for BBDB in message-mode
 
-;; Copyright (C) 2013-2014, 2016  Free Software Foundation, Inc.
+;; Copyright (C) 2013-2016, 2020  Free Software Foundation, Inc.
 
 ;; Author: Jan Tatarik <jan.tatarik@gmail.com>
 
--- a/company-capf.el
+++ b/company-capf.el
@@ -1,6 +1,6 @@
 ;;; company-capf.el --- company-mode completion-at-point-functions backend -*- lexical-binding: t -*-
 
-;; Copyright (C) 2013-2019, 2021  Free Software Foundation, Inc.
+;; Copyright (C) 2013-2021  Free Software Foundation, Inc.
 
 ;; Author: Stefan Monnier <monnier@iro.umontreal.ca>
 
--- a/company-clang.el
+++ b/company-clang.el
@@ -1,6 +1,6 @@
 ;;; company-clang.el --- company-mode completion backend for Clang  -*- lexical-binding: t -*-
 
-;; Copyright (C) 2009, 2011, 2013-2019  Free Software Foundation, Inc.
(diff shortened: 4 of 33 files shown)


dgutov:
Thanks for the initiative. :-)

> Add copyright to the files larger than 15 lines

That seems too complicated (and I'm not sure if this is the right criterion). Updating files that were changed at all would be fine, too.

yugaego:
> That seems too complicated (and I'm not sure if this is the right criterion). 

I mean the following sentences from the `<emacs>/admin/notes/copyright` file. No strong opinion regarding this on my side anyway, and Company probably is not strictly obliged to follow all the Emacs rules, so I may simply not be bothered to add them.

> Every non-trivial file distributed through the Emacs repository should be
> self-explanatory in terms of copyright and license. 
> ...
> The definition of triviality is a little vague, but a rule of thumb is
> that any file with less than 15 lines of actual content is trivial.
> ...
> It can be helpful to put a reminder comment at the start of a trivial
> file, eg: "add a license notice if this grows to > 10 lines of code".

<br>

> Updating files that were changed at all would be fine, too.

Okay.
I've clarified for myself that in Emacs copyright year is changed in all the files just to save trouble for people making changes to the files during the year. In the case of Company, not all the files get updated during the year and it has not that many files, so it makes sense to change copyright year only when needed.

yugaego:
What if just add one line at the top of the makefiles:
`# Copyright (C) 2021  Free Software Foundation, Inc.`

And combine monstrosities such as `2009-2011, 2013-2015, 2020-2021` to `2009-2015, 2020-2021`?

Would it be fine with you?

dgutov:
> What if just add one line at the top of the makefiles:

That might not mean much, or be misleading (like "all rights reserved") without the license notice.

> And combine monstrosities such as 2009-2011, 2013-2015, 2020-2021 to 2009-2015, 2020-2021?

I don't understand this issue very well, but, for example, `src/w32menu.c` has a notice like that:

```
   Copyright (C) 1986, 1988, 1993-1994, 1996, 1998-1999, 2001-2022 Free
```

So there must be a point to this?

dgutov:
> I mean the following sentences from the <emacs>/admin/notes/copyright file.

The rule is more complicated than just about 15 lines, though. And "15" is a rough guideline anyway. So let's just mark all changed files.

yugaego:
> So let's just mark all changed files.

OK 👍🏻 

> or be misleading (like "all rights reserved") without the license notice.

Ah, yes, it makes sense and explains why

> Copyright (C) 2007-2022 Free Software Foundation, Inc.
> See the end of the file for license conditions.

is often used: keep the copyright notice at the top short, but still put the license at the bottom.

> > Copyright (C) 1986, 1988, 1993-1994, 1996, 1998-1999, 2001-2022 Free

> So there must be a point to this?

Well, I don't know too, it always seemed odd to me. That might be due to the file/law history or older FSF/Emacs rules probably; since currently it's documented, as a common rule, to add new year to each GNU Emacs file. The strategy chosen for Company seem to be common in the "outer-space" anyway, so it should be fine too.

yugaego:
And done. Past years with the changes were missed in many of the files, so extended to mark them too.

dgutov:
Thanks!

dgutov:
Did you do it by hand, or with a script of some kind?

If it's the latter, it might be worth it to put it in Makefile as a new task.

yugaego:
> it might be worth it to put it in Makefile as a new task.

Yes, initially I thought to adapt a similar Emacs's script (`build-aux/update-copyright`), but ended-up doing it by hands with `git log --stat` (or similar) information.

dgutov:
That's a fair amount of work. Thanks again!

PR #1311, opened 2022-03-09
Add php keywords

zonuexe:
This list contained keywords that are not PHP reserved words today.

I propose adding the following keywords:

 * [Keywords and compile-time constants](https://www.php.net/manual/reserved.keywords.php)
 * [Type name for type declaration](https://www.php.net/manual/language.types.declarations.php)
 * Frequently mentioned classes/interfaces in code among [Predefined Classes](https://www.php.net/manual/reserved.classes.php)

This PR is same as https://github.com/minad/cape/pull/31.


dgutov:
Thanks!

PR #1317, opened 2022-03-29
Improve company code to search remote gtags remote executable.

Ergus:
Enable to use the customs company-gtags-executable as
connection-local-variables...

Simplify and improves the code and conditions. Rely more in
make-local-variable and local-variable-p because
connection-local-variables are set as buffer-local-variables. Also
relies on hack-connection-local-variables-apply and
connection-local-criteria-for-default-directory.

### DIFF
--- a/company-gtags.el
+++ b/company-gtags.el
@@ -48,7 +48,7 @@
   :package-version '(company . "0.8.1"))
 
 (defvar-local company-gtags--tags-available-p 'unknown)
-(defvar-local company-gtags--executable 'unknown)
+(defvar company-gtags--executable nil)
 
 (defcustom company-gtags-modes '(prog-mode jde-mode)
   "Modes that use `company-gtags'.
@@ -70,30 +70,28 @@ completion."
 
 (defun company-gtags--executable ()
   (cond
-   ((not (eq company-gtags--executable 'unknown)) ;; the value is already cached
+   ((or (not (file-remote-p default-directory))        ;; not remote
+        (local-variable-p 'company-gtags--executable)) ;; the value is already local
     company-gtags--executable)
-   ((and (version<= "27" emacs-version)           ;; can search remotely to set
-         (file-remote-p default-directory))
-
-    (with-connection-local-variables
-     (if (boundp 'company-gtags--executable-connection)
-         (setq-local company-gtags--executable     ;; use if defined as connection-local
-                     company-gtags--executable-connection)
-
-       ;; Else search and set as connection local for next uses.
-       (setq-local company-gtags--executable
-                   (with-no-warnings (executable-find "global" t)))
-       (let* ((host (file-remote-p default-directory 'host))
-              (symvars (intern (concat host "-vars")))) ;; profile name
-
-         (connection-local-set-profile-variables
-          symvars
-          `((company-gtags--executable-connection . ,company-gtags--executable)))
-
-         (connection-local-set-profiles `(:machine ,host) symvars))
-       company-gtags--executable)))
-   (t     ;; use default value (searched locally)
-    company-gtags-executable)))
+   ((when-let* (((version<= "27" emacs-version))       ;; search remote and set connection-local
+	        (criteria (connection-local-criteria-for-default-directory))
+	        (symvars (intern (format "company-gtags--%s-vars"
+                                         (file-remote-p default-directory))))
+	        (enable-connection-local-variables t))
+      (unless (alist-get symvars connection-local-profile-alist)
+        (with-connection-local-variables
+         (let ((gtags-executable (if (local-variable-p 'company-gtags-executable)
+			             company-gtags-executable
+			           (file-name-nondirectory company-gtags-executable))))
+	   (connection-local-set-profile-variables
+	    symvars
+	    `((company-gtags--executable . ,(executable-find gtags-executable t))))
+	   (connection-local-set-profiles criteria symvars))))
+      (hack-connection-local-variables-apply criteria)
+      t)
+    company-gtags--executable)
+   (t     ;; If here it means that it is an old emacs and the value es not local...
+    (make-local-variable company-gtags--executable))))
 
 (defun company-gtags--fetch-tags (prefix)
   (with-temp-buffer


dgutov:
Hi!

First of all, could you reindent without using tabs? Reading the code in this PR is hampered by wrong-looking indentation.

Second: could you explain what is added by the call to `hack-connection-local-variables-apply`? I don't understand the connection-local stuff very well.

I see this adds the notion of "criteria" which include not only the host but the user and other stuff as well. Which might be marginally more correct. But what happens if you omit the call to `hack-connection-local-variables-apply`?

Ergus:
> Hi!
> 
> First of all, could you reindent without using tabs? Reading the code in this PR is hampered by wrong-looking indentation.
> 
Sorry. I used the default indentation...

> Second: could you explain what is added by the call to `hack-connection-local-variables-apply`? I don't understand the connection-local stuff very well.
> 
`hack-connection-local-variables-apply` is what `with-connection-local-variables` does at the beginning; it checks for the
connection-local variables that apply to `criteria`, and apply them to the local buffer. `with-connection-local-variables` just has some extra code latter to restore back the original values.

> I see this adds the notion of "criteria" which include not only the host but the user and other stuff as well. Which might be marginally more correct. But what happens if you omit the call to `hack-connection-local-variables-apply`?

What happens is that the variables are stored in `connection-local-profile-alist' and will be applied for the next files you open with similar `criteria` by default; but won't be applied to the current buffer `hack-connection-local-variables-apply` is just the way to apply  some criteria variables locally so after this we should call `(hack-connection-local-variables-apply (connection-local-criteria-for-default-directory))` but as we already recomputed criteria in the code I just put criteria instead of `(connection-local-criteria-for-default-directory)` 

Basically for all the files you open afterward; this call is not be needed because the variables will be already applied; but we are not confident if those files called company the first or after this search was already done; that's why I the conditions on the beginning try to avoid somehow unneeded calls to the inner part of the code when possible.

dgutov:
> Sorry. I used the default indentation...

This repository has a specific `dir-locals.el` entry to avoid this.

> hack-connection-local-variables-apply is what with-connection-local-variables does at the beginning; it checks for the
connection-local variables that apply to criteria, and apply them to the local buffer. with-connection-local-variables just has some extra code latter to restore back the original values.

So `hack-connection-local-variables-apply` is called twice, then?

If this change shows any significant difference in performance, you could inline the definition of `with-connection-local-variables` and then delete the part that restores the values. That would make it very short.

I guess that could be problematic, though, since it could leave a bunch of variables with changed (buffer-local) values. And some other code might not expect to see them changed in that particular buffer. I don't have a specific package in mind, but it could lead to some odd scenarios where another package might end up functioning as long as `company-gtags` is used in the same buffer, but not without.

Setting our own variable to specific value feels safer in that regard. Or just going through `with-connection-local-variables` every time, functions in the core library do that, after all. E.g. `path-separator`, `start-file-process-shell-command`, etc.

Ergus:
> > Sorry. I used the default indentation...
> 
> This repository has a specific `dir-locals.el` entry to avoid this.
> 
> 
> So `hack-connection-local-variables-apply` is called twice, then?
>
Yes, the first call is to assert that we have the latest value for company-gtags-executable that now can be a connection local. (See bellow)

And the second one is to apply the new variables (after the search) to the buffer... the values can be nil to imply that gtags was not found on the server... which is perfect to skip directly this backend I think. In general when you open a new remote buffer all the predefined connection local variables for it's criteria, automatically apply, so the with-connection-local-variables call is just to assert that the user haven`t executed a code that like the one bellow after this buffer/file was opened so there is a new value for gtags. This function (and the original version already in company-gtags master) just adds some work to define the variables latter and apply them.

But also to assert that all the buffers already opened  don't repeat the search but just "refresh" their scope if needed but lazily.
 
> If this change shows any significant difference in performance, you could inline the definition of `with-connection-local-variables` and then delete the part that restores the values. That would make it very short.
> 
That means to call hack-connection-local-variables-apply and then make the setq-local for the extra variables... connection-local-variable already do that for you.. But any way the other code (the second call) still needs to be there because the other buffers already opened will call this function latter, but will not enter in the search part (because the connection-local-profile-alist already knows the profile/criteria, so it is not needed to create and search again; just apply) 

The change is not only for performance, it is for simplicity and to allow the user to manually set the path for gtags on the remote nodes with connection-local-variables throw the same customs than locally.

For example the user could have something like this in the config:
```
(connection-local-set-profile-variables
	    myvars
	    `((company-gtags-executable . "/my/not/usual/path/global")))
(connection-local-set-profiles '(:machine myserver) myvars))))
```

This is the way to specify that for that server, the custom `company-gtags-executable` is in `/my/not/usual/path/global`...  And now this code is aware of that, so the user can specify for every machine/user/etc... if needed

> I guess that could be problematic, though, since it could leave a bunch of variables with changed (buffer-local) values. And some other code might not expect to see them changed in that particular buffer. I don't have a specific package in mind, but it could lead to some odd scenarios where another package might end up functioning as long as `company-gtags` is used in the same buffer, but not without.
> 
I don't think this is an issue because all the files that will open after the profile is set; will use these variables if the criteria apply to them (same for all the predefined variables for example, from the init file). Actually this code only runs 1) the first time to do the search, 2) the first time an already opened buffer calls this function after 1), but just applies and don't search. 

Any other buffer/file that opens a remote with this criteria latter won even get into this part and exit directly in the  `(local-variable-p 'company-gtags--executable)` condition

> Setting our own variable to specific value feels safer in that regard. Or just going through `with-connection-local-variables` every time, functions in the core library do that, after all. E.g. `path-separator`, `start-file-process-shell-command`, etc.

That's what we do. We add the variable company-gtags--executable. The code is the same we had before, just that instead of calling with-connection-local-variables and checking inside the scope and using setq-local... we just leave that work to the connection-local variable infrastructure, we check from outside if the scope already exists, else, we create it and finally we apply it only once/buffer already oppened.

BTW: I am not using this module anymore because I found pretty easy to add a capf backend with my new module... so, company uses it pretty nicely throw the `company-capf` backend.

dgutov:
Should we just drop the `(local-variable-p 'company-gtags--executable)` check, then? And always go through `with-connection-local-variables` in Emacs 27+? As mentioned above, `(path-separator)` does that. That probably implies that performance should be good enough.

> BTW: I am not using this module anymore because I found pretty easy to add a capf backend with my new module... so, company uses it pretty nicely throw the company-capf backend

Good to know. I'll take that to mean that this PR isn't urgent, too.

PR #1320, opened 2022-04-06
company-capf: only rm tags-capf from capf if we use company-etags

### LINKED ISSUE #1318: company-capf messes with completion-at-point-functions
#### Describe the issue
I am transitioning to use the `company-capf` backend exclusively.  Now I am working on tweaking my usage of the `tags-completion-at-point-function` and noticed that it's removed from the `capf` default value in `company--capf-data-real`.

I understand the rationale, the defaults of `capf` clash with the defaults of `company-backends`.  But I deliberately removed tags-related backends from the latter --- to me, it's a bug that company messes with my `capf` value.

#### Steps to reproduce
```elisp
(require 'company)
(setq company-backends '(company-capf))
```
and check that no matter what you do (company-wise, that is), `tags-completion-at-point-function` is never called.

#### Expected behavior
I think there are a few possibilities.

1. Much preferred: Just do not mess with the `capf` value at all.
    a) Present a warning message in `company-etags` initialization if `capf` contains the `tags-capf..`
    b) Remove the `company-etags` backend from the `company-backends` default and put a note into its docstring.
2. Only remove `tags-capf` if `company-backends` contains `company-etags`.. somewhere.

Option 2 might use sth like
```elisp
(defun company--capf-data-real ()
  (cl-letf* (((default-value 'completion-at-point-functions)
              ;; Ignore tags-completion-at-point-function because it subverts
              ;; company-etags in the default value of company-backends, where
              ;; the latter comes later.
              (if (memq 'company-etags (flatten-tree company-backends))
                  (remove 'tags-completion-at-point-function
                          (default-value 'completion-at-point-functions))
                (default-value 'completion-at-point-functions)))
[...]
```
but that uses a `subr` function only present in Emacs 27.1.. (was just a quick writeup).

WDYT?

ilohmar:
Fixes #1318.

### DIFF
--- a/company-capf.el
+++ b/company-capf.el
@@ -57,13 +57,25 @@ so we can't just use the preceding variable instead.")
               (list (current-buffer) (point) (buffer-chars-modified-tick) data))
         data))))
 
+(defun company--contains (elt lst)
+  (when-let ((cur (car lst)))
+    (cond
+     ((symbolp cur)
+      (or (eq elt cur)
+          (company--contains elt (cdr lst))))
+     ((listp cur)
+      (or (company--contains elt cur)
+          (company--contains elt (cdr lst)))))))
+
 (defun company--capf-data-real ()
   (cl-letf* (((default-value 'completion-at-point-functions)
-              ;; Ignore tags-completion-at-point-function because it subverts
-              ;; company-etags in the default value of company-backends, where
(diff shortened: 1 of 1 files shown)


dgutov:
Cool, thanks!

PR #1321, opened 2022-04-23
company-yasnippet: avoid choose value prompts when fetching snippet doc

galeo:
Avoid choose value prompts when fetching snippet documentation by setting local buffer variable `yas-prompt-functions` to `'(yas-no-prompt)`. This is needed when a snippet field has multiple values to choose.

### DIFF
--- a/company-yasnippet.el
+++ b/company-yasnippet.el
@@ -119,6 +119,7 @@ It has to accept one argument: the snippet's name.")
     (with-current-buffer (company-doc-buffer)
       (let ((buffer-file-name file-name))
         (yas-minor-mode 1)
+        (set (make-local-variable 'yas-prompt-functions) '(yas-no-prompt))
         (condition-case error
             (yas-expand-snippet (yas--template-content template))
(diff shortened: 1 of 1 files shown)


dgutov:
Hi!

Are you sure this is a good change for all (or most) of our users? Personally, having to make a choice when a template calls for it doesn't sound too bad. At least if I have chosen to insert said template using `RET`.

Do you perchance use `company-tng-mode`?

galeo:
> Personally, having to make a choice when a template calls for it doesn't sound too bad. At least if I have chosen to insert said template using `RET`.

It just affects getting the candidate documentation in the company doc buffer. When browsing the completion candidates, the completion frontend fetches the doc of current item and displays it. There is no need for the user to select the field value at this time, it will break the current code completion.

> Do you perchance use `company-tng-mode`?
I don't use `company-tng-mode`.

Thanks.

dgutov:
Ah! Thanks for clarification.

PR #1324, opened 2022-07-02
CI: Run tests on latest releases

yugaego:
Add `Emacs 28.1` and switch to `ubuntu-latest` (still 20.04 now; 22.04 environment is in beta).


dgutov:
Thanks!

PR #1326, opened 2022-07-04
Update company-keywords.el

cordelia04m:
add scheme-mode's keywords


cordelia04m:
Done. I saw failed tests in the GH workflow, will this be a problem?

cordelia04m:
This should fix the fail tests, let's see how it'll go.

PR #1330, opened 2022-07-20
company-tempo: Add snippet hint

condy0919:


yugaego:
@dgutov , any objections on this PR? Thanks.

dgutov:
@yugaego Thanks for the reminder, I missed this somehow.

@condy0919 Sorry.

I like the part which adds `kind`. But about the `annotation` field, doesn't that look extraneous to you? We'll already show it's a snippet via the `kind` icon.

`company-yasnippet` uses `annotation` to show snippet names. Do Tempo snippets perhaps have a similar notion of names?

condy0919:
> `company-yasnippet` uses `annotation` to show snippet names. Do Tempo snippets perhaps have a similar notion of names?

Yes and No.

``` emacs-lisp
(defvar c-tempo-tags nil)

(tempo-define-template "c-ifndef"
                       '("#ifndef " (P "ifndef-clause: " clause) > n
                         "#define " (s clause) n> p n
                         "#endif // " (s clause) n>
                         )
                       "ifndef"
                       "Insert #ifndef #define #endif directives"
                       'c-tempo-tags)
```

The first argument of `tempo-define-template` is NAME, but it's used as the created function's name (`tempo-templat-c-ifndef` in the above code). What the above code does is to cons `"ifndef"` with `tempo-templat-c-ifndef`, funcall `tempo-templat-c-ifndef` after completion. `substring("tempo-templat-c-ifndef", 18)`returns the NAME when it's passed to `tempo-define-template`.

If we would use [18, -1) as `annotation` is still open to discussion. For myself, No use and let it simple.

dgutov:
> If we would use [18, -1) as annotation is still open to discussion. For myself, No use and let it simple.

Just add `kind`, then?

condy0919:
> > If we would use [18, -1) as annotation is still open to discussion. For myself, No use and let it simple.
> 
> Just add `kind`, then?

tempo has no equivalent snippet name, but it can be retrieved from the completion function. But I argue is it worth to do that, what's your thought?

dgutov:
> But I argue is it worth to do that, what's your thought?

If you want my opinion, please implement this locally, take a screenshot and post it here.

> Using "" as annotation is still helpful since the tempo item can be identified from other completion items. e.g., "main" is a tempo, it can also be a "main" function.

I think you're missing some text here.

condy0919:
Currently use `<Snippet>` as the annotation.

dgutov:
What about the `[18, -1)` approach?

condy0919:
re-pushed. Just the `kind` change.

dgutov:
Thanks!

PR #1333, opened 2022-07-29
fix company-clang-version function

yuchunzhou:
hello @dgutov, this pull request fixs the #1332 issue, please merge it.


dgutov:
Seems like the idea here was wrong, since `clangd` isn't, in fact, able to substitute for `clang`.

PR #1334, opened 2022-07-30
company-show-doc-buffer: Add prefix argument toggle-auto-update

nemethf:
This is my new attempt at implementing and documenting this feature. Feedback is very much welcome. 

---

`company-show-doc-buffer` now can accept a prefix argument to toggle a new
variable `company-auto-update-doc`.  When this variable is non-nil, it keeps the
documentation buffer up-to-date whenever the selection changes (#1331).

### DIFF
--- a/company.el
+++ b/company.el
@@ -1340,7 +1340,14 @@ can retrieve meta-data for them."
             (and (symbolp this-command)
                  (string-match-p "\\`company-" (symbol-name this-command)))))))
 
+(defvar company-auto-update-doc nil
+  "If non-nil, company updates the doc-buffer whenever the selection changes.
+Use `company-show-doc-buffer' to toggle this variable.")
+
 (defun company-call-frontends (command)
+  (when (and company-auto-update-doc
+             (memq command '(update show)))
+    (company-show-doc-buffer))
   (cl-loop for frontend in company-frontends collect
            (condition-case-unless-debug err
                (funcall frontend command)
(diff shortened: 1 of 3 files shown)


company.el
```diff
@@ -1366,6 +1373,9 @@ update if FORCE-UPDATE."
   (when (or force-update (not (equal selection company-selection)))
     (setq company-selection selection
           company-selection-changed t)
+    (when company-auto-update-doc
```
dgutov, line 1376: Do you still need this call here? What happens if you remove it?

nemethf: Oh, no! This has just accidentally crept back in when I moved from elpa sources to a git fork.  I'll force-push a new version soon.  I'm sorry.

company.el
```diff
@@ -1340,7 +1340,14 @@ can retrieve meta-data for them."
...
                  (string-match-p "\\`company-" (symbol-name this-command)))))))
 
+(defvar company-auto-update-doc nil
+  "If non-nil, company updates the doc-buffer whenever the selection changes.
+Use `company-show-doc-buffer' to toggle this variable.")
+
 (defun company-call-frontends (command)
+  (when (and company-auto-update-doc
+             (memq command '(update show)))
+    (company-show-doc-buffer))
```
dgutov, line 1350: I'm not quite happy about this call (it does something other than what the function name advertises), but the effect is probably what we want for this feature.

Looking good otherwise, thanks!

nemethf: One possibility to solve your concern is to refactor these three lines into a company-doc-frontend and append that to company-frontends or a new company-internal-frontends list. This solution, however, seems unjustifiably complicated to me.

dgutov: Two other, lighter options:

* Create a wrapper function for `company-call-frontends` which would start by calling this code.
* Create a helper with this code, which would be invoked explicitly before every call to `(company-call-frontends 'update)`. All 4 of them.

nemethf: The first one is more maintainable, I think.  But I'm unable to come up with a good name for the wrapper function.  `company-frontend-event` or `company-handle-event`?

dgutov: `company-frontends-update`? Maybe.

I couldn't think of a good name, that's why I didn't suggest this in the first place. :-)

PR #1337, opened 2022-08-12
Adjust recently added docs

yugaego:
Let a doc-bore to suggest the adjustments.
There was also a notion to replace the argument name `toggle-auto-update` by `arg` or `toggle` but it was dropped.


dgutov:
Thanks!

PR #1353, opened 2022-12-06
Add keywords for Elixir

ashton314:
Hi there! I haven't signed the FSF copyright assignment papers—could someone please help me get that started?


dgutov:
I think it's not required for this PR (the list of keywords is more data than code), but for any future contributions, here's the instructions: https://github.com/coreutils/gnulib/blob/master/doc/Copyright/request-assign.future

Just use "Emacs" as the answer to the first question.

PR #1366, opened 2023-01-29
Bits of Maintenance

yugaego:
CI filter by repository is not needed since Actions are disabled by default on forks. 

Use the latest libs to prevent [GitHub Actions warnings](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/).


dgutov:
Thank you :pray:

PR #1387, opened 2023-06-11
company--posn-col-row: fix wrong col value with RTL direction

nagyation:
`posn-col-row` return value relative to the left, which make the overlay appears in completely wrong position when using a RTL language and `bidi-paragraph-direction` set no nil or forcing right-to-left.

To reproduce set bidi-paragraph-direction to right-to-left.

### DIFF
--- a/company.el
+++ b/company.el
@@ -1061,6 +1061,10 @@ means that `company-mode' is always turned on except in `message-mode' buffers."
         (row (cdr (or (posn-actual-col-row posn)
                       ;; When position is non-visible for some reason.
                       (posn-col-row posn)))))
+    ;; posn-col-row return value relative to the left
+    (when (eq (current-bidi-paragraph-direction) 'right-to-left)
+      (let ((ww (window-body-width)))
+        (setq col (- ww col))))
     (when (bound-and-true-p display-line-numbers)
       (cl-decf col (+ 2 (line-number-display-width))))
     (cons (+ col (window-hscroll)) row)))
--- a/test/frontends-tests.el
+++ b/test/frontends-tests.el
@@ -59,6 +59,22 @@
         (should (string= (overlay-get ov 'company-display)
                          "  123 \nc 45  c\nddd\n")))))))
 
+(ert-deftest company-pseudo-tooltip-show-at-point-RTL ()
+  :tags '(interactive)
+  (with-temp-buffer
+    (save-window-excursion
+    (set-window-buffer nil (current-buffer))
+    (setq bidi-display-reordering t)
+    (setq bidi-paragraph-direction 'right-to-left)
+    (insert "انا مثال للكتابة بالعربية")
+    (search-backward "مثال")
+    (let ((company-candidates-length 2)
+          (company-candidates '("123" "45"))
+          (company-backend 'ignore))
+      (company-pseudo-tooltip-show-at-point (point) 0)
+      (let ((ov company-pseudo-tooltip-overlay))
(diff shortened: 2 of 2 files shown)


dgutov:
Hi!

It would be nice to have a test which would ensure we don't break this later.

Also: is everything else working fine for you in RTL buffers? Some quick testing seems to show that the preview frontend might be positioned at the wrong side. But I'm not really sure, not being used to right-to-left texts.

nagyation:
This how it looks:
‏لقطة الشاشة 2023-06-13 في 10 10 18 م

This with the patch
‏لقطة الشاشة 2023-06-13 في 10 10 34 م

I will work on implementing a test for the tooltip position with direction set to right-to-left, from quick search I think I could find something similar from [frontends-tests.el](https://github.com/company-mode/company-mode/blob/master/test/frontends-tests.el) to use as a reference.

dgutov:
Thank you, yes, that looks better. I meant a different frontend, though. This one is only shown, by default, when there is just one completion remaining.

But maybe it's premature to worry about: does `M-x company-dabbrev` work for you at all, with Arabic?

dgutov:
> I will work on implementing a test for the tooltip position with direction set to right-to-left, from quick search I think I could find something similar from [frontends-tests.el](https://github.com/company-mode/company-mode/blob/master/test/frontends-tests.el) to use as a reference.

Thanks. The test can be as high or low level as you see more fitting.

nagyation:
> Thank you, yes, that looks better. I meant a different frontend, though. This one is only shown, by default, when there is just one completion remaining.

Yes, the inline completion of the last word works fine.

> 
> But maybe it's premature to worry about: does `M-x company-dabbrev` work for you at all, with Arabic?

It works well, it shows all Engilsh and Arabic words.

dgutov:
> It works well, it shows all Engilsh and Arabic words.

Excellent.

Let me know if you need any help for writing a test, the one we discussed previously.

nagyation:
I think somehow tests are not affected by `bidi-paragraph-direction`, I tried to setq it locally, but I can't reproduce on a test. I think this might be because the fact that this control direction on display time. do you have ideas on how to make it work?

dgutov:
That depends on what you are looking for. If you were testing `buffer-substring`, for example, and expecting it to return a different value, then this would not work because, indeed, `bidi-paragraph-direction` takes effect on redisplay.

But you can test the return value of the function being changed here. And/or the display string of the popup. But the former sounds easier.

nagyation:
I didn't know that `make test` doesn't recompile, so I was running the code with the fix, silly me..

Sorry, I hadn't much time last few days to work on this. I tested the company-column value, I think this is sufficient with finding the issue.

dgutov:
Looking good, thank you.

yugaego:
The test raises byte-compilation warning with snapshot.

```
test/frontends-tests.el:491:6: Error: ‘put-text-property’ on constant string (arg 5)
```

Does [suppression](https://github.com/emacs-mirror/emacs/blob/f559bd1248a265662665c462b750eb5fc64dd811/etc/NEWS#L636) look good to you in this case or maybe another approach could be used in the code to avoid the warning? On the other hand, maybe checking for compilation warnings on tests is an overkill. WDYT?

The warning was added in [bfc07100](https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=bfc07100d28d0f687da0a1dd5fdfa42a92a93f88).

dgutov:
It's a good warning, even if it's not too important in tests. We can fix it by calling `copy-sequence` first on the literal value.

PR #1391, opened 2023-07-02
Fix byte-compiler warning

yugaego:
Adds `copy-sequence`, as was suggested in https://github.com/company-mode/company-mode/pull/1387#issuecomment-1615997805


dgutov:
Thanks!

Issue #1388, opened 2023-06-21
The tooltip scrollbar alignment is broken on the candidate includes `…` character


templateK:
#### Output of the command `M-x company-diag`
Emacs 28.2 (x86_64-apple-darwin18.6.0) of 2022-10-17 on foo-iMac
Company 0.9.13

company-backends: (company-bbdb company-semantic company-cmake company-capf company-clang company-files
              (company-dabbrev-code company-gtags company-etags company-keywords)
              company-oddmuse company-dabbrev)

Used backend: company-capf
Value of c-a-p-f: (eglot-completion-at-point t)
Major mode: rust-mode
Prefix: stop
Completions: none

#### Describe the issue

The tooltip scrollbar alignment is broken on the candidate includes `…` character 
when `set-language-environment`  is set to CJK environment.

#### Steps to reproduce
1. install/setup rust-analyzer, make new rust project.

2. start `emacs -q` and  evaluate a buffer with following elisp code.
```elisp
(package-initialize)
(require 'use-package)
(use-package company :config (global-company-mode))
(use-package eglot)
(use-package rust-mode :config (add-hook 'rust-mode-hook 'eglot-ensure))
(set-language-environment "Chinese-BIG5")
```

3. try auto complete with following rust code.
```rust
fn main() {
    let s = String::new();
    s.
}
```

#### Expected behavior
The scrollabr should align.

#### Screenshots

On default language environment.
![company_default](https://github.com/company-mode/company-mode/assets/3207776/690b8d8e-4dcc-4dd3-9cd9-bafbfa7d7b48)

On Chinese-BIG5/Japanese/Korean language environment.
![company_cjk](https://github.com/company-mode/company-mode/assets/3207776/15d7bc86-3918-4357-8f0f-0003b8c1356a)

#### Additional context
Add any other relevant information here.

templateK:
@dgutov Is this issue will be addressed?

templateK:
I assume this issue won't be address as there's no response from the project owner and I found a workaround what I think not a solid one. so closing this issue. Feel free reopen this issue whenever you have a time for this.

dgutov:
Hi! It will need some investigation, possible with someone who understands the multi-language environment in Emacs better.

But perhaps there are some easier solutions as well.

In your testing, does `(company--clean-string "…")` when the environment is Chinese from when it is English?

dgutov:
I don't have RLS installed, so if you have a simple reproduction available, that could help speed this up.

[...]

Issue #693, opened 2017-08-16
Broken Completion Menu in `markdown-mode`


evgeniysharapov:
There's something wrong with company completion front-end. I am using it in `markdown-mode` and here's the example of how it looks. This is probably clashes of the overlays in both modes. 

![2017-08-16_09-39-25](https://user-images.githubusercontent.com/39337/29366332-c2a4cb04-8267-11e7-8b83-efb45dd2fbce.png)

`M-x company-diag` gives this 

```
Emacs 25.2.1 (x86_64-w64-mingw32) of 2017-04-24 on LAPHROAIG
Company 0.9.4

company-backends: (company-capf
 (company-dabbrev-code company-gtags company-etags company-keywords)
 company-files company-dabbrev)

Used backend: company-dabbrev
Major mode: markdown-mode
Prefix: "fo"
Completions:
  "foundation"
  "font"
  "fonts"
  "font"
  "font"
  "font"
  "fonts"
  "font"
  "font"
  "font"
  "font"
  "font"
  "font"
  "font"
  "font"
  "font"
  "font"
  "font"
  "font"
  "font"
  "font"
  "font"
  "font"
  "font"
  "font"
  "font"
  "font"
  "font"
  "font"
  "font"
  "font"
  "font"
  "font"
  "following"
  "following"
  "forms"
  "force"
  "following"
  "folder"
  "folder"
  "following"
  "folder"
  "following"
  "force"
  "force"
  "following"
  "force"
  "found"
  "following"
  "following"
  "former"
  "following"
  "following"
  "found"
  "following"
  "following"
  "force"
  "formats"
  "following"
  "foul"
  "fodder"
  "fodder"
  "fodder"
  "forward"
  "following"
  "forward"
  "forward"
  "force"
  "force"
  "format"
  "format"
  "forward"
  "format"
  "format"
  "forward"
  "forward"
  "forward"
  "forward"
  "forward"
  "forward"
  "forward"
  "forward"
  "fold"
  "form"
  "forward"
  "forward"
  "forward"
  "format"
  "forward"
  "following"
  "forward"
  "format"
  "forward"
  "formed"
  "format"
  "forward"
  "forward"
  "formatting"
  "format"
  "format"
  "format"
  "format"
  "format"
  "format"
  "format"
  "format"
  "format"
  "forward"
  "formatting"
  "format"
  "format"
  "format"
  "format"
  "format"
  "format"
  "format"
  "format"
  "format"
  "format"
  "found"
[...]

PR #1394, opened 2023-07-13
Improve popup rendering with wide characters in buffer or popup

dgutov:
This feature started out in the issue #1388.

The method is fairly different from the previous one, so this could use more testing.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,9 @@
 
 # Next
 
+* The tooltip uses a more complex rendering approach, supporting double
+  width/CJK characters, as well as buffer text of different sizes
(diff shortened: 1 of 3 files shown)


dgutov:
Some remaining problems:

- [ ] `text-scale-mode` is not supported (the popup does not scale properly). `global-text-scale-adjust` seems to work fine.
- [ ] The handling of `invisible` textprop is still wonky: #693 regresses here.
- [ ] The new code is designed to handle characters wider than 1. It can glitch with characters displayed narrower than normal.

dgutov:
Buffer-local text scaling is a bit of a pain to support, and other parts of the code rely on the col/row addressing anyway (the `select-mouse` action, for example), so it'll remain an exercise of the future.

PR #1403, opened 2023-09-24
Fix #1402: Restore window-configuration after auto-update-doc

### LINKED ISSUE #1402: `company-auto-update-doc` leaves open the buffer that it used
Hi,

TIL about `company-auto-update-doc`. Very handy!

However, after I'm done completing (whether it is by successfully choosing a completion, or aborting my intent), the buffer that was created, in my case `*cider-doc*`, remains open. This is problematic because leaving a buffer that wasn't before displaces my other buffers.

This isn't the case if using `<f1>` (`company-show-doc-buffer`) manually with each completion - with f1, the `*cider-doc*`  buffer always ends up closed.

I can show a GIF if that helps, but I believe the description is clear enough.

Finally, in technical terms, CIDER is specifying `:company-doc-buffer #'cider-create-doc-buffer`. We (the CIDER maintainers) certainly are completely fine with this ancilliary buffer being destroyed. It's very cheap to re-create.

Thanks - V

nemethf:


dgutov:
As discussed, I went with a slightly different solution (f5a1ab9, f3ee625).

So I'm closing this PR, but let's continue work on resolving whatever problems remain.

Issue #840, opened 2018-11-02
Feature Request: Allow Filtering yasnippet completions when triggered by trigger sequence


wberrier:
This seems somewhat related to #485 , but the idea is that I'm grouping company-lsp/cquery (c++) and company-yasnippet together via:

```elisp
(push '(company-lsp :with company-yasnippet) company-backends)
```

Completing without input (prefix) shows all globals and snippets as expected.

But, when I have completions from a trigger character ('.' or '->') then all the snippets show up in the completions, which is never desired behavior in this setup.

It would be great to only show snippets in completions if completion list isn't triggered by a trigger sequence.

I've considered trying to set up `company-transformers` to perform this filter, and maybe that's a way to implement this?  Not sure...

Thoughts?

wberrier:
Also note, using `:separate` as suggested in #485 does indeed help by putting all the snippets at the bottom, but I think it would be preferable if the snippets weren't in the list at all.

dgutov:
Try this:

```lisp
(defun company-yasnippet-unless-member-access (command &optional arg &rest ignore)
  (if (eq command 'prefix)
      (let ((prefix (company-yasnippet 'prefix)))
        (and prefix
             (save-excursion
               (forward-char (- (length prefix)))
               (not (looking-back (rx (or "." "->")) (line-beginning-position))))
             prefix))
    (company-yasnippet command arg)))
```

And use this backend instead of `company-yasnippet` directly.

wberrier:
That works great, thanks!

Some questions:

Is there a way to automatically use all the "triggers" specified for that particular language?  Seems like different languages are going to have their particular cases when it's not appropriate to show snippets...

For example, here's some additional cases for c++ where snippets probably shouldn't be shown:

* `std::`
* `#include <`
* `#include "`

Also, is this something that would be provided with company-yasnippet (and have the lookback chars be a defcustom, or language specific alist, etc...), or should I just put it in my local config?

dgutov:
> Is there a way to automatically use all the "triggers" specified for that particular language?

If there were an existing convention to specify them in language mode definitions, we could. I'm not aware of one, though.

> or should I just put it in my local config?

This, at least for now. And I also think it works better as a wrapper backend, because whether we want to avoid snippets after trigger characters depends on the presence of other backends (and whether they provide snippets for completions).

wberrier:
Sounds good, and thanks, it works much better.

I'll leave it up to you whether you want to close this issue or not...

[...]

Issue #641, opened 2017-02-13
company-yasnippet and returning "" for prefix


jojojames:
```elisp
(defun company-yasnippet (command &optional arg &rest ignore)
  "`company-mode' backend for `yasnippet'."
  (interactive (list 'interactive))
  (cl-case command
    (interactive (company-begin-backend 'company-yasnippet))
    (prefix
     ;; Should probably use `yas--current-key', but that's bound to be slower.
     ;; How many trigger keys start with non-symbol characters anyway?
     (and (bound-and-true-p yas-minor-mode)
          (company-grab-symbol)))
    (annotation
     (concat
      (unless company-tooltip-align-annotations " -> ")
      (get-text-property 0 'yas-annotation arg)))
    (candidates (company-yasnippet--candidates arg))
    (no-cache t)
    (post-completion
     (let ((template (get-text-property 0 'yas-template arg))
           (prefix-offset (get-text-property 0 'yas-prefix-offset arg)))
       (yas-expand-snippet (yas--template-content template)
                           (- (point) (length arg) prefix-offset)
                           (point)
                           (yas--template-expand-env template))))))
```

It looks like for 'prefix case, company-yasnippet always returns at least "".

This is somewhat painful when setting minimum-prefix-length to 1 and after something like "." completion.

this.| triggers company-yasnippet because the prefix will be "" but I'm not sure I'll ever want yasnippet to trigger in that case.

Changing the function to be,
```elisp
  (defun company-yasnippet (command &optional arg &rest ignore)
    "`company-mode' backend for `yasnippet'."
    (interactive (list 'interactive))
    (cl-case command
      (interactive (company-begin-backend 'company-yasnippet))
      (prefix
       ;; Should probably use `yas--current-key', but that's bound to be slower.
       ;; How many trigger keys start with non-symbol characters anyway?
       (and (bound-and-true-p yas-minor-mode)
            (let ((company-symbol (company-grab-symbol)))
              (if (string-equal company-symbol "")
                  nil
                company-symbol))))
      (annotation
       (concat
        (unless company-tooltip-align-annotations " -> ")
        (get-text-property 0 'yas-annotation arg)))
      (candidates (company-yasnippet--candidates arg))
      (no-cache t)
      (post-completion
       (let ((template (get-text-property 0 'yas-template arg))
             (prefix-offset (get-text-property 0 'yas-prefix-offset arg)))
         (yas-expand-snippet (yas--template-content template)
[...]

Issue #553, opened 2016-07-13
unexpected completion item list


netjune:
mode: c-mode
backends: (setq company-backends '((company-clang company-dabbrev-code)))
other setting: (setq company-minimum-prefix-length 3)
version: 0.9.0 and above

1. create a c-mode buffer
2. yank the following code into buffer
3. goto end of buffer(after "st"), insert the char '.' to trigger completion
4. all symbols in the buffer is in the compeletion item list

-------------------------------
struct st_test
{
	int long_name_var1;
	int long_name_var2;
};

void test()
{
	struct st_test st;

	st

dgutov:
What did you expect?

netjune:
I expect all items in completion list be A+B.
A: all items when (setq company-backends '(company-clang)). 2
B: all items when (setq company-backends '(company-dabbrev-code)). 0
A+B=2+0=2.

But it is not. Too many items in B(I guess).

dgutov:
> B: all items when (setq company-backends '(company-dabbrev-code)). 0

What does that mean? Do you not get any completions with `company-dabbrev-code` alone? That's not what I'm seeing here.

netjune:
Yes. I set company-minimum-prefix-length to 3. So after '.' there is no completions, because it is not a symbol-char.

dgutov:
OK, so your problem is that with `company-dabbrev-code` you get no idle compeltion after `st.`. And with `company-clang`, you do.

This is working as expected. See the description of `prefix` in `company-backends`. Also try `M-x company-complete-common` to see how many completions the current backend actually returns.

netjune:
My problem:
A. (setq company-backends '(company-clang)). The completion items are as expected (2 items).
B. (setq company-backends '(company-dabbrev-code)). The completion items are also as expected (0 items because company-minimum-prefix-length is 3).
C. (setq company-backends '((company-clang company-dabbrev-code))). The completion items are NOT as expected (A + all symbol in buffer). 

A and B is right. C is not. C != A + B.

The problem seems to be that: company-minimum-prefix-length is ignored In C case.

dgutov:
Go to step B and type `M-x company-complete-common`. There will be completions.

netjune:
In B, there should be no completions because company-minimum-prefix-length is 3.
Command company-complete-common ignores company-minimum-prefix-length?

dgutov:
Like its docstring says, `company-minimum-prefix-length` only affects *idle* completion.

[...]

PR #1405, opened 2023-10-04
company--multi-backend-adapter: Consider min prefix in 'candidates' separately

dgutov:
This should help with
https://github.com/company-mode/company-mode/discussions/1398, https://github.com/company-mode/company-mode/issues/641, https://github.com/company-mode/company-mode/issues/553, https://github.com/company-mode/company-mode/issues/485, https://github.com/company-mode/company-mode/pull/1147, https://github.com/company-mode/company-mode/issues/840

With some of the above already closed, and some describing more serious scenarios than others.

Anyway, this is a halfway step to dropping framework-level caching (or moving it to backends).  For now this seems like an adequate solution, given that the most "heavy" backend these days is `company-capf`, which disables caching already. Though `company-dabbrev-code` is a close next.

### DIFF
--- a/company.el
+++ b/company.el
@@ -1165,6 +1165,8 @@ matches IDLE-BEGIN-AFTER-RE, return it wrapped in a cons."
     (error (error "Company: backend %s error \"%s\" with args %s"
                   company-backend (error-message-string err) args))))
 
+(defvar company--multi-uncached-backends nil)
+
 (defun company--multi-backend-adapter (backends command &rest args)
   (let ((backends (cl-loop for b in backends
                            when (or (keywordp b)
@@ -1182,6 +1184,22 @@ matches IDLE-BEGIN-AFTER-RE, return it wrapped in a cons."
        (company--multi-backend-adapter-candidates backends (car args) separate))
       (`sorted separate)
       (`duplicates (not separate))
+      ((and `no-cache
+            (pred (lambda (_)
+                    (let* (found
+                           (uncached company--multi-uncached-backends))
+                      (dolist (backend backends)
+                        (when
+                            (and (member backend uncached)
+                                 (company--good-prefix-p
+                                  (let ((company-backend backend))
+                                    (company-call-backend 'prefix))))
+                          (setq found t
+                                company--multi-uncached-backends
+                                (delete backend
+                                        company--multi-uncached-backends))))
+                      found))))
+       t)
       ((or `prefix `ignore-case `no-cache `require-match)
        (let (value)
          (cl-dolist (backend backends)
@@ -1200,10 +1218,16 @@ matches IDLE-BEGIN-AFTER-RE, return it wrapped in a cons."
 
 (defun company--multi-backend-adapter-candidates (backends prefix separate)
   (let ((pairs (cl-loop for backend in backends
-                        when (equal (company--prefix-str
-                                     (let ((company-backend backend))
-                                       (company-call-backend 'prefix)))
-                                    prefix)
+                        when (let ((bp (let ((company-backend backend))
+                                         (company-call-backend 'prefix))))
+                               (and
+                                ;; It's important that the lengths match.
+                                (equal (company--prefix-str bp) prefix)
+                                ;; One might override min-length, another not.
+                                (if (company--good-prefix-p bp)
+                                    t
+                                  (push backend company--multi-uncached-backends)
+                                  nil)))
                         collect (cons (funcall backend 'candidates prefix)
                                       (company--multi-candidates-mapper
                                        backend
@@ -2160,6 +2184,7 @@ For more details see `company-insertion-on-trigger' and
           company--manual-action nil
           company--manual-prefix nil
           company--point-max nil
+          company--multi-uncached-backends nil
           company-point nil)
     (when company-timer
       (cancel-timer company-timer))
--- a/test/core-tests.el
+++ b/test/core-tests.el
@@ -122,7 +122,8 @@
     (should (equal "abcd" company-common))))
 
 (ert-deftest company-multi-backend-with-lambdas ()
-  (let ((company-backend
+  (let ((company-minimum-prefix-length 1)
+        (company-backend
          (list (lambda (command &optional _ &rest _r)
                  (cl-case command
                    (prefix "z")
@@ -134,7 +135,8 @@
     (should (equal (company-call-backend 'candidates "z") '("a" "b" "c" "d")))))
 
 (ert-deftest company-multi-backend-filters-backends-by-prefix ()
-  (let ((company-backend
+  (let ((company-minimum-prefix-length 1)
+        (company-backend
          (list (lambda (command &optional _ &rest _r)
                  (cl-case command
                    (prefix (cons "z" t))
@@ -150,25 +152,29 @@
     (should (equal (company-call-backend 'candidates "z") '("a" "b" "e" "f")))))
 
 (ert-deftest company-multi-backend-remembers-candidate-backend ()
-  (let ((company-backend
+  (let ((company-minimum-prefix-length 0)
+        (company-backend
          (list (lambda (command &optional _)
                  (cl-case command
+                   (prefix "")
                    (ignore-case nil)
                    (annotation "1")
                    (candidates '("a" "c"))
                    (post-completion "13")))
                (lambda (command &optional _)
                  (cl-case command
+                   (prefix "")
                    (ignore-case t)
                    (annotation "2")
                    (candidates '("b" "d"))
                    (post-completion "42")))
                (lambda (command &optional _)
                  (cl-case command
+                   (prefix "")
                    (annotation "3")
                    (candidates '("e"))
                    (post-completion "74"))))))
-    (let ((candidates (company-calculate-candidates nil nil)))
+    (let ((candidates (company-calculate-candidates "" nil)))
       (should (equal candidates '("a" "b" "c" "d" "e")))
       (should (equal t (company-call-backend 'ignore-case)))
       (should (equal "1" (company-call-backend 'annotation (nth 0 candidates))))
@@ -179,7 +185,8 @@
       (should (equal "74" (company-call-backend 'post-completion (nth 4 candidates)))))))
 
 (ert-deftest company-multi-backend-handles-keyword-with ()
-  (let ((primo (lambda (command &optional _)
+  (let ((company-minimum-prefix-length 1)
+        (primo (lambda (command &optional _)
                  (cl-case command
                    (prefix "a")
                    (candidates '("abb" "abc" "abd")))))
@@ -195,7 +202,8 @@
                      (company-call-backend 'candidates "a"))))))
 
 (ert-deftest company-multi-backend-handles-keyword-separate ()
-  (let ((one (lambda (command &optional _)
+  (let ((company-minimum-prefix-length 1)
+        (one (lambda (command &optional _)
                (cl-case command
                  (prefix "a")
                  (candidates (list "aa" "ca" "ba")))))
@@ -213,6 +221,50 @@
       (should-not (company-call-backend 'duplicates))
       (should (equal '("aa" "ba" "ca" "ab" "bb" "cc" "bc" "ac")
                      (company-call-backend 'candidates "a"))))))
+
+(ert-deftest company-multi-backend-handles-length-overrides-separately ()
+  (let ((company-minimum-prefix-length 2)
+        (one (lambda (command &optional _)
+               (cl-case command
+                 (prefix "a")
+                 (candidates (list "aa" "ca" "ba")))))
+        (two (lambda (command &optional _)
+               (cl-case command
+                 (prefix (cons "a" 2))
+                 (candidates (list "bb" "ab")))))
+        (tri (lambda (command &optional _)
+               (cl-case command
+                 (prefix "")
+                 (candidates (list "cc" "bc" "ac"))))))
+    (let ((company-backend (list one two tri)))
+      (should (equal '("bb" "ab")
+                     (company-call-backend 'candidates "a"))))
+    (let ((company-backend (list one two tri))
+          (company--manual-prefix "a"))
+      (should (equal '("aa" "ca" "ba" "bb" "ab")
+                     (company-call-backend 'candidates "a"))))))
+
+(ert-deftest company-multi-backend-handles-clears-cache-when-needed ()
+  (let* ((company-minimum-prefix-length 2)
+         (one (lambda (command &optional _)
+                (cl-case command
+                  (prefix "aa")
+                  (candidates (list "aa")))))
+         (two (lambda (command &optional _)
+                (cl-case command
+                  (prefix (cons "aa" t))
+                  (candidates (list "aab" )))))
+         (tri (lambda (command &optional _)
+                (cl-case command
+                  (prefix "")
+                  (candidates (list "aac")))))
+         (company--multi-uncached-backends (list one tri)))
+    (let ((company-backend (list one two tri)))
+      (should
+       (equal (company-call-backend 'no-cache) t))
+      (should (equal company--multi-uncached-backends (list tri)))
+      (should (equal '("aa" "aab")
+                     (company-call-backend 'candidates "aa"))))))
 
 (ert-deftest company-begin-backend-failure-doesnt-break-company-backends ()
   (with-temp-buffer


dgutov:
This came after a new and timely reminder in the recent discussion (first link).

Aside from increased complexity, one specific downside is that `company--multi-backend-adapter` depends on the external state more (values of `company-minimum-prefix-length` and `company--manual-prefix`). Which will probably break the related recipe in `cape` by @minad. :disappointed:

minad:
Hi Dmitry,

could you please explain what this PR is doing? I have similar code in Cape (`cape-capf-super`) which has the ability to merge multiple Capfs. It is likely that the merging strategy and the prefix handling could also be improved there.

Regarding `company--multi-backend-adapter`, how will this change? I like the ability to merge multiple Company backends into one and then capfify the result via `cape-company-to-capf`. However from what I've heard, not many people use `cape-company-to-capf` since they prefer to go all-in with Capfs then.

Thank you!

minad:
> Doing that requires consulting the two global vars above, unfortunately.

So what happens if these variables are nil? This would happen when we use `company--multi-backend-adapter` as described in https://github.com/minad/cape#company-adapter. If the current behavior could be preserved for the nil variables, there should not be a problem (assuming that this doesn't somehow hold back your development).

> You'll probably need to create a wrapper around it, or direct users to first pipe every backend through cape-company-to-capf, and then combine the results using cape-capf-super. 

Yes, that's always possible. The downside is that the merged asynchronous backends won't be executed in parallel.

dgutov:
> So what happens if these variables are nil? The variables are probably uninitialized when we use company--multi-backend-adapter as described in https://github.com/minad/cape#company-adapter:

The values' lookup will happen later, when the completion table is queried. So that's not a problem.

But `company--manual-prefix` being nil means that the current completion wasn't initiated explicitly (meaning, it's idle/auto). And that means that the value of `company-minimum-prefix-length` has to take effect.

The latter var has a default numeric value (can't be nil). The problem with it is that Corfu has a different variable with the same goal (`corfu-auto-prefix`). So the capf creates this way might even function properly when used with Company (called through `company-capf`), but its behavior with Corfu can have problems.

I'm not sure what's the best solution. As per above, a Corfu-only wrapper could work. The current one will work under Company.

Not sure how to make it agnostic. Adding one more argument to the `candidates` action might help (to pass the "effective minimum prefix length to check against"). As long as most third-part backends dutifully use the `&rest _ignored` pattern, and don't break when one more argument is added...

minad:
On 10/4/23 22:26, Dmitry Gutov wrote:
> But |company--manual-prefix| being nil means that the current completion
> wasn't initiated explicitly (meaning, it's idle/auto). And that means
> that the value of |company-minimum-prefix-length| has to take effect.
> 
> The latter var has a default numeric value (can't be nil). The problem
> with it is that Corfu has a different variable with the same goal
> (|corfu-auto-prefix|). So the capf creates this way might even function
> properly when used with Company (called through |company-capf|), but its
> behavior with Corfu can have problems.

So the only problem seems to be that there could be a discrepancy
between the prefix limits? This shouldn't be hard to avoid. I could even
add a check to `cape-company-to-capf` to ensure that the values match.

> I'm not sure what's the best solution. As per above, a Corfu-only
> wrapper could work. The current one will work under Company.

It would be ideal if we could avoid creating a separate wrapper.

> Not sure how to make it agnostic. Adding one more argument to the
> |candidates| action might help (to pass the "effective minimum prefix
> length to check against"). As long as most third-part backends dutifully
> use the |&rest _ignored| pattern, and don't break when one more argument
> is added...

Right now Corfu doesn't even inform the backends if they were triggered
manually or automatically. The same is the case with Capfs called via
`company-capf` as long as they don't secretly access
`company--manual-prefix`. Maybe another Capf extension function could be
added.

dgutov:
> So the only problem seems to be that there could be a discrepancy
between the prefix limits? This shouldn't be hard to avoid. I could even
add a check to `cape-company-to-capf` to ensure that the values match.

It's about the configured value, though. Not about the text in the buffer.

> Right now Corfu doesn't even inform the backends if they were triggered
manually or automatically.

Neither did Company, before the changes in this PR (backends just reported their "prefix" along with, optionally, length override).

But I'm not seeing a better/easier way to address the linked issues.

minad:
On 10/4/23 23:18, Dmitry Gutov wrote:
>     So the only problem seems to be that there could be a discrepancy
>     between the prefix limits? This shouldn't be hard to avoid. I could even
>     add a check to |cape-company-to-capf| to ensure that the values match.
> 
> It's about the configured value, though. Not about the text in the buffer.

Yes, understood. I meant that I could check the configured values and
emit a warning, if corfu-auto-prefix and company-minimum-prefix-length
are bound and differ during execution of cape-company-to-capf. If that's
all that creates a problem?

Otoh the result will be that the multi adapter backends will be treated
as always triggered automatically if I understood you correctly, which
may be undesired? I could also turn it around and always let bind
company--manual-prefix to something in cape-company-to-capf in order to
fake manual completion.

I am not sure what is better since I don't really understand the
implications. The problem is that the manual/auto information is lacking
in cape-company-to-capf.

>     Right now Corfu doesn't even inform the backends if they were triggered
>     manually or automatically.
> 
> Neither did Company, before the changes in this PR (backends just
> reported their "prefix" along with, optionally, length override).

Well the variable company--manual-prefix has already been there? Instead
you could also pass an argument to the prefix action if you want to make
this information available as part of the public API?

dgutov:
> I meant that I could check the configured values and
emit a warning, if corfu-auto-prefix and company-minimum-prefix-length
are bound and differ during execution of cape-company-to-capf. If that's
all that creates a problem?

If we're going for an agnostic adapter, I'd imagine looking up a set of known variables from two existing UIs is not ideal.

> Well the variable company--manual-prefix has already been there? Instead
you could also pass an argument to the prefix action if you want to make
this information available as part of the public API?

See the experimental change I just pushed. If your `cape-company-to-capf` translator could work with that change in the interface (i.e. known how to compute the value `min-length` to pass along), we can try to make it work.

The parts that I don't like are that whole two actions (`candidates` and `no-cache`) get a new argument, and that the shape of the API is becoming weirder. It's going to be tricky to explain/document.

minad:
On 10/5/23 01:07, Dmitry Gutov wrote:
>     I meant that I could check the configured values and
>     emit a warning, if corfu-auto-prefix and company-minimum-prefix-length
>     are bound and differ during execution of cape-company-to-capf. If that's
>     all that creates a problem?
> 
> If we're going for an agnostic adapter, I'd imagine looking up a set of
> known variables from two existing UIs is not ideal.

Sure, but it would be a minor kludge and the consequences would be
harmless, just a warning.

>     Well the variable company--manual-prefix has already been there? Instead
>     you could also pass an argument to the prefix action if you want to make
>     this information available as part of the public API?
> 
> See the experimental change I just pushed. If your
> |cape-company-to-capf| translator could work with that change in the
> interface (i.e. known how to compute the value |min-length| to pass
> along), we can try to make it work.

The problem is - I don't know how to obtain/compute this minimum length
in the Capf. For manual it should be zero, for auto it should be the
auto prefix length (e.g. corfu-auto-prefix)? The Capf doesn't have
access to this information (at least not in an agnostic way).

If you keep using the global variables I could still do a similar thing
and bind them in cape-company-to-capf. Leaves the problem that I don't
have access to the necessary information...

> The parts that I don't like are that whole two actions (|candidates| and
> |no-cache|) get a new argument, and that the shape of the API is
> becoming weirder. It's going to be tricky to explain/document.

Yeah. I agree it is not great.

dgutov:
> Sure, but it would be a minor kludge and the consequences would be
harmless, just a warning.

What I mean is, a capf should be usable with just any UI, right? Including e.g. `ivy-completion-in-region` and those that might yet be written.

> The problem is - I don't know how to obtain/compute this minimum length
in the Capf. For manual it should be zero, for auto it should be the
auto prefix length (e.g. corfu-auto-prefix)? The Capf doesn't have
access to this information (at least not in an agnostic way).

That is indeed a problem. Like, if we're translating to a `candidates` call, then in a capf this should be a query against a completion table. And there is nowhere to pass this value through.

I suppose we could add an extension, like `:company-set-min-length`, which would be a function that takes a value from the UI and keeps it... Considering we have `:company-prefix-length` in there already, that would be a rather odd combination.

> If you keep using the global variables I could still do a similar thing
and bind them in cape-company-to-capf. Leaves the problem that I don't
have access to the necessary information...

To be fair, you could just as well recompute it, if you access said variables (which aren't going anywhere), just like `company--prefix-min-length` does in the latest patch.

dgutov:
This `company-set-min-length` vs `company-prefix-length` got me thinking: what if we pass min-length to `prefix` instead (and require the comparison to be done in there). Then it should get a new type of return value (e.g. `too-short`), but otherwise things *probably* get easier.

Aside from the small fact that this is a major API change, requiring all backends to be changed, or migrate using some two-step scheme.

And that still leaves `no-cache`. Luckily, capfs don't need to handle that (at least, not yet), so this could be solved in a number of ways, including being passed `min-length` anyway.

But fussing with `prefix` means I would have to try to resolve another long-standing problem (#1106 and duplicates), to only have one incompatible migration rather than two. And here I was hoping to merge this one quickly...

minad:
> What I mean is, a capf should be usable with just any UI, right? Including e.g. ivy-completion-in-region and those that might yet be written.

Sure. The function should still work with arbitrary UIs, in particular the default completion UI. But this doesn't necessarily prevent us from adding checks/warnings for popular UIs, as long as we don't change the functionality in any other way.

> To be fair, you could just as well recompute it, if you access said variables (which aren't going anywhere), just like company--prefix-min-length does in the latest patch.

No, this wouldn't work if `cape-company-to-corfu` is used with other UIs. Then `company--manual-prefix` would always be nil, right?

> I suppose we could add an extension, like :company-set-min-length, which would be a function that takes a value from the UI and keeps it... Considering we have :company-prefix-length in there already, that would be a rather odd combination.

Hmm, let's better avoid adding extensions as long as they are not well thought out. Btw, did you consider "standardizing" some of the stable Capf extensions, I mean dropping the keyword prefix, since we are talking about UI agnosticism?

:company-docsig        -> :signature-function
:company-doc-buffer    -> :help-function
:company-location      -> :location-function
:company-deprecated    -> :deprecated-function

(:company-prefix-length -> :prefix-length or :auto-prefix-length; Maybe not this one not just yet given that we are debating about prefix length handling.)

> This company-set-min-length vs company-prefix-length got me thinking: what if we pass min-length to prefix instead (and require the comparison to be done in there). Then it should get a new type of return value (e.g. too-short), but otherwise things probably get easier.

Yes, this is what I had thought initially, that you pass the value to the prefix action, see https://github.com/company-mode/company-mode/pull/1405#issuecomment-1747705089. Would it work to add a *new action* instead? You could call that for arbitrary backends and catch an error to handle backward compatibility? For the start, the new action could only be supported by the multi adapter, if that's all that is needed to make this work? Also remains the question if the model can be adapted somehow to Capfs.

It would be interesting to find ways to implement a compatible strategy update. I mean we don't end up with the most beautiful design like that, but it is still a little interesting puzzle ;)

dgutov:
> No, this wouldn't work if cape-company-to-corfu is used with other UIs. Then company--manual-prefix would always be nil, right?

What I meant, it would work just as well (not perfectly, but not worse) than `company--multi-backend-adapter` referencing those variables directly. The wrapper could look them up instead.

> Btw, did you consider "standardizing" some of the stable Capf extensions, I mean dropping the keyword prefix, since we are talking about UI agnosticism?

As long as nobody is working on any significant improvements to c-a-p-f (e.g. asynchrony, simplification, whatever else I'm forgetting), I don't see much point in forcing migrations like that. Though if you spearhead that process, I'm not going to oppose it either.

> company-prefix-length -> :prefix-length or :auto-prefix-length; Maybe not this one not just yet given that we are debating about prefix length handling

Indeed, it might also be worth waiting until the resolution of this and 1-2 other issues in this repo.

> You could call that for arbitrary backends and catch an error to handle backward compatibility?

Catching errors and doing something else is indeed the meat of 2-step migrations. I don't see the full picture yet, though.

> For the start, the new action could only be supported by the multi adapter, if that's all that is needed to make this work?

No, the issue is for the multi-adapter to be able to get new information from somewhere. E.g. from the "new action" supported by the backends being grouped, that is, most or all of them.

> It would be interesting to find ways to implement a compatible strategy update. I mean we don't end up with the most beautiful design like that, but it is still a little interesting puzzle ;)

Sure it would, ideas welcome ;) For now I only see a possibly decent design in the end (details to be worked out), with 1-2 breaking steps in the middle (breaking for `cape` as well).

minad:
>     No, this wouldn't work if cape-company-to-corfu is used with other
>     UIs. Then company--manual-prefix would always be nil, right?
> 
> What I meant, it would work just as well (not perfectly, but not worse)
> than |company--multi-backend-adapter| referencing those variables
> directly. The wrapper could look them up instead.

Right, I see.

>     Btw, did you consider "standardizing" some of the stable Capf
>     extensions, I mean dropping the keyword prefix, since we are talking
>     about UI agnosticism?
> 
> As long as nobody is working on any significant improvements to c-a-p-f
> (e.g. asynchrony, simplification, whatever else I'm forgetting), I don't
> see much point in forcing migrations like that. Though if you spearhead
> that process, I'm not going to oppose it either.

I think the migration could be quite graceful, since the old names would
have to be supported anyway for a long time in the frontends Company and
Corfu. The point is mainly to make the extensions "official" in some sense.

In a few cases the new API could be overhauled slightly, improving minor
details. In trivial cases like :company-docsig -> :signature-function
all that would needed in Company or Corfu is to look up
:signature-function in addition to :company-docsig. Then a few Capfs
could be overhauled and updated but this can take its time.

>     company-prefix-length -> :prefix-length or :auto-prefix-length;
>     Maybe not this one not just yet given that we are debating about
>     prefix length handling
> 
> Indeed, it might also be worth waiting until the resolution of this and
> 1-2 other issues in this repo.

Agree.

>     For the start, the new action could only be supported by the multi
>     adapter, if that's all that is needed to make this work?
> 
> No, the issue is for the multi-adapter to be able to get new information
> from somewhere. E.g. from the "new action" supported by the backends
> being grouped, that is, most or all of them.

I see. Then this is of course not a reasonable way forward.

>     It would be interesting to find ways to implement a compatible
>     strategy update. I mean we don't end up with the most beautiful
>     design like that, but it is still a little interesting puzzle ;)
> 
> Sure it would, ideas welcome ;) For now I only see a possibly decent
> design in the end (details to be worked out), with 1-2 breaking steps in
> the middle (breaking for |cape| as well).

Please keep me updated. If I come up with something I let you know.

I've tried to come up with potential improvements to the Capf protocol a
few times, e.g., asynchronicity, but my attempts weren't really
successful so far. The goal would also be to find something which can
gracefully upgrade the existing infrastructure.

If one looks at simple Capfs, they are not that ugly, see below. It
starts simple but gets uglier quickly for more complex tasks, but that's
not too unnatural. The question is if we can find simplifications and
extensions which add what we want, like asynchronicity, and if possible
in the least ugliest way ;)

(defun word-capf ()
  (when-let ((bounds (bounds-of-thing-at-point 'word)))
    (list (car bounds) (cdr bounds) (split-string (buffer-string)))))

dgutov:
> I think the migration could be quite graceful, since the old names would
have to be supported anyway for a long time in the frontends Company and
Corfu. The point is mainly to make the extensions "official" in some sense.

Sure. I just don't see the point on spending time on *this* rather than just about any other thing.

> Please keep me updated. If I come up with something I let you know.

Will do. I'm currently mulling over the ideas that have already come up.

> I've tried to come up with potential improvements to the Capf protocol a
few times, e.g., asynchronicity, but my attempts weren't really
successful so far. The goal would also be to find something which can
gracefully upgrade the existing infrastructure.

That's the problem with CAPF: it's not terrible by itself (and the idea was pretty decent), but all the changes over the years have made backward compatibility the highest priority. It doesn't seem to me that continuing this way is tenable.

> The question is if we can find simplifications and
extensions which add what we want, like asynchronicity, and if possible
in the least ugliest way ;)

At some point Someone (tm) will propose a whole alternative protocol, get buy-in, push it through with a migration that keeps both working for a time, and we'll all be better for it.

> If one looks at simple Capfs, they are not that ugly, see below.

As you can see, this simple example's makes `prefix` a non-trivial operation (the larger the buffer, the slower it is). A protocol which is easy to misuse is not a good thing.

minad:
> At some point Someone (tm) will propose a whole alternative protocol, get buy-in, push it through with a migration that keeps both working for a time, and we'll all be better for it.

Sure, but I've yet to see such a proposal. You are basically saying the same thng as I do. The new solution should still be *somewhat backward compatible* such that both solutions continue to work at the same time. Furthermore the new solution should be able to cover all the existing functionality, which is pretty difficult.

> As you can see, this simple example's makes prefix a non-trivial operation (the larger the buffer, the slower it is). A protocol which is easy to misuse is not a good thing.

No, there is nothing wrong or misused here according to the Capf definition. But it is obviously not a good Capf, I agree. But make the candidates a constant list of words ala company-keyword or cape-keyword, then it is good. ;)

dgutov:
> No, there is nothing wrong or misused here according to the Capf definition. But it is obviously not a good Capf, I agree. But make the candidates a constant list of words ala company-keyword or cape-keyword, then it is good. ;)

That was my point: it fully abides by the protocol, but the result is not good. Hence, not good protocol. Could be better, at least.

> Sure, but I've yet to see such a proposal. You are basically saying the same thng as I do. The new solution should still be somewhat backward compatible such that both solutions continue to work at the same time. Furthermore the new solution should be able to cover all the existing functionality, which is pretty difficult.

Perhaps you could clarify something to me: are completion boundaries/field *necessary*? I.e., are there capfs that use fields that cannot be rewritted without them while still providing the same completion behavior? My impression is mostly that they have been added so that one could still pass completion-table values to `completing-read` (and not full capfs). But file-name completion in the default UI also has this thing where you can enter `em/li/pro/`, press TAB, and have completion be performed across all fields (expanding into `emacs/lisp/progmodes/`. Which is probably handy, if I ever remembered that capability at opportune times.

So let's limit that question to in-buffer completion: should the company backend API have something like that as well (e.g. for file name completion)? I've been going back and forth on this, trying to imagine how the counterpart extension would look.

minad:
Completion boundaries are useful if you want to support completion styles, since the completion style can create a filter from only a substring of the input. That's one reason. You can implement stepwise completion where only a substring of the input is displayed at each point by the UI.

However with Capfs you could probably do the same if you stop and restart completion at the separators. But then Capfs lack a feature to communicate to the frontend that the want to continue with the completion at the position behind the separator. One could emulate this with auto completion and a short prefix, but then we are not running a single completion session but multiple sequential sessions.

dgutov:
> Completion boundaries are useful if you want to support completion styles, since the completion style can create a filter from only a substring of the input.

IIUC completion styles use delimiters from `completion-pcm-word-delimiters` to split the string. Boundaries are something in addition to that (even though they can be additionally used for splitting, it seems).

So are they foremost for moving the point to the next field after certain completion interactions?

minad:
On 10/7/23 16:08, Dmitry Gutov wrote:
>     Completion boundaries are useful if you want to support completion
>     styles, since the completion style can create a filter from only a
>     substring of the input.
> 
> IIUC completion styles use delimiters from
> |completion-pcm-word-delimiters| to split the string. Boundaries are
> something in addition to that (even though they can be additionally used
> for splitting, it seems).

Most completion styles do not know anything about words and splitting
(partial-completion is an exception). The styles simply filter the set
of candidates for the current field, but this means that the UI has to
identify the field boundaries beforehand, passing only the field
substring to the completion style for filtering.

dgutov:
> Most completion styles do not know anything about words and splitting
(partial-completion is an exception). The styles simply filter the set
of candidates for the current field, but this means that the UI has to
identify the field boundaries beforehand. The UI passes only the field
substring to the completion style.

Suppose the configured completion style is something simpler, then. Like `substring`.

Would the behavior be much different if the capf simply returned those (current) boundaries as its (beg end) instead?

minad:
> Would the behavior be much different if the capf simply returned those (current) boundaries as its (beg end) instead?

Not much, I guess, except that the completion table would be reloaded every time. It might be possible to write an adapter `capf-to-completion-table-with-boundaries`, where the resulting table could be passed to `completing-read`.

dgutov:
> where the resulting table could be passed to completing-read

...or change the API of completing-read. That's the direction I'm thinking of: could we remove more pieces from the CAPF interface while keeping the same functionality (or close to it).

minad:
Yes, but I don't think restricting the existing API will work because of backward compatibility. One could try to generalize the completion tables via cl-defgenerics somehow. This would then make it possible to introduce a new (restricted, extended, asynchronous, modern, ...) API while at the same time supporting existing completion tables. Iirc Stefan had experimented with such an approach.

But anyway, from where I am coming, Capfs returning completion tables without boundaries are not less complex than completion tables with boundaries. Of course Capfs returning tables with boundaries has a duplication of functionality.

dgutov:
> Capfs returning completion tables without boundaries are not less complex than completion tables with boundaries

Every extra detail is a complication. Those who claim to understand CAPF well, don't seem to add any new features to it, for some reason. Maybe "boundaries" don't stand out in particular, but they do for me (people are welcome to come forward with other candidates for reduction). 

> One could try to generalize the completion tables via cl-defgenerics somehow. This would then make it possible to introduce a new (restricted, extended, asynchronous, modern, ...) API while at the same time supporting existing completion tables. Iirc Stefan had experimented with such an approach.

Yep, and this approach also needs a set of primitives to base the functionalities on (the generic functions and their semantics).

dgutov:
Anyway, thank you for this digression.

I think I've settled on a backward-compatible solution, which was staring us right in the eyes, really (called `set-min-prefix`).

It's a new backend method which doesn't need support from other backends. And when not invoked, `company--multi-backend-adapter` should default to the previous behavior anyway. You might want to call it in your adapter, though, because

* Having this persistent problem fixes is better for the users of Corfu as well.
* If someone tries Company and then Corfu in the same session, in the same buffer, the variable will retain value it had been set to the last time.

minad:
> Those who claim to understand CAPF well, don't seem to add any new features to it, for some reason. Maybe "boundaries" don't stand out in particular, but they do for me (people are welcome to come forward with other candidates for reduction).

I agree that simplifications are good if possible. I claim that I understand Capfs and completion tables very well and I think you know the reason why I am not attempting to change anything.

> It's a new backend method which doesn't need support from other backends. And when not invoked, company--multi-backend-adapter should default to the previous behavior anyway. You might want to call it in your adapter, though, because...

I'll take a look.

dgutov:
> I claim that I understand Capfs and completion tables very well and I think you know the reason why I am not attempting to change anything.

Right. OTOH, you just said that adding asynchrony is hard.

Going back to the "reason", I really hope we can return to the patchset of discord one day and work on it, so that at least I can understand it more thoroughly (and like it) and better advocate for it. And with more benchmarks: I'm pretty sure the proper "deferred highlighting" has to be faster than any mutating solution (at least slightly faster anyway, if the other solution also avoids copying). I've just re-read that unfortunate discussion, and FWIW was able to understand the original patch much better than before.

minad:
On 10/8/23 17:16, Dmitry Gutov wrote:
>     I claim that I understand Capfs and completion tables very well and
>     I think you know the reason why I am not attempting to change anything.
> 
> Right. OTOH, you just said that adding asynchrony is hard.

It definitely is, but I wasn't trying hard and long enough. Maybe
motivational problems because of "reasons"?

minad:
To come back to this issue - did I understand you correctly that if I don't set the prefix in `cape-company-to-capf`, everything should continue to work as before? In `cape-company-to-capf` I don't really have access to any kind of minimal prefix value, e.g., like `corfu-auto-prefix`.

dgutov:
> To come back to this issue - did I understand you correctly that if I don't set the prefix in cape-company-to-capf, everything should continue to work as before?

I expect so.

> In cape-company-to-capf I don't really have access to any kind of minimal prefix value, e.g., like corfu-auto-prefix.

I suppose there are roughly two other options:

- The bridge looks up Corfu's variables directly. Clunky, but, again, other frontends continue working as before.
- You add a new also private-ish property to capf which you would use (in Corfu) just like Company does with `set-min-prefix`.

PR #1411, opened 2023-10-11
Add explicit caching mechanism (for backends) and fold the `keep-prefix` logic inside

### LINKED ISSUE #285: Deprecate `keep-prefix` as the value of `ignore-case`, and move it into backends
- It's ugly.
- We pretty much can't avoid duplicates doing things the current way, since we have differentiate between `don` and `Don` up until the prefix is replaced.
- The performance impact is likely to be small. If not, wait until we support `bounds` and return completions sans prefix.

### LINKED ISSUE #413: Duplicates in popup list when company-dabbrev-downcase is nil
same itme in list like this:

image 1

there are tow "subclass" in the list, how to merge it as one?

### LINKED ISSUE #1072: company-dabbrev-ignore-case and company-dabbrev-downcase are ignored when completing strings and comments in programming modes
I only use `company-dabbrev` for strings and comments completion and lately I encountered a strange behaviour: applying the completion always downcases the whole word.
For example: if I am coding in the `*scratch*` buffer and I type `"Lis"` I expect `company-dabbrev` to complete it with `"Lisp"`. This was true since I first started using this package, but now the completion resolves it with the lowercase word `"lisp"`.

This behaviour is a little annoying so I started searching for its root cause. My .emacs.d init file does not change `company-dabbrev` configuration, so its configuration variables should all be set to their default value. `company-dabbrev-ignore-case` and `company-dabbrev-downcase` are set to default too (`keep-prefix` and `case-replace` respectively) and changing their values as suggested in other issues does not produce any noticeable effect.

Searching the log history I found this commit: https://github.com/company-mode/company-mode/commit/288a7617c4c3e5f22a887a17a697ac2bfe4e567d
I am almost sure that this commit is the cause of the regression: I reverted the changes in `company--multi-backend-adapter` function and the downcasing effect disappeared.

Is this the intended behaviour? Are there any ways to circumvent the downcasing of completed words, such as customizing the configuration variables of `company-dabbrev` or using other completion backends for strings and comments?

Thank you for your hard work!

### LINKED ISSUE #1096: How to prevent loss of case in company-ispell with grouped backends
I'm using company-ispell to suggest valid words when entering text in org-mode and similair buffers. The problem I'm facing at the moment is when I try to enter a capitalised word `company-ispell` suggests the lowercase variant and if I accept it then it overwrites my captialised version with the lowercase one.

For example

```
* Hello Worl<point>
```

At point I start company and it suggests several words including `world` as a completion candidate. I select it and end up with:

```
* Hello world
```

Where I actually wanted:

```
* Hello World
```

There is an option `company-dabbrev-ignore-case` for dabbrev which I believe tries to solve a similair issue. Is there an equivalent for company-ispell?

## Version Information
- `uname -a`: `Linux mk-desktop 5.11.13-arch1-1 #1 SMP PREEMPT Sat, 10 Apr 2021 20:47:14 +0000 x86_64 GNU/Linux`
- emacs-version: `GNU Emacs 28.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.28, cairo version 1.17.4) of 2021-04-09`
- company-version: `9bc678a`

dgutov:
Implement `keep-prefix` inside the backends that advertise this capability (dabbrev and ispell), using the cache helpers, with the intention to drop support for this backend action at some later date.
 
Previously described in https://github.com/company-mode/company-mode/issues/285#issuecomment-105237695.

Resolves #285.

Fixes #413, fixes #1072, fixes #1096, resolves #1063, addresses #1226.

The more ambitious full case conversion (upcase/downcase/snake/camel, mentioned in https://github.com/company-mode/company-mode/issues/1096#issuecomment-824949311), is out of scope for this particular change, but could be added later by either altering the new helper, or adding another one. I'm concerned about the performance for certain backends (such as ispell), so maybe the latter.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,13 @@
 
 # Next
 
+* The backend command `keep-prefix` is being phased out.  The built-in backends
+  implement it internally now, which resolved a number of sharp edges (mostly)
+  around "grouped" backends.  To make that easier, several helpers were added,
+  such as `company-cache-fetch` and `company-substitute-prefix`
+  ([#1411](https://github.com/company-mode/company-mode/pull/1411)).  And
+  `company-ispell` uses the cache to keep the currently selected dictionary
+  loaded in memory between completions.
 * The "length override" behavior in grouped backends now acts on each backend
   separately ([#1405](https://github.com/company-mode/company-mode/pull/1405)).
 
--- a/company-dabbrev.el
+++ b/company-dabbrev.el
@@ -1,6 +1,6 @@
 ;;; company-dabbrev.el --- dabbrev-like company-mode completion backend  -*- lexical-binding: t -*-
 
-;; Copyright (C) 2009-2011, 2013-2018, 2021  Free Software Foundation, Inc.
+;; Copyright (C) 2009-2011, 2013-2018, 2021-2023  Free Software Foundation, Inc.
 
 ;; Author: Nikolaj Schumacher
 
@@ -180,6 +180,19 @@ This variable affects both `company-dabbrev' and `company-dabbrev-code'."
   (let ((completion-ignore-case company-dabbrev-ignore-case))
     (all-completions prefix candidates)))
 
+(defun company-dabbrev--fetch ()
+  (let ((words (company-dabbrev--search (company-dabbrev--make-regexp)
+                                        company-dabbrev-time-limit
+                                        (pcase company-dabbrev-other-buffers
+                                          (`t (list major-mode))
+                                          (`all `all))))
+        (downcase-p (if (eq company-dabbrev-downcase 'case-replace)
+                        case-replace
+                      company-dabbrev-downcase)))
+    (if downcase-p
+        (mapcar 'downcase words)
+      words)))
+
 ;;;###autoload
 (defun company-dabbrev (command &optional arg &rest _ignored)
   "dabbrev-like `company-mode' completion backend."
@@ -188,20 +201,12 @@ This variable affects both `company-dabbrev' and `company-dabbrev-code'."
     (interactive (company-begin-backend 'company-dabbrev))
     (prefix (company-dabbrev--prefix))
     (candidates
-     (let* ((case-fold-search company-dabbrev-ignore-case)
-            (words (company-dabbrev--search (company-dabbrev--make-regexp)
-                                            company-dabbrev-time-limit
-                                            (pcase company-dabbrev-other-buffers
-                                              (`t (list major-mode))
-                                              (`all `all))))
-            (downcase-p (if (eq company-dabbrev-downcase 'case-replace)
-                            case-replace
-                          company-dabbrev-downcase)))
-       (setq words (company-dabbrev--filter arg words))
-       (if downcase-p
-           (mapcar 'downcase words)
-         words)))
+     (company-dabbrev--filter
+      arg
+      (company-cache-fetch 'dabbrev-candidates #'company-dabbrev--fetch
+                           :expire t)))
     (kind 'text)
+    (no-cache t)
     (ignore-case company-dabbrev-ignore-case)
     (duplicates t)))
 
--- a/company-dabbrev.el
+++ b/company-dabbrev.el
@@ -70,10 +70,7 @@ candidate is inserted, even some of its characters have different case."
 
 The value of nil means keep them as-is.
 `case-replace' means use the value of `case-replace'.
-Any other value means downcase.
-
-If you set this value to nil, you may also want to set
-`company-dabbrev-ignore-case' to any value other than `keep-prefix'."
+Any other value means downcase."
   :type '(choice
           (const :tag "Keep as-is" nil)
           (const :tag "Downcase" t)
@@ -177,8 +174,13 @@ This variable affects both `company-dabbrev' and `company-dabbrev-code'."
                        1)))
 
 (defun company-dabbrev--filter (prefix candidates)
-  (let ((completion-ignore-case company-dabbrev-ignore-case))
-    (all-completions prefix candidates)))
+  (let* ((completion-ignore-case company-dabbrev-ignore-case)
+         (filtered (all-completions prefix candidates))
+         (lp (length prefix)))
+    (if (and (eq company-dabbrev-ignore-case 'keep-prefix)
+             (not (= lp 0)))
+        (company-substitute-prefix prefix filtered)
+      filtered)))
 
 (defun company-dabbrev--fetch ()
   (let ((words (company-dabbrev--search (company-dabbrev--make-regexp)
@@ -207,7 +209,7 @@ This variable affects both `company-dabbrev' and `company-dabbrev-code'."
                            :expire t)))
     (kind 'text)
     (no-cache t)
-    (ignore-case company-dabbrev-ignore-case)
+    (ignore-case (and company-dabbrev-ignore-case t))
     (duplicates t)))
 
 (provide 'company-dabbrev)
--- a/company-ispell.el
+++ b/company-ispell.el
@@ -80,12 +80,14 @@ If nil, use `ispell-complete-word-dict'."
        (if (string= arg "")
            ;; Small optimization.
            all-words
-         ;; Work around issue #284.
-         (all-completions arg all-words))))
+         (company-substitute-prefix
+          arg
+          ;; Work around issue #284.
+          (all-completions arg all-words)))))
     (kind 'text)
     (no-cache t)
     (sorted t)
-    (ignore-case 'keep-prefix)))
+    (ignore-case t)))
 
 (provide 'company-ispell)
 ;;; company-ispell.el ends here
--- a/company-ispell.el
+++ b/company-ispell.el
@@ -1,6 +1,6 @@
 ;;; company-ispell.el --- company-mode completion backend using Ispell
 
-;; Copyright (C) 2009-2011, 2013-2016, 2018, 2021  Free Software Foundation, Inc.
+;; Copyright (C) 2009-2011, 2013-2016, 2018, 2021, 2023  Free Software Foundation, Inc.
 
 ;; Author: Nikolaj Schumacher
 
@@ -33,29 +33,34 @@
   "Completion backend using Ispell."
   :group 'company)
 
+(defun company--set-dictionary (symbol value)
+  (set-default-toplevel-value symbol value)
+  (company-cache-delete 'ispell-candidates))
+
 (defcustom company-ispell-dictionary nil
   "Dictionary to use for `company-ispell'.
 If nil, use `ispell-complete-word-dict'."
   :type '(choice (const :tag "default (nil)" nil)
-                 (file :tag "dictionary" t)))
+                 (file :tag "dictionary" t))
+  :set #'company--set-dictionary)
 
 (defvar company-ispell-available 'unknown)
-
-(defalias 'company-ispell--lookup-words
-  (if (fboundp 'ispell-lookup-words)
-      'ispell-lookup-words
-    'lookup-words))
 
 (defun company-ispell-available ()
   (when (eq company-ispell-available 'unknown)
     (condition-case err
         (progn
-          (company-ispell--lookup-words "WHATEVER")
+          (ispell-lookup-words "WHATEVER")
           (setq company-ispell-available t))
       (error
        (message "Company-Ispell: %s" (error-message-string err))
        (setq company-ispell-available nil))))
   company-ispell-available)
+
+(defun company--ispell-dict ()
+  (or company-ispell-dictionary
+      ispell-complete-word-dict
+      ispell-alternate-dictionary))
 
 ;;;###autoload
 (defun company-ispell (command &optional arg &rest ignored)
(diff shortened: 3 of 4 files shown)


dgutov:
@iquiw I'll keep support for `keep-prefix` at least until the next release, so you can migrate at your own pace.

dgutov:
@minad I think you'll need to clear the cache from `completion-in-region-mode-hook` when the company->capf adapter is used.

minad:
Thanks!

When does this cache clearing happen in Company? `cape-company-to-capf` already calls `post-completion` via the `:exit-function` which can be used internally in the Company backend to clear the cache. See https://github.com/minad/cape/blob/116063b9ee912cbaa7318dbe6597ade4a62b3f59/cape.el#L812-L815. However the `:exit-function` is only called when successfully ending completion, not when cancelling iirc. As alternative I could also clear the cache when the `cape-company-to-capf` completion table candidates are recomputed, see https://github.com/minad/cape/blob/116063b9ee912cbaa7318dbe6597ade4a62b3f59/cape.el#L795. `cape-company-to-cache` implements its own caching based on the current input via `cape--dynamic-table`. If the input became invalid (changed), the candidates are recomputed, and the cache clearing could happen. The input validity predicate can also be overridden via the VALID argument of `cape-company-to-capf`.

This brings me to the question how I can clear the cache. Is there a backend action `clear-cache` I could call? It seems that I have to instead look at the `company--cache` internal variable directly. So far I've avoided accessing any kind of Company internals, which I find quite elegant. Every access goes through the backend itself, which even allows to implement the Company protocol without Company (see https://github.com/minad/cape#company-adapter):

~~~
cape-company-to-capf -> backend -> internals
~~~

I would like to preserve this property and avoid doing this:

~~~
cape-company-to-capf -> backend -> internals
     |                               ^  
     |                               |
     `-------------------------------'
~~~

dgutov:
> When does this cache clearing happen in Company?

In `company-cancel`, see the diff.

> This brings me to the question how I can clear the cache. Is there a backend action clear-cache I could call?

No backend action, but I could add a public function that would do that. Whether to access a private variable or not, that's a minor concern at this point, I think.

> cape-company-to-cache implements its own caching based on the current input via cape--dynamic-table. If the input became invalid (input changed), the candidates are recomputed.

The idea is to keep the cache around while the user continues typing, and then clear it at the end.

minad:
On 10/16/23 13:50, Dmitry Gutov wrote:
> No backend action, but I could add a public function that would do that.
> Whether to access a private variable or not, that's a minor concern at
> this point, I think.

Actually for me it is a concern. Right now I am not accessing anything
besides the backend as I tried to demonstrate above. I think it would be
worthwhile to preserve this property if it doesn't lead to major
complications on your side. I see some values in keeping such an
interface intact, and I believe you will also profit from such a clean
separation in the long term in Company.

>     cape-company-to-cache implements its own caching based on the
>     current input via cape--dynamic-table. If the input became invalid
>     (input changed), the candidates are recomputed.
> 
> The idea is to keep the cache around while the user continues typing,
> and then clear it at the end.

Okay, and if there is no obvious end? It should be almost as good to
clear the cache in the beginning, except that we unnecessary occupy a
little more memory?

dgutov:
> Actually for me it is a concern. Right now I am not accessing anything
besides the backend as I tried to demonstrate above. I think it would be
worthwhile to preserve this property if it doesn't lead to major
complications on your side. I see some values in keeping such an
interface intact, and I believe you will also profit from such a clean
separation in the long term in Company.

Ah no, I don't have a problem adding a new public function, e.g. something like below. As long as that works for you.

You would still have to call a function from the Company package, though. AFAICT, `cape` currently does not even do that.

> Okay, and if there is no obvious end? It should be almost as good to
clear the cache in the beginning, except that we unnecessary occupy a
little more memory?

That should also be fine. I just thought that `completion-in-region-mode-hook` is run when completion-in-region is exited, too.

```diff
diff --git a/company.el b/company.el
index 4e89910..c7441c4 100644
--- a/company.el
+++ b/company.el
@@ -1189,6 +1189,13 @@ be recomputed when this value changes."
   "Delete KEY from cache."
   (remhash key company--cache))
 
+(defun company-cache-expire ()
+  "Delete all keys from the cache that are set to be expired."
+  (maphash (lambda (k v)
+             (when (assoc-default :expire v)
+               (remhash k company--cache)))
+           company--cache))
+
 (defun company-call-backend (&rest args)
   (company--force-sync #'company-call-backend-raw args company-backend))
 
@@ -2263,10 +2270,7 @@ For more details see `company-insertion-on-trigger' and
           company--multi-uncached-backends nil
           company--multi-min-prefix nil
           company-point nil)
-    (maphash (lambda (k v)
-               (when (assoc-default :expire v)
-                 (remhash k company--cache)))
-             company--cache)
+    (company-cache-expire)
     (when company-timer
       (cancel-timer company-timer))
     (company-echo-cancel t)
```

minad:
On 10/16/23 19:54, Dmitry Gutov wrote:
>     Actually for me it is a concern. Right now I am not accessing anything
>     besides the backend as I tried to demonstrate above. I think it would be
>     worthwhile to preserve this property if it doesn't lead to major
>     complications on your side. I see some values in keeping such an
>     interface intact, and I believe you will also profit from such a clean
>     separation in the long term in Company.
> 
> Ah no, I don't have a problem adding a new public function, e.g.
> something like below. As long as that works for you.
> 
> You would still have to call a function from the Company package,
> though. AFAICT, |cape| currently does not even do that.

No, a public function wouldn't help. Ideally Company defines another
optional backend action called `reset-cache` (or `reset` or `cleanup`)
which can be called to cleanup any cached state before or after
performing completion. The idea is that everything should be channeled
through the backend, as is currently the case.

>     Okay, and if there is no obvious end? It should be almost as good to
>     clear the cache in the beginning, except that we unnecessary occupy a
>     little more memory?
> 
> That should also be fine. I just thought that
> |completion-in-region-mode-hook| is run when completion-in-region is
> exited, too.

Cape is agnostic regarding the completion UI and
`completion-in-region-mode-hook` may not be called.

minad:
On 10/16/23 21:00, Dmitry Gutov wrote:
>     Ideally Company defines another
>     optional backend action called |reset-cache| (or |reset| or |cleanup|)
>     which can be called to cleanup any cached state before or after
>     performing completion.
> 
> Then every backend would have to contain the line(s) to run the cache
> expiration, too. Departing from the brevity of how |company-cache-fetch|
> is used now. I spent more time than I care to admit optimizing that ;-(

Hmm, I thought backends don't have to implement optional actions. Only
backends which use the cache have to contain the action to clear the cache?

> If we were discussing a new common backend interface, I'd of course see
> the point of doing something less elegant but common for everybody. But
> there would be different potential ways to solve this requirement too,
> e.g. the proverbial "session object".

Well, right now the interface is sufficient for the entire access so I
thought it would be nice to preserve this property, but in the end it is
your decision of course.

>     Cape is agnostic regarding the completion UI and
>     |completion-in-region-mode-hook| may not be called.
> 
> I wonder, would this approach work if the resulting capf is used with
> |company-capf| later? If the cache is cleared every time the capf
> function is called (and the cache persists thanks to the capf data
> staying around in UIs like Corfu), then it could be a problem for
> Company, leading to much more frequent invalidations (as previously
> discussed).

I've not tried cape-company-to-capf with company-capf, but that's also
pointless or not? If cache clearing happens in the completion table, the
live time of the completion table decides on the live time of the
backend cache.

dgutov:
> Only backends which use the cache have to contain the action to clear the cache?

Yes, and they currently don't need to.

> Well, right now the interface is sufficient for the entire access so I
thought it would be nice to preserve this property, but in the end it is
your decision of course.

I think I'd rather err on the side of brevity in this case. We can revisit this at any time in the future, though.

> I've not tried cape-company-to-capf with company-capf, but that's also
pointless or not? If cache clearing happens in the completion table, the
live time of the completion table decides on the live time of the
backend cache.

I was just making a point that if the resulting capf doesn't work inside `company-capf` as well as the original backend, then it's not really UI-agnostic. But it's a matter of perspective, I suppose.

minad:
Dmitry there is no need to build up such a line of arguments if all you want is to refute my design principles. You can just tell me right away. Of course I have to adapt to some of the fine differences. I've before considered to add a cape-capf-retain middleware function which ensures that the completion table of the wrapped Capf is kept alive even for separate Capf calls. This function would be symmetric to cape-capf-buster. So far it seemed that there is no need to do that and nobody asked for such a feature.

Regarding this issue I wonder how efficient the changed backends are which rely on the cache. It seems that for ispell and dabbrev you retrieve all (!) the words and then filter them later on. Did I understand this correctly? That's quite a change of behavior. For ispell this may potentially load a huge word list?

minad:
For comparison, in Cape I have two related Capfs, cape-dict and cape-dabbrev. For cape-dict I always load the words on demand via grepping and for cape-dabbrev I use prefix search. The cached word lists are reloaded if the input changes. The capfs look quite efficient in my tests but of course don't offer the full completion style flexibility.

dgutov:
@minad But I like your design principles, so this is the way I felt appropriate to comment on them.

We could perhaps consider the cache as a separate entity, which is put on the same level as the completion tables and the UIs. Then it would make sense for the UIs and the completion tables to interact with it.

> I've before considered to add a cape-capf-retain middleware function which ensures that the completion table of the wrapped Capf is kept alive even for separate Capf calls. This function would be symmetric to cape-capf-buster. So far it seemed that there is no need to do that and nobody asked for such a feature.

Where would the cache expiration reside in such case, though?

> Regarding this issue I wonder how efficient the changed backends are which rely on the cache. It seems that for ispell and dabbrev you retrieve all (!) the words and then filter them later on. Did I understand this correctly? That's quite a change of behavior. For ispell this may potentially load a huge word list?

Indeed. For company-dabbrev, that was actually the same as before (as opposed to company-dabbrev-code). The new version's performance didn't change in that aspect, but the separate decision to use the hash table made the GC pauses much better when the prefix is empty.

The company-ispell change was related to https://github.com/company-mode/company-mode/pull/1063, where apparently each process invocation created a stutter already. I don't see such stutters now, but that naturally depends on the size of the dictionary. I don't expect there to be dictionaries that don't fit in RAM with a lot of space to spare, but if somebody reports problems with that, we could of course make caching optional. Or set up threshold, or change caching scheme.

minad:
Thanks. What I mean is that there is no need for any such indirect argumentation. I am well aware of all the limitations and I prefer getting straight to the point. ;)

> We could perhaps consider the cache as a separate entity, which is put on the same level as the completion tables and the UIs. Then it would make sense for the UIs and the completion tables to interact with it.

I think this is an overcomplication. It is simpler to just use the completion table closure as cache. This is an old idea, see for example `completion-table-with-cache`.

> Where would the cache expiration reside in such case, though?

Cache expiring wouldn't be difficult to implement actually. It could rely on the input at point. As soon as the input changes too much ("too much" is up to definition) or point moves to another line, the cache would expire. I could try to implement this. Maybe I should, given that the the closure cache used by Capfs like `cape-dict` is not effective with Company, but it would be with such a wrapper:

~~~ elisp
(setq completion-at-point-functions (list (cape-capf-retain #'cape-dict)))
~~~

> The new version's performance didn't change in that aspect, but the separate decision to use the hash table made the GC pauses much better when the prefix is empty.

My solution in Cape is different. For `cape-dict` I only return up to 100 candidates if the input prefix is too short. As soon as the input prefix gets long enough, all matching candidates are returned. If the prefix gets even longer, no recomputation is needed since all the candidates are already there anyway.

dgutov:
> I only return up to 100 candidates if the input prefix is too short. As soon as the input prefix gets long enough, all matching candidates are returned. If the prefix gets even longer, no recomputation is needed since all the candidates are already there anyway.

`ispell-lookup-words` works in similar way, so that's what our previous implementation did, but I guess in the scenario of that other PR's author the process invocation was slow, or the function wasn't optimized enough. The reporter seemingly lost interest, though, so I think I'll just wait for somebody else to come forward with a problem.

> I think this is an overcomplication. It is simpler to just use the completion table closure as cache. This is an old idea, see for example completion-table-with-cache.

Having completion table interact with the cache directly, choose keys and invalidation strategies, seems useful. E.g. see how `company-dabbrev-code` currently uses the search regexp as an additional key. Or `company-ispell` could use additional shorter-lived caches for words starting with each character of the alphabet (to amortize the `all-completions` call, which currently takes ~30ms in my testing with any input, because the list is that big).

minad:
> Having completion table interact with the cache directly, choose keys and invalidation strategies, seems useful. 

Yes, the easiest to do this is if the completion table maintains the cache itself in itself :)

dgutov:
> Yes, the easiest to do this is if the completion table maintains the cache itself in itself :)

But then the UI misses the opportunity to indicate the best times to invalidate the cache.

minad:
> Speaking of cape-dabbrev, last time I tried it out, it looked very slow. 

I think `cape-dabbrev` is not particularly good since Dabbrev itself is not very good and it cannot be better than that. I intentionally didn't implement my own Dabbrev. You probably remember our discussion a while ago about implementing a better Xabbrev which maintains some kind of tree structure internally and could be accessed more quickly for completion.

> I suppose that was because I had many buffers open, and it doesn't have any provision to abort while the input is pending, or a time limit.

Yes, there is `cape-dabbrev-check-other-buffers` which should be adjusted if necessary. Corfu interrupts backends via `while-no-input`, so we get a time limit there, see https://github.com/minad/corfu/blob/b2b9a2312f58117514724e729fda43efc4433ecd/corfu.el#L662.

dgutov:
> You probably remember our discussion a while ago about implementing a better Xabbrev which maintains some kind of tree structure internally and could be accessed more quickly for completion.

We have. I've been spending time on micro-optimizing the current optimization, though. And I think the results might be helpful when some Upstanding Citizen(tm) decides to implement a new dabbrev completion package which could be used by everybody.

In particular, using hash table(s) looks better for the GC, but that limits the implementation choices for incremental scanning.

We could create an issue here to collect the ideas, though I'm not sure if the implementation would reside in this repo.

BTW, the most overhead for `company-dabbrev-code` is in interaction with `syntax-ppss` and `parse-partial-sexp`. The non-syntax-aware dabbrev is pretty fast for most purposes, I think.

> Corfu interrupts backends via while-no-input, so we get a time limit there

Interesting. I think I would hesitate to use such an external limiter, wary of corrupting some internal state/cleanup/cache-saving.

minad:
> BTW, the most overhead for company-dabbrev-code is in interaction with syntax-ppss and parse-partial-sexp. The non-syntax-aware dabbrev is pretty fast for most purposes, I think.

I am not deeply familiar with how Dabbrev works internally. Cape calls `dabbrev--find-all-expansions`. It could very well be that there is overly expensive parsing going on there.

> Interesting. I think I would hesitate to use such an external limiter, wary of corrupting some internal state/cleanup/cache-saving.

Completion table interrupts are commonly used, e.g., in Icomplete or Vertico. But there certainly exist brittle completion backends which may not handle this well.

PR #1413, opened 2023-10-15
company-dabbrev: Fix mismatched type in company-dabbrev--fetch

projectgus:
This fixes a regression introduced in 374322cea.

Before that commit, `company-dabbrev--fetch` calls `company-dabbrev--search` which returns a hash table. Then `company-dabbrev--filter` converts it to a list (hash table's keys), then `(mapcar 'downcase words)` is applied to the list.

After that commit, if `downcase-p` is set then the hash table is passed directly to `(mapcar 'downcase words)` which produces an error `Wrong type argument: sequencep, #s(hash-table ...`

This commit changes the return type of `company-dabbrev--fetch` from a hash table to a list, although that doesn't seem to break other functionality in quick testing.

Thanks for all your work on company-mode, it's incredibly handy! :grin:

### DIFF
--- a/company-dabbrev.el
+++ b/company-dabbrev.el
@@ -183,11 +183,12 @@ This variable affects both `company-dabbrev' and `company-dabbrev-code'."
       filtered)))
 
 (defun company-dabbrev--fetch ()
-  (let ((words (company-dabbrev--search (company-dabbrev--make-regexp)
-                                        company-dabbrev-time-limit
-                                        (pcase company-dabbrev-other-buffers
-                                          (`t (list major-mode))
-                                          (`all `all))))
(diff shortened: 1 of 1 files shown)


dgutov:
Hi! Thank you for testing and for the patch.

I've pushed a slightly different one, to be more GC-friendly.

The regression was a combination of changes: you're actually more correct about pointing at the later commit (than the one I referenced in the commit message).

projectgus:
@dgutov No worries! I thought this might be something that broke in passing in the middle of the other work you were doing, but by the time I'd figured that out it seemed easier to send a prospective fix than a bug report. :grin:

PR #1417, opened 2023-10-18
Add the 'smart ignore-case' behavior

dgutov:
If the prefix has uppercase chars, then match case-sensitively, otherwise insensitively.

This seems easy enough to do when the choice is still on the backend level.

TODO: Update user options and other backends.

### DIFF
--- a/company-dabbrev-code.el
+++ b/company-dabbrev-code.el
@@ -107,7 +107,8 @@ comments or strings."
                      (not (company-in-string-or-comment)))
                  (or (company-grab-symbol) 'stop)))
     (candidates
-     (let* ((case-fold-search company-dabbrev-code-ignore-case)
+     (let* ((case-fold-search (and company-dabbrev-code-ignore-case
+                                   (company-smart-ignore-case arg)))
             (regexp (company-dabbrev-code--make-regexp arg)))
        (company-dabbrev-code--filter
         arg
@@ -126,7 +127,8 @@ comments or strings."
          :check-tag regexp))))
     (kind 'text)
     (no-cache t)
-    (ignore-case company-dabbrev-code-ignore-case)
+    (ignore-case (and company-dabbrev-code-ignore-case
+                      (company-smart-ignore-case arg)))
     (match (when company-dabbrev-code-completion-styles
              (company--match-from-capf-face arg)))
     (duplicates t)))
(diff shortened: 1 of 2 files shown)


minad:
The completion style Orderless implements such smart casing. If the input contains an uppercase string, it will default to upper case, otherwise it will match upper and lower case. Not sure, perhaps smart casing should generally be the business of the completion styles in the completion table setting?

dgutov:
> Not sure, perhaps smart casing should generally be the business of the completion styles in the completion table setting?

Maybe, but I think technically the completion styles should follow the value of `completion-ignore-case`?

Another problem (which is apparently minor enough that people don't report any issues with using Company+Orderless) is that some code here wants to know whether a backend ignores case. E.g. to implement something similar to try-completion at the end of `company-update-candidates` we need to know which value to bind `completion-ignore-case` to. Both for the sake of `try-completion` call and for the `string-prefix-p` call afterwards.

Hmm, it actually might be bad if due to smart-case the lowercase input results in a `company-common` string that includes capitals because all completions start with one, and then matching switches to case-sensitive which in turn reduces the number of completions. So maybe the `string-prefix-p` shouldn't use a third argument there anyway.

And another major use of knowing whether a backend ignores case is the internal prefix-based cache (first half of `company-calculate-candidates`), which could be removed after the next proper release.

minad:
If I understand you correctly, you mean the problem that the casing of the completion result should be adjusted? I think the semantics of the completions API with respect to that point are not clearly defined. In Cape I have `cape-dict-case-fold` and `cape-dabbrev-case-fold` to handle the case folding,  but this needs separate configuration and cannot be toggled in a smart way.

dgutov:
> If I understand you correctly, you mean the problem that the casing of the completion result should be adjusted?

Adjusting completions' casing (to camel-case or snake-case, or etc, to mirror the input) would be a separate feature.

I just meant that the current capf APIs only have a single user option, and its semantics only define two values. But if Orderless's smart-casing doesn't create any obvious problems, perhaps `completion-ignore-case` is easily extensible.

> In Cape I have cape-dict-case-fold and cape-dabbrev-case-fold to handle the case folding, but this needs separate configuration and cannot be toggled in a smart way.

What if those capfs just obeyed `completion-ignore-case`? Did this distinction come from users preferring case-folding for only some sources, and no the rest?

PR #1430, opened 2023-11-15
Fix bbdb completion

### LINKED ISSUE #1429: f64f73e broke bbdb completion
I am using emacs from git master and a recent update broke completions with bbdb.

hexmode:
Also update bbdb interface.

fixes #1429


dgutov:
Thanks for the report and the patch!

You seem to have some odd indentation rules set up (perhaps they are tailored for CL), so the last two changed lines are indented incorrectly. I've committed in your name a version where this is fixed.

PR #1459, opened 2024-03-02
Let company-ispell-dictionary be compatible with the ~user notation.

whatacold:
Hi,

Currently, the company-ispell dictionary path doesn't allow to have a `~` in it. For example, with below setting:

```elisp
(setq company-ispell-dictionary "~/.emacs.d/data/misc/english-words.txt")
```

It will error out with message:

```
Company: An error occurred in auto-begin
Company: backend (company-dabbrev company-ispell) error "grep error: look: ~/.emacs.d/data/misc/english-words.txt: No such file or directory" with args (candidates he)
```

So I'm making this PR trying to fix that.

### DIFF
--- a/company-ispell.el
+++ b/company-ispell.el
@@ -39,7 +39,8 @@
 
 (defcustom company-ispell-dictionary nil
   "Dictionary to use for `company-ispell'.
-If nil, use `ispell-complete-word-dict'."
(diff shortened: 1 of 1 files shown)


dgutov:
Thanks.

PR #1460, opened 2024-03-02
Inhibit modification hooks in *company-sps*

corngood:
See https://github.com/slime/slime/issues/814 and https://github.com/slime/slime/issues/815

`slime` does:

```
      (add-text-properties start end
                           `(modification-hooks (slime-after-change-function)
                             insert-in-front-hooks (slime-after-change-function)
                             insert-behind-hooks (slime-after-change-function)
```

`company-safe-pixel-substring` adds text with these properties into `*company-sps*` which results in the hooks being called in an unexpected buffer.

I'm not sure if this is the best way to fix this.  Perhaps the text should be read with `buffer-substring-no-properties`?


dgutov:
Yeah, I think this is better than using `buffer-substring-no-properties` (because then we'll miss all `invisible` and other display-modifying properties too). Thanks.

PR #1462, opened 2024-03-11
Replace `string-match` with `string-match-p` when appropriate and use `string-empty-p` helper

Hi-Angel:
This is a minor refactoring.

1.-st commit uses `string-match-p` if match data is unused. The places of replacement were manually examined to make sure it is the case.
2.nd commit should be safe and straightworward replacement of `(string= … "")` with `string-empty-p`
3.rd commit is similar to 2-nd one, except we getting rid of `(equal … "")`. Due to `(equal)` allowing for its arg to be of any type, each changed line was manually examined to make sure that code around it expects a string, so the parameter must be a string as well and not an arbitrary type.

### DIFF
--- a/company-dabbrev-code.el
+++ b/company-dabbrev-code.el
@@ -78,7 +78,7 @@ also `company-dabbrev-code-time-limit'."
 (defun company-dabbrev-code--make-regexp (prefix)
   (let ((prefix-re
          (cond
-          ((equal prefix "")
+          ((string-empty-p prefix)
            "\\([a-zA-Z]\\|\\s_\\)")
           ((not company-dabbrev-code-completion-styles)
            (regexp-quote prefix))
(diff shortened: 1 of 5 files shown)


Hi-Angel:
> It seems like `string-empty-p` is actually a little slower than either `equal` or `string-equal` because of the dispatch overhead (the `defsubst` facility doesn't help enough)

Oh… ☹ Do you think reimplementation of `string-empty-p` in pure C might be accepted in upstream Emacs?

dgutov:
Sure, maybe.

If I was making that decision, I would probably ask whether there exists a practical program where this function's performance would make any kind of difference.

BTW, we also have `define-inline` (a more recent addition compared to `defsubst`) which can eliminate that additional dispatch overhead.

Issue #931, opened 2019-11-13
File completion from project root


vspinu:
For most of the project infrastructures completion from the project root is the one which is needed, not the one starting from the current file. 

Would you be interested in extending files completions based on `project-roots`? I can look into it. 

I think #668 is related and should be tackled simultaneously. 

Related SE [question](https://emacs.stackexchange.com/questions/33486/specify-base-initial-path-for-company-files-completion).

dgutov:
I like your idea about triggering only after a slash has been typed.

But is starting from project root such common case? Not from one of the includes directories, for instance? E.g. in C/C++ #include directives, or similar statements in other languages.

vspinu:
I would guess that for all languages which have a notion of pwd this could be useful. At least for the development stage it often makes sense to assume pwd is the root. I always need this in R and my Clojure projects, mostly for various resource and and data localization. 

You are right that this might be too simplistic, but generalization is not hard. One could have a list of `company-file-root-directories` which are either strings or functions which produce file paths, and try completions relative to those directories.

dgutov:
> At least for the development stage it often makes sense to assume pwd is the root

I was thinking that projects usually have something like the `src` directory. Anyway...

> One could have a list of company-file-root-directories which are either strings or functions which produce file paths, and try completions relative to those directories

How about the name `company-files-search-path`? It could be a list of directories (defaulting to nil) or a function that returns a list of directories.

vspinu:
> How about the name company-files-search-path? It could be a list of directories (defaulting to nil) or a function that returns a list of directories.

Sounds good. Shall I give it a try or you want to do it yourself?

dgutov:
Please go ahead.

dgutov:
@vspinu Would you also start project-rooted completions with `./`?

dgutov:
Maybe the customization should recognize the file types and syntactic constructs which would use completion from project-root. It it would indeed differ from project to project (because search paths would often start from a project subdirectory like `src` or `lib`, not root directly).

PR #1466, opened 2024-03-26
company-files: Support completing from project-root

zhenhua-wang:
Support completing from project-root using the builtin `project.el`.

### DIFF
--- a/company-files.el
+++ b/company-files.el
@@ -46,6 +46,10 @@ This way it's easy to continue completion by typing `/' again.
 Set this to nil to disable that behavior."
   :type 'boolean)
 
+(defcustom company-files-from-project-root nil
+  "Non-nil to complete from `project-root'."
+  :type 'boolean)
+
 (defun company-files--directory-files (dir prefix)
   ;; Don't use directory-files. It produces directories without trailing /.
   (condition-case _err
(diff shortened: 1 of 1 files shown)


dgutov:
Hi!

Thanks for the PR. See the test failures (especially in Emacs 26 and newer): the value returned by `(file-exists-p (expand-file-name dir project-root))` is unused.

dgutov:
And... the code doesn't seem to work, in particular because non-rooted entries don't match anything in `company-files--regexps`.

Or was the idea to have strings starting with `./` complete relative to project root?

Perhaps you'll want to add a test or two, for examples.

zhenhua-wang:
> Hi!
> 
> Thanks for the PR. See the test failures (especially in Emacs 26 and newer): the value returned by `(file-exists-p (expand-file-name dir project-root))` is unused.

Just solved the unused issue. 

> Or was the idea to have strings starting with ./ complete relative to project root?

Thanks for the catch. I might overlook how other would think about completing from project-root, because in my own workflow, I only need this when completing relative paths... Let me rethink about this PR and get back to you later...

dgutov:
If we decide to support multiple approaches, it could be configured from the same variable, I think.

> I might overlook how other would think about completing from project-root

See the previous discussion in https://github.com/company-mode/company-mode/issues/931.

PR #1471, opened 2024-05-14
Async no cache flicker

dgutov:
* Internal handling of input interruption with throw-catch that ensures completion is not aborted.
* An interface to opt into `while-no-input` for company-capf which reuses the mechanism.
* As an semi-private feature, the completion table can just `(throw 'interrupted 'new-input)` instead, if it detects new input using its own methods.

~I suppose the main thing left to decide is whether `while-no-input` should be opt-in or opt-out for the capf backend.~

EDIT: just opt-out: see NEWS.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,14 @@
 
 # Next
 
+* Improved behavior when user types new character while completion is being
+  computed: better performance, less blinking (in the rare cases when it still
+  happened). The improvement extends to native async backends and to
+  `company-capf`.
+* As such `company-capf` now interrupts computation on new user
+  input. Completion tables that are incompatible with this behavior should get
+  updated: bind `inhibit-quit` to non-nil around their sensitive sections, or
+  simply around the whole implementation (as a fallback).
 * `company-elisp` has been removed.  It's not needed since Emacs 24.4, with all
   of its features having been incorporated into the built-in Elisp completion.
 * `company-files` shows shorter completions.  Previously, the popup spanned
--- a/company-capf.el
+++ b/company-capf.el
@@ -1,6 +1,6 @@
 ;;; company-capf.el --- company-mode completion-at-point-functions backend -*- lexical-binding: t -*-
 
-;; Copyright (C) 2013-2023  Free Software Foundation, Inc.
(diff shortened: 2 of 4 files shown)


dgutov:
@yyoncho @kiennq I think you might be interested in this too, we discussed this kind of solution back in https://github.com/emacs-lsp/lsp-mode/issues/1894, and this moves the same approach inside company-mode.

dgutov:
@minad With the new info that Emacs 29's version of Eglot [can misbehave](https://github.com/joaotavora/eglot/discussions/1127#discussioncomment-7698442) inside `while-no-input` it seems I'll have to make this opt-in, at least until that version of Emacs becomes obsolete enough.

Though if you had some clever workaround for it, I'm happy to consider it as an alternative.

BTW, have you seen any similar reports from the lsp-mode side as well?

minad:
> Though if you had some clever workaround for it, I'm happy to consider it as an alternative.

Why not simply detect the Eglot Capf and Eglot version? This would be a trivial fix.

dgutov:
Do you keep around the CAPF function that produced the current completion table? It's when querying it we'd have to decide whether to allow interruptions.

Also, regarding the version check -- any preference? The below is not free

```lisp
(benchmark-run 1 (find-library-name "eglot") (lm-version))
```

...although almost of the time is spent in `find-library-name`.

minad:
> Do you keep around the CAPF function that produced the current completion table?

No. I don't have such a workaround in Corfu. I recommend to fix broken Capfs via Cape *manually*.

> Also, regarding the version check -- any preference? The below is not free

As a cheaper alternative you could check `emacs-major-version` only. Or use `package-get-version`.

dgutov:
> No. I don't have such a workaround in Corfu. I recommend to fix broken Capfs via Cape manually.

I'd like to, but that seems bad for the OotB experience. If only it had been ELPA-only, then we wouldn't be talking about this now.

> As a cheaper alternative you could check emacs-major-version only. Or use package-get-version.

`package-get-version` works if I bind `buffer-file-name` appropriately, but that's what the call to `find-library-name` returns in my example above. The expensive one.

PR #1478, opened 2024-06-23
abbrev backend: use abbrev--active-tables to get all tables

blahgeek:
previous, tables in `abbrev-minor-mode-table-alist` was not considered


dgutov:
Thanks!

PR #1479, opened 2024-06-24
Make the user of while-no-input in company-capf opt-in

dgutov:
The problem described in https://github.com/joaotavora/eglot/discussions/1127#discussioncomment-7698442 still seems to apply to the Eglot version that comes with Emacs 29.1-29.4, and so might affect a fair number of users, given that an "ELPA core" package doesn't update by default (until the user triggers the first update manually), and a lot of users won't.

I went through the thread, but only managed to reproduce it once (with Emacs 29.1), and not for the lack of trying.

Even if being interruption-safe is the "proper" behavior for completion functions, it seems that at least for this case exposing the problem (i.e. using a capf function inside interrupting context) doesn't really help to have it fixed - and not for the lack of trying.

So it the right move seems to be the change to opt-in for now, and the re-evaluate when the 29 version of Emacs has been superseded, and also see whether some/many/most capfs have adopted the new property, indicating that they find this feature helpful.

The downside to the change seems minor: as you @minad have noted, only a handful of CAPFs constitute the majority of usage, and both lsp-mode and Eglot currently have their own solution to interruption on new input (so their performance should stay unchanged).  But this might help them migrate, too.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,13 +4,11 @@
 
 * Improved behavior when user types new character while completion is being
   computed: better performance, less blinking (in the rare cases when it still
-  happened). The improvement extends to native async backends and to
+  happened). This affects native async backends and is opt-in with
   `company-capf`.
-* As such `company-capf` now interrupts computation on new user
-  input. Completion tables that are incompatible with this behavior should get
-  updated: bind `inhibit-quit` to non-nil around their sensitive sections, or
-  simply around the whole implementation (e.g. using
-  `cape-capf-noninterruptible` from [cape](https://github.com/minad/cape/)).
+* For that, `company-capf` supports interrupting computation on new user
+  input. Completion functions that want to take advantage of this behavior
+  should include `:company-use-while-no-input` in the returned properties list.
 * `company-elisp` has been removed.  It's not needed since Emacs 24.4, with all
   of its features having been incorporated into the built-in Elisp completion.
 * `company-files` shows shorter completions.  Previously, the popup spanned
--- a/company-capf.el
+++ b/company-capf.el
@@ -189,10 +189,12 @@ so we can't just use the preceding variable instead.")
                      table pred))))
     (company-capf--save-current-data res meta)
     (when res
-      (let* ((candidates (company-capf--candidates-1 input table pred
+      (let* ((interrupt (plist-get (nthcdr 4 res) :company-use-while-no-input))
+             (candidates (company-capf--candidates-1 input table pred
                                                      (length input)
                                                      meta
-                                                     non-essential))
+                                                     (and non-essential
+                                                          interrupt)))
              (sortfun (cdr (assq 'display-sort-function meta)))
              (last (last candidates))
              (base-size (and (numberp (cdr last)) (cdr last))))


dgutov:
Thanks to @jdtsmith for the reference and the related work.

minad:
Not a useful change in my opinion.

jdtsmith:
Perhaps a brief how-to guide for CAPF authors on how to protect sensitive code against interruption with `(let ((throw-on-input nil...`?  The change to eglot was in the end quite tiny, but it took some time to identify the conflict.

dgutov:
> Not a useful change in my opinion, since this creates a lot of work for all existing Capfs.

I'm not quite happy with it myself, but I don't see other proposals that would reach the same goals.

> Most other Capfs will work with interrupts, so it would make more sense to add a configuration option to opt out for certain Capfs, which don't handle interrupts well.

An opt-out can be handled in code. But more importantly, this is not something we could enact for a lot of code that is currently out there. If Eglot weren't merged into core, perhaps the situation would be different.

> Note that completion-preview-mode and icomplete-mode (in buffer) also use interrupts.

Using `icomplete-mode` as in-buffer completion seems very rare (IME), and `completion-preview-mode` is new in Emacs 30, so they don't have the same constraints.

dgutov:
@jdtsmith

> Perhaps a brief how-to guide for CAPF authors on how to protect sensitive code against interruption with (let ((throw-on-input nil...? The change to eglot was in the end quite tiny, but it took some time to identify the conflict.

Indeed, tiny it was. What that change did, however, is (AFAICT) basically disabled the `while-no-input` context is large portions of the execution. So I'm not sure if the same recipe can or should be scaled to other completions functions. Eglot basically ignores the `while-no-input` macro in which we would wrap it because it uses its own solution for interrupting.

But if you have some recommendations for how to write completion functions in safes manner, you're welcome to post them on this wiki or somewhere else (e.g. Corfu's), and I can link to the resulting page from NEWS where the `:company-while-no-input` property is introduced.

Note that the question of how to write these better (interruption-proof) has been brought up on the mailing lists recently, see [this thread](https://lists.gnu.org/archive/html/emacs-devel/2024-05/msg00230.html) for the recommendations that came up.

minad:
What I meant above is that it might be simpler and less work for everyone involved if we would add a customization option to Company instead. Things would remain backward compatible this way. The option could be set to a reasonable default in the beginning, e.g., `(not eglot-completion-at-point-function)`.

~~~
(defcustom company-capf-interruptible nil
  "Configurable Capf interrupts.
- t           : Interrupt all
- nil         : Interrupt none
- '(icapf jcapf) : Interrupt only icapf and jcapf
- '(not ncapf mcapf) : Interrupt all except ncapf and mcapf."
  :type '(choice (const :tag "None" nil)
                 (const :tag "All" t)
                 (repeat :tag "Only these" symbol)
                 (cons :tag "Except these" (const not) (repeat symbol))))
~~~

dgutov:
> (defcustom company-capf-interruptible nil

Right, something like that will be necessary either way medium-term, if we're to flip the behavior to on someday.

But short-term, what happens if one of the capf functions supports this, and another does not? You might miss an opportunity to handle the disparity in `cape-capf-super` - wrapping the non-supporting ones in a `throw-on-input nil` block.

And if authors of some completion backends, instead of adding the new property, will start recommending that the users just flip the default, we a) won't be able to track those completion functions separately in code (see above), b) it would be harder to evaluate the ecosystem's readiness for the flip, c) some people will flip this on, see the occasional problem with a certain capf that is hard to identify, and fail to report this because this happens rarely and is difficult pinpoint and reproduce (not a good thing, IMO).

In any case, are there particular capfs that you suppose will benefit from this right away? Like we discussed, eglot and lsp-mode have their own solutions. Just elisp and pcomplete in some shell buffers? Is the difference noticeable with elisp?

dgutov:
Anyway, I'm merging this, but that doesn't mean the end of the discussion.

minad:
> But short-term, what happens if one of the capf functions supports this, and another does not? You might miss an opportunity to handle the disparity in cape-capf-super - wrapping the non-supporting ones in a throw-on-input nil block.

No, the concept in Cape is different. Unsupported Capfs have to be wrapped by `cape-capf-noninterruptible`, before passing them to `cape-capf-super`.

dgutov:
> No, the concept in Cape is different.

Yeah, very well, it's up to you. I hope you see how it could use this to do things more automatically (at the expense of the capf authors, of course).

PR #1488, opened 2024-08-05
company-complete-common: Generalized "expand common" behavior

### LINKED ISSUE #1260: Composite backend ((company-capf company-files)) does not work
How to reproduce:
File
```
int main( int argc, char** argv )
{
  std::string a = "./001_";
  return 0;
}
```
Create any file starting from 001_ . In my case it was 001_test.txt

company-diag:
```
Emacs 27.2 (x86_64-slackware-linux-gnu) of 2021-10-18 on laptop.lockywolf.net
Company 0.9.13

company-backends: ((company-capf company-files))

Used backend: (company-capf company-files)

Value of c-a-p-f: (eglot-completion-at-point t)
Major mode: c++-mode
Prefix: ("001_" . t)

Completions: none
```

#### Describe the issue

M-x company-manual-begin RET does not produce candidates. However, `company-files` does produce candidates. 
Setting `company-backends` to `(company-files)` produces candidates as well.

I tried edebug on company's code, and found that the two backends produce diffent prefixes. `company-files` produces `./001_`, `company-capf` produces `("001_" . t)`. Not sure if it is related.

#### Expected behavior
See company-files' list of candidates. (As the c-a-p-f's one would be empty).

#### The error backtrace
No Emacs error present.

#### Additional context
I use eglot for IDE-like functionality, and only using `company-capf` works just fine, produces the list of candidates as expected. 
It seems to me that it is the merging functionality that is somehow self-contradictory.

dgutov:
* New backend action `expand-common`.
* `company-capf` implementation that calls `completion-try-completion` (and support in `company-dabbrev-code` and `company-etags` - when the use of completion styles is enabled in each backend).
* Support for grouped backends returning prefixes/suffixes of different length (resolves #1260).
* Support for `expand-common` in grouped backends as well.

Demo:

[screencast.webm](https://github.com/user-attachments/assets/61d68d98-129b-4249-be20-9aea03a055b6)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,9 +2,11 @@
 
 # Next
 
-* `company-complete-common` now performs generalized "expand common part"
-  completion when the current backend supports that.  In particular, for
-  `completion-at-point-functions` it queries `completion-try-completion`.
+* `company-complete-common` now performs generalized ([expand common
+  part](https://github.com/company-mode/company-mode/pull/1488)) completion when
+  the backend supports that. In particular, for `completion-at-point-functions`
+  it queries `completion-try-completion`. `company-dabbrev-code` and
+  `company-etags` also do that when `completion-styles` support is enabled.
 * `company-dabbrev-other-buffers` and `company-dabbrev-code-other-buffers` can
   now take a function as its value (#[1485](https://github.com/company-mode/company-mode/issues/1485))
 * Completion works in the middle of a symbol
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,9 @@
 
 # Next
 
+* `company-complete-common` now performs generalized "expand common part"
+  completion now when the current backend support that.  In particular, for
+  `completion-at-point-functions` it queries `completion-try-completion`.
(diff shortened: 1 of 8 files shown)


dgutov:
Some testing with LSP clients (both of them) shows some unexpected behavior for try-completion:

[Screencast from 2024-08-19 04-57-37.webm](https://github.com/user-attachments/assets/48cdc678-dbb6-4ff4-bc6d-91d99fde69d5)

Report for Eglot: https://debbugs.gnu.org/72705

Same behavior with Corfu, so maybe it's still okay for most users most of the time.

dgutov:
Shout out to @minad. Never mind the previous comment - although if you have any suggestions, those are welcome too.

Since this is the point where company backends seem reasonably complete (and yeah, this required adding features similar to CAPF's boundaries and try-completion), maybe it would help evolving CAPFv2. Anyway, here's hoping `company--multi-backend-adapter` continues working okay with `cape`.

minad:
Thanks. I'll take a look.

Issue #1491, opened 2024-08-29
Severe performance regression after the recent update


ranjeethmahankali:
#### Output of the command `M-x company-diag`
```
Emacs 29.4 (x86_64-w64-mingw32) of 2024-07-05 on AVALON
Company 0.10.2

company-backends: (company-bbdb company-semantic company-cmake company-capf company-clang company-files
              (company-dabbrev-code company-gtags company-etags company-keywords)
              company-oddmuse company-dabbrev)

Used backend: company-capf
Value of c-a-p-f: (lsp-completion-at-point tags-completion-at-point-function)
Major mode: c++-mode
Prefix: ("" "" nil)

Completions:
  #("VertH" 0 5 (lsp-completion-item (:filterText #("VertH" 0 5 (match-data (0 0))) :insertText "VertH" :insertTextFormat 2 :kind 8 :label "VertH" :score 93.18599700927734 :sortText "3d45a0c5VertH" :textEdit (:newText "VertH" :range (:end (:character 33 :line 538) :start (:character 33 :line 538))) :_emacsStartPoint 16545) lsp-sort-text "3d45a0c5VertH" lsp-completion-start-point 16545 lsp-completion-markers (16545 #<marker (moves after insertion) at 16545 in filename_redacted.cpp>) lsp-completion-prefix "")) " (Interface)"
  #("mMesh" 0 5 (lsp-completion-item (:detail "Mesh &" :filterText #("mMesh" 0 5 (match-data (0 0))) :insertText "mMesh" :insertTextFormat 2 :kind 5 :label "mMesh" :score 74.3667984008789 :sortText "3d6b4433mMesh" :textEdit (:newText "mMesh" :range (:end (:character 33 :line 538) :start (:character 33 :line 538))) :_emacsStartPoint 16545) lsp-sort-text "3d6b4433mMesh" lsp-completion-start-point 16545 lsp-completion-markers (16545 #<marker (moves after insertion) at 16545 in filename_redacted.cpp>) lsp-completion-prefix "")) " Mesh & (Field)"
[...]

PR #1492, opened 2024-09-08
fix performance issue with projectile running hook on candidate switching

sk1e:
Hi, Dmitry! I encountered performance issue on completion candidate switching and discovered that problem is in projectile running something heavy in `buffer-list-update-hook`. This PR helps me to fix this issue.

### DIFF
--- a/company.el
+++ b/company.el
@@ -3327,7 +3327,7 @@ from the candidates list.")
         (inhibit-modification-hooks t)
         (dedicated (window-dedicated-p))
         (hscroll (window-hscroll))
-        window-configuration-change-hook)
+        window-configuration-change-hook buffer-list-update-hook)
     (with-current-buffer (get-buffer-create " *company-sps*")
       (unwind-protect
           (progn


sk1e:
Looks like that [this](https://github.com/company-mode/company-mode/issues/1491) is related

dgutov:
Hi! Thanks for the report.

I haven't been able to reproduce the performance regression so far, but the change makes sense.

Could you test this addition too? Does it make things even faster?

Which version of Emacs are you using?

```diff
diff --git a/company.el b/company.el
index cd31a07..963c25d 100644
--- a/company.el
+++ b/company.el
@@ -3295,7 +3295,8 @@ from the candidates list.")
           (car (buffer-text-pixel-size nil nil t))
         (let ((wb (window-buffer))
               (hscroll (window-hscroll))
-              (dedicated (window-dedicated-p)))
+              (dedicated (window-dedicated-p))
+              buffer-list-update-hook)
           (unwind-protect
               (progn
                 (when dedicated
@@ -3327,7 +3328,8 @@ from the candidates list.")
         (inhibit-modification-hooks t)
         (dedicated (window-dedicated-p))
         (hscroll (window-hscroll))
-        window-configuration-change-hook)
+        window-configuration-change-hook
+        buffer-list-update-hook)
     (with-current-buffer (get-buffer-create " *company-sps*")
       (unwind-protect
           (progn

```

sk1e:
I am using GNU Emacs 29.4 (build 3, x86_64-pc-linux-gnu, GTK+ Version 3.24.33, cairo version 1.16.0) of 2024-08-19

> Could you test this addition too? Does it make things even faster?

I don't get performance increase in candidate selection, but probably it will affect some other functions. Here are my benchmarks:

Original version: 
emacs3

PR version:
emacs4

PR+diff version:
emacs5

PR #1505, opened 2024-12-29
dabbrev-code: optionally ignore candidate prefixes

dvzubarev:
Hi, I've added a custom variable `company-dabbrev-code-ignore-cand-prefix` that is regexp  of a prefix that should be ignored when collecting candidates. It works well with flex style.
For example, given this buffer and `company-dabbrev-code-ignore-cand-prefix` is set to "_":
```
_set_metric
set_metric

sm|
```
I will receive  both keywords as candidates.

What do you think?

### DIFF
--- a/company-dabbrev-code.el
+++ b/company-dabbrev-code.el
@@ -79,6 +79,16 @@ parameter and returns a list of major modes to search.  See also
                  (list :tag "Custom list of styles" symbol))
   :package-version '(company . "1.0.0"))
 
+(defcustom company-dabbrev-code-ignore-cand-prefix nil
+  "Regex of prefix that should be ignored when collecting candidates.
+For example, you may want to set this variable to \"_\".  Then if you type
+\"temp\" you will receive both candidates: \"_tempvar\" and \"tempvar\".
+Note that if you input is matched against this regex this variable will be
+ignored when collecting candidates.  So for the input \"_temp\" you will
+receive only \"_tempvar\" candidate."
+  :type '(regexp :tag "Regexp")
+  :package-version '(company . "1.0.2"))
+
 (defvar-local company-dabbrev--boundaries nil)
 (defvar-local company-dabbrev-code--sorted nil)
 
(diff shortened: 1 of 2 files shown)


dgutov:
Hi!

Won't you get (almost) the same list of completions by using a `substring` completion style? Or `partial-completion`, etc.

dvzubarev:
When I set `(setq company-dabbrev-code-completion-styles '(partial-completion))` or to `substring` I do not get any competitions at all for the example above. 
For `partial-completion`  and buffer 
```
set_metric
_set_metric
http_server

1. se|
2. _se|
```
For 1. I get only `set_metric`. When prefix is changed to `_se|` I get `_set_metric` and `http_server`.
The same for substring.

dgutov:
> For 1. I get only set_metric. ... The same for substring.

Good point, the assumption that the first prefix char (when present) matches the beginning of the symbol was baked in.

I've pushed a change now that makes the backend return more symbols when (for example) the `substring` or `flex` style are used.

`partial-completion` doesn't match in your example, but that's by design. To change that, you'll need a recent build of Emacs 31 and `completion-pcm-leading-wildcard` set to `t`. So the full example will be:

```el
(setq company-dabbrev-code-completion-styles '(partial-completion)
      completion-pcm-leading-wildcard t)
```

Now, the downside of that change (6b4f285) is that the scan will do more work even when the configured completion styles is not that lax, resulting in fewer completions on average (given the time limit of 0.1s). Not sure if there are "legal" improvements to that - without hardcoding the list of supported styles. It can be revisited if/when people actually notice the difference.

dgutov:
@dvzubarev LMK how the change works for you, it's in master.

dvzubarev:
It definitely works, thank you. It is more general solution since it also matches arbitrary substrings in the candidates.

> without hardcoding the list of supported styles. 

It works now with third-party styles (e.g. fussy). I think, it would be better to introduce a custom variable with the list of styles that might benefit from more candidates. Or somehow derive this from the style itself.

dgutov:
Thanks for testing!

> I think, it would be better to introduce a custom variable with the list of styles that might benefit from more candidates. Or somehow derive this from the style itself.

And extra knob is an option indeed, except it's one more non-standard thing the user'd have to pay attention to.

I was thinking about automatic detection as well, but the style might require special delimiters (like pcm's `completion-pcm-word-delimiters` but different), so it would be a 90% solution as well. So I think we'll wait for feedback indicating that this really is a problem.

PR #1514, opened 2025-04-26
Add keywords and builtins for sh-mode

### LINKED ISSUE #820: Addition of sh-mode keywords
I was planning to add some keywords for `sh-mode` but then realized it is not so clear at what level you should put it. It is tricky because there are many different shells with their specialties and that much of the functionality in shells is in regular built-ins (e.g. `return`) which kind of could be regarded as keywords as they are heavily used in syntax.

First problem I thought about solving by basing the keywords only on the POSIX-specification and leave out the custom things.

The list could be at one of three levels or further additions could also be made.

```emacs-lisp
  '(sh-mode
    ;; Keywords
    "case" "do" "done" "elif" "else" "esac" "fi" "for" "if" "in" "then" "until" "while"
    ;; Special Built-In Utilities
    "break" "continue" "eval" "exec" "exit" "export" "readonly" "return" "set" "shift" "times" "trap" "unset"
    ;; Utilities with specified behaviour
    "alias" "bg" "cd" "command" "false" "fc" "fg" "getopts" "hash" "jobs" "kill" "newgrp" "pwd" "read" "true" "umask" "unalias" "wait")
```

Which level would you suggest or is this a lost cause in general?

References:
- [Reserved Words (keywords)](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_04)
 - [Special Built-In Utilities](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_14)
 - [Utilities with specified behaviour](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_09_01_01)

saf-dmitry:
This commit will add keywords and names of shell builtins for `sh-mode`. This should fix #820.

### DIFF
--- a/company-keywords.el
+++ b/company-keywords.el
@@ -377,6 +377,11 @@
      "new" "null" "object" "override" "package" "private" "protected"
      "return" "sealed" "super" "this" "throw" "trait" "true" "try" "type" "val"
      "var" "while" "with" "yield")
+    (sh-mode
+     "break" "case" "continue" "do" "done" "elif" "else" "esac" "eval"
+     "exec" "exit" "export" "false" "fi" "for" "function" "if" "in" "readonly"
+     "return" "set" "shift" "test" "then" "time" "times" "trap" "true" "unset"
+     "until" "while")
     (swift-mode
      "Protocol" "Self" "Type" "and" "as" "assignment" "associatedtype"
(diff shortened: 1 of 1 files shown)


dgutov:
Thanks!

PR #1525, opened 2026-01-18
Adding company-childframe frontend, for graphically rendered popups

dgutov:
Most of the code has been imported from  company-childframe, so credit is due to its authors: Clément Pit-Claudel, Feng Shu and others. In part this was motivated by some parts of that package which don't have "clean" copyright status from the FSF's point of view, so here we have removed a bunch of features:

* The list of backends or metadata in the popup.
* The posframe-quickhelp feature.

Neither is a decision set in stone - we can bring those later (probably with a clean rewrite, if only for copyright assignment purposes). Feature requests could move motivate it too.

### Improvements

* Better integration - no advices or overriding functions.
* Mouse interactions with the popup work (selecting, changing selection, scrolling).
* [Linux/X11] Less flickering when the popup is resized, moved or hidden.
* Border around the popup.

image

### Things that stayed the same

* "Real" popup implementation, displayed over variable font text, images, inner window borders.
* Works in TTY too.

### How to try it

```el
(setq company-frontends '(company-childframe-unless-just-one-frontend
                          company-preview-if-just-one-frontend
```

### Tips

[*Linux/X11 only*] Changes upstream that improve how smoothly the popup is shown/resized/hidden:

* https://github.com/tumashu/posframe/pull/151, https://github.com/tumashu/posframe/pull/156.
* Emacs core: [debbugs#80369](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=80369) (fixed), [debbugs#80662](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=80662) (**pending**).

_If your Emacs does not include the fix_ (e.g. the current release 30.1), and you're using GNU/Linux, try the _GTK3_, _PGTK_, or the _no-toolkit_ build. Each has some issues, but PGTK is best WRT to flickering (its absence).

The macOS port doesn't seem to be affected. [Untested, but W32/Haiku/Android should be same.]

Also to reduce the amount of popup resizing, try `(setq company-tooltip-width-grow-only t)`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,9 @@ jobs:
       - name: Checkout Company
         uses: actions/checkout@v3
 
+      - name: Download dependency
+        run: make download
+
       - name: Run tests
         run: make test-batch
 
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,6 @@
 EMACS = emacs
+CURL=curl --silent
+POSFRAME_URL=https://raw.githubusercontent.com/tumashu/posframe/refs/heads/master/posframe.el
 
 ALL_TARGETS = help package clean test test-gui test-batch compile compile-warn
 
@@ -33,6 +35,9 @@ test-gui:
 test-batch:
 	${EMACS} -Q --batch -L . -l test/all.el \
 	--eval "(ert-run-tests-batch-and-exit '(not (or (tag interactive) (tag gui))))"
+
+download:
+	${CURL} ${POSFRAME_URL} > posframe.el
 
 compile:
 	${EMACS} -Q --batch -L . -f batch-byte-compile company.el company-*.el
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,9 @@
 
 # Next
 
+* New built-in frontend using "real graphical" widget for the popup
+  ([#1525](https://github.com/company-mode/company-mode/pull/1525)).
+  This also adds a hard dependency on the package `posframe`.
 * The minimum required version of Emacs is now 26.1.
 * `TAB` binding changed to `company-complete-common-or-cycle`, and `backtab`
   binding to `company-cycle-backward`
--- a/company-childframe.el
+++ b/company-childframe.el
@@ -171,9 +171,11 @@ For COMMAND refer to `company-frontends'."
 
 (defun company-childframe-unless-just-one-frontend (command)
   "`company-childframe-frontend', but not shown for single candidates."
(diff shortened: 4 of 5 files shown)


yugaego:
AFAIU, the package `posframe` needs to be installed before enabling this new frontend? Is this the only requirement?

dgutov:
@yugaego It's the only new dependency, yes.

Thanks for expressing interest, I'm still brewing a couple of things here, with a potential improvement in https://debbugs.gnu.org/cgi/bugreport.cgi?bug=80662 and another pending PR for posframe, to improve the flickering aspect further (it bothers me a lot).

Will update this soon.

yugaego:
@dgutov I enabled the childframe frontend today to test it. With it, Emacs hangs when I click a candidate row with the left mouse button (which I did accidentally, but the issue is reproducible). I noticed no issues so far interacting with the pop-up via keyboard.

yugaego:
@dgutov The same happens with clean Emacs install (Emacs 30.2 aarch64-apple-darwin25.3.0), having only company & posframe installed (+ tested with #'posframe-mouse-banish-simple). `C-g` or `toggle-debug-on-error` aren't functional. 

Unfortunately crash reports and backtraces I can produce don't reveal obvious fixes to me. I am attaching a backtrace just in case. Anyway, this does not look like something `company-childframe` is responsible for per se.

https://debbugs.gnu.org/cgi/bugreport.cgi?bug=79911 may be related, I guess.

[backtrace.txt](https://github.com/user-attachments/files/26635022/backtrace.txt)

dgutov:
@yugaego Hey, thank you. I've pushed a fix, hopefully.

At least over here I could `C-g C-g C-g` out of that state to be able to investigate. If it really went into a "full" freeze on macOS, that would be concerning.

yugaego:
@dgutov Yes, it was a "full" freeze: `C-g C-g C-g` didn't interrupt the state either.
The fix works for me.
[As I understand it, the child frame mouse clicks now are properly registered by `company-childframe`. Otherwise a mouse event repeats endlessly as unhandled, getting back into `unread-command-event` via `company-select-mouse`.]

dgutov:
@yugaego Thanks for testing! The possibility of such a loop is indeed concerning.

I've tried to tweak the code more, to avoid it more reliably and make the execution path more regular. Hopefully it will work the same across platforms.

dgutov:
Also, the flickering issues that I've been focusing on, seem to be limited to Linux/X11.

MacOS, W32, PGTK should be mostly fine in this regard already.

dgutov:
It's time :rocket:

PR #1536, opened 2026-05-19
Remap the C-M-i bindings to the built-in command

dgutov:
Seems handy and closes one of the usability hurdles.

Would be nicer to use some existing hook, but doing a `completion-in-region-function` most likely implies migrating to `c-a-p-f` entirely; maybe someday.


dgutov:
@monnier Stefan, hi, do you think the command `complete-symbol` and its `info-complete-symbol` path have any use these days? It's bound in some modes but not others, rather arbitrarily, and scanning the Info files for completions might not be too popular anymore (though clever). But perhaps in some academia circles, with Octave/Python/etc.

Or if you have other thoughts regarding it, please share.

monnier:
> @monnier Stefan, hi, do you think the command `complete-symbol` and its
> `info-complete-symbol` path have any use these days?

It dates back to before `completion-at-point`.  I didn't want to rock
the boat any further back then, so I tried my best to keep that as
it was, but IMO `complete-symbol` and `info-complete(-symbol)` should
be deprecated.

Nowadays modes/people that want those completions should use
`info-lookup-completions-at-point` (in a `completion-at-point-function`).

dgutov:
> I tried my best to keep that as
it was, but IMO `complete-symbol` and `info-complete(-symbol)` should
be deprecated.

Makes sense, thanks! Then I'm not going to carry that part of the UI over here.

> Nowadays modes/people that want those completions should use
`info-lookup-completions-at-point` (in a `completion-at-point-function`)

I'll just note that it doesn't work as-is (both of the arguments are marked as required, and TOPIC is *really* required). Also `:exclusive 'no` forces prefix completion, which I've just spent some time puzzling about.

Just filed the former as bug#81079. And related bug#81072, previously.

PR #1545, opened 2026-07-17
Editorial documentation changes

yugaego:
I like how that split into childframe/pseudo-tooltip subsections worked. And I've got a couple more suggestions.

### DIFF
--- a/doc/company.texi
+++ b/doc/company.texi
@@ -626,7 +626,16 @@ This group of frontends displays completion candidates in a
 rectangular tooltip (aka @w{popup}).  Company provides multiple
 @emph{tooltip frontends}, listed below.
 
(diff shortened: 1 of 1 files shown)


dgutov:
Solid change, thanks!

I've expanded the intro there a bit. Might be redundant, but it seems to fit the customary style for Emacs' manual: https://github.com/company-mode/company-mode/commit/f68784794b2b1bffa1acc3a27a493d7b4a142038

dgutov:
I've pushed a couple more edits on top of that.

Also it might be time to revisit the contents of https://github.com/company-mode/company-mode/wiki/Third-Party-Packages - a lot of them are either archived or not really recommended these days.

